PR #1, opened 2019-07-31
Initial build with CRA

hatched:


package.json
```diff
@@ -0,0 +1,53 @@
...
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
+  },
+  "eslintConfig": {
+    "extends": "react-app"
+  },
+  "dependencies": {
+    "node-sass": "^4.12.0",
```
squidsoup: You should probably just pin these dependencies now, otherwise renovate will be angry with you.

src/index.scss
```diff
@@ -0,0 +1,4 @@
+@import 'node_modules/vanilla-framework/scss/vanilla';
```
squidsoup: I think this can just be `@import "vanilla"`?

package.json
hatched: Done

src/index.scss
hatched: It complains that it's not able to find it and won't build when I do that. You might be mistaking it with `@include vanilla` to import the whole framework.

hatched:
Thanks @squidsoup

PR #4, opened 2019-08-01
Add Prettier

barrymcgee:
## Done

Copy over Prettier config from juju-gui

## QA

- Run `./run exec yarn run prettier-check`
- Verify no errors


.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "bracketSpacing": false,
+  "singleQuote": true,
+  "printWidth": 95
+}
```
squidsoup: Hey @barrymcgee, when did we decide on these overrides? We just the defaults in rbac and maas-ui.

barrymcgee: I agree. However, these were the compromises I had to make to get Prettier accepted onto the old GUI codebase. I'd happily use the defaults, especially as our other projects do - over to you @hatched?

PR #10, opened 2019-08-07
Tweak side nav designs to match designs

barrymcgee:
## Done

Tweak side nav designs to match designs

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check layout sidebar is full height

### DIFF
--- a/src/scss/_layout.scss
+++ b/src/scss/_layout.scss
@@ -5,7 +5,7 @@
 }
 
 .l-side {
-  background: $color-light;
+  background: #F7F7F7;
   padding: 1.5rem;
 }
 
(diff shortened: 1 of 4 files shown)


src/scss/_layout.scss
```diff
@@ -0,0 +1,14 @@
+.l-container {
+  display: grid;
+  grid-template-columns: 150px 1fr;
+  min-height: calc(100vh - 48px);
+}
+
+.l-side {
+  background: #F7F7F7;
```
barrymcgee: Can't access a Vanilla variable here.. will address.

hatched: Thanks for the note, are you going to do this in a quick follow-up? Or should we add an issue to follow-up?

barrymcgee: Raised issue: https://github.com/canonical-web-and-design/jaas-dashboard/issues/13

PR #11, opened 2019-08-07
Add jujulib and connect to JAAS on load.

hatched:
## Done

When starting the application, asynchronously connect to JAAS, present the user the macaroon login link and then list their models on successful login in the browser console.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The browser console will have a link to click to log in, do so.
- After a moment it'll list your models in the browser console.

### DIFF
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -1,0 +1,35 @@
+import jujulib from '@canonical/jujulib';
+import modelManager from '@canonical/jujulib/api/facades/model-manager-v5.js';
+import {Bakery} from '@canonical/macaroon-bakery';
+
+
+const options = {
+  debug: true,
+  facades: [
+    modelManager
+  ],
+  bakery: new Bakery({
+    visitPage: resp => {
+      // XXX Surface message to UI.
+      console.log('visit this URL to login:', resp.Info.VisitURL);
+    }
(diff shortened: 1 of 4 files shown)


src/juju/index.js
```diff
@@ -0,0 +1,35 @@
...
+  ],
+  bakery: new Bakery({
+    visitPage: resp => {
+      // XXX Surface message to UI.
+      console.log('visit this URL to login:', resp.Info.VisitURL);
+    }
+  })
+};
+
+const controllerURL = 'wss://jimm.jujucharms.com/api';
```
squidsoup: Worth putting this in a `.env`? https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables

hatched: Yeah actually there will be a number of options that we will want to provide to the application so I wasn't sure what the best way to do that is. I am partial to a configuration file that's loaded client side. This allows us to give the user the option to easily set the values without too much extra code. Using the `.env` will hard code any of those configuration values requiring us to add some extra wrapper code to allow for overrides...

With all that said, I think that the `.env` approach will work until it doesn't and we can tackle that then :-) I'll do it in a follow-up.

PR #20, opened 2019-08-12
Add Jest snapshots to existing components

barrymcgee:
## Done

Add Jest snapshots to existing components 📸

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Run `./run test`
- Verify no errors


src/components/ErrorBoundary/ErrorBoundary.test.js
```diff
@@ -1,16 +1,14 @@
...
 describe("Error Boundary", () => {
-  it("generates a error message when an error is caught", () => {
-    const wrapper = shallow(<ErrorBoundary />);
-    wrapper.setState({
-      hasError: true
-    });
-    expect(wrapper.text()).toEqual(
-      "Error: Something has gone wrong. If this issue persists, please raise an issue on GitHub."
-    );
+  const wrapper = shallow(<ErrorBoundary />);
```
hatched: This should be within an `it`.

jest.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  snapshotSerializers: ["enzyme-to-json/serializer"]
```
hatched: Thanks for moving this here.

src/components/Nav/Nav.test.js
```diff
@@ -1,15 +1,16 @@
...
+import Nav from "./Nav";
 
 describe("Nav", () => {
-  it("renders without crashing", () => {
-    shallow(
-      <Router>
-        <Nav />
-      </Router>
-    );
+  const wrapper = mount(
```
hatched: Can you move this within the `it`

src/components/ErrorBoundary/ErrorBoundary.test.js
```diff
@@ -1,16 +1,14 @@
 import React from "react";
-import ErrorBoundary from "./ErrorBoundary";
-
 import { shallow } from "enzyme";
 
+import ErrorBoundary from "./ErrorBoundary";
+
 describe("Error Boundary", () => {
-  it("generates a error message when an error is caught", () => {
+  it("renders without crashing and matches snapshot", () => {
```
hatched: < pedantic >You don't typically include how you're testing in the title. Just the what< /pedantic > ;-)

src/components/ErrorBoundary/__snapshots__/ErrorBoundary.test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Error Boundary renders without crashing and matches snapshot 1`] = `ShallowWrapper {}`;
```
hatched: You can see here that this test doesn't actually work. As you're effectively testing nothing. Same as below.

src/components/ErrorBoundary/ErrorBoundary.test.js
```diff
@@ -1,16 +1,14 @@
...
     const wrapper = shallow(<ErrorBoundary />);
     wrapper.setState({
       hasError: true
     });
-    expect(wrapper.text()).toEqual(
-      "Error: Something has gone wrong. If this issue persists, please raise an issue on GitHub."
-    );
+    expect(wrapper).toMatchSnapshot();
```
hatched: For this to work properly, you may have to put it in the callback for `setState` and then make this test async by adding `done`. See comments below about the snapshot not working.

squidsoup: This test would be better if it simulated an error condition for an error boundary, rather than asserting that setting the internal state does what you expect.

You can use `wrapper.find(ChildComponent).simulateError(error);` to simulate a child component throwing an error, and then check that the ErrorBoundary renders appropriately.

hatched: Ignore my comment about the `setState` stuff, I was going down the wrong rabbit hole.

@barrymcgee I believe that we discussed what @squidsoup is proposing. He has said it much more eloquently and should probably be done that way 😺

hatched:
@barrymcgee I've resolved the issue where the snapshots weren't being generated properly, so you'll want to pull your branch here before continuing to work on it 👍

src/components/ErrorBoundary/ErrorBoundary.test.js
barrymcgee: done - throwing an error means Jest test console spits out the stack trace every time though. Could get noisy over time? Jest has a `verbose` config setting but it's not supported by CRA overrides. 
Screenshot 2019-08-13 at 12 07 28

hatched:
Thanks @barrymcgee 

I've created an issue with CRA about the `verbose` configuration option

PR #22, opened 2019-08-14
Update secondary nav to match designs

barrymcgee:
## Done

- Abstract secondary nav out to a component
- Hard-code `is-selected` style until routing is implemented
- Rename `<Nav>` to `<PrimaryNav>`
- Add external links from Vanilla so icons show in Primary Nav

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify everything is as expected

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/799

### DIFF
--- a/src/components/SecondaryNav/_secondary-nav.scss
+++ b/src/components/SecondaryNav/_secondary-nav.scss
@@ -8,8 +8,8 @@
 
     &.is-selected {
       font-weight: 400;
-      border-left: 5px solid #335280;
-      padding-left: calc(1rem - 5px);
+      border-left: 3px solid #335280;
+      padding-left: calc(1rem - 3px);
     }
   }
 }
--- a/src/components/SecondaryNav/_secondary-nav.scss
+++ b/src/components/SecondaryNav/_secondary-nav.scss
@@ -1,0 +1,15 @@
+.p-secondary-nav {
+  .p-list__link {
+    color: #111;
+  }
+
+  .p-list__item {
+    padding-left: 1rem;
+
+    &.is-selected {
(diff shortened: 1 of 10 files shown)


src/components/SecondaryNav/_secondary-nav.scss
```diff
@@ -0,0 +1,15 @@
+.p-secondary-nav {
+  .p-list__link {
+    color: #111;
```
barrymcgee: I'm aware this should be a Vanilla variable but they cannot currently be accessed. Issue raised: https://github.com/canonical-web-and-design/jaas-dashboard/issues/13

barrymcgee:
@spencerbygraves Any ideas yet on how this secondary nav might look on small screens?

spencerbygraves:
@barrymcgee no, not at the moment.

src/components/SecondaryNav/_secondary-nav.scss
huwshimi: In maas-ui and the old gui we import the Vanilla bits we need at the top of the individual files, in this case that would be `@import "node_modules/vanilla-framework/scss/settings_colors";`

barrymcgee: I've submitted a PR here: https://github.com/canonical-web-and-design/jaas-dashboard/pull/23

Managed to simplify the import to `"@import "vanilla-framework/scss/vanilla";"` which can be used uniformly in all partials and allow all Vanilla entities to be referenced. As Vanilla is engineering so all patterns are wrapped in mixins, only what is referenced is included so no performance overhead.

PR #23, opened 2019-08-15
Fix vanilla imports

### LINKED ISSUE #13: Cannot access Vanilla variable in Sass partials
When splitting off individual Sass files for performance - those files do not currently inherit Vanilla so we cannot use Sass variables.

barrymcgee:
## Done

Fix vanilla imports by adding env variable to `node_modules` folder and adding `@import "vanilla-framework/scss/colors_settings";` to each SCSS partial.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify no errors in console
- Verify page loads correctly

## Details

Fixes #13

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
 REACT_APP_CONTROLLER_URL=wss://jimm.jujucharms.com/api
+SASS_PATH=node_modules:src
--- a/.env
+++ b/.env
@@ -1,2 +1,4 @@
+# API host to allow app to connect and retrieve models
 REACT_APP_CONTROLLER_URL=wss://jimm.jujucharms.com/api
+# Sass path to allow Sass to reference NPM packages directly
 SASS_PATH=node_modules:src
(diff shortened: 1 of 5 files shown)


barrymcgee:
Found a Vanilla bug, raised here: https://github.com/canonical-web-and-design/vanilla-framework/issues/2487

.env
```diff
@@ -1 +1,2 @@
 REACT_APP_CONTROLLER_URL=wss://jimm.jujucharms.com/api
+SASS_PATH=node_modules:src
```
hatched: Can we have comments in this file? It would be nice to add a comment as to where these are used.

src/components/SecondaryNav/_secondary-nav.scss
```diff
@@ -1,14 +1,16 @@
+@import "vanilla-framework/scss/settings_colors";
```
hatched: Can you confirm that sass dedupes these?

barrymcgee: Yes, confirmed.

PR #25, opened 2019-08-17
Add model list into the redux store

hatched:
## Done

Add model list into the redux store

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click the link in the console to log in
- After a few moments check the redux console you should have the models in state.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,9 +9,23 @@ import App from "./App";
 import rootReducer from "./reducers/root";
 import * as serviceWorker from "./serviceWorker";
 
-import { loginWithBakery } from "./juju";
+import { fetchModelList, loginWithBakery } from "./juju";
 
-loginWithBakery();
+async function connectAndListModels() {
+  try {
+    const conn = await loginWithBakery();
+    const modelList = await fetchModelList(conn);
+    // Store models in the redux store.
+    console.log(modelList.userModels);
+  } catch (error) {
+    // XXX Surface error to UI.
+    // XXX Send to sentry.
+    console.log("Something went wrong: ", error);
+    return;
+  }
+}
+
+connectAndListModels();
 
 const store = createStore(rootReducer);
 const rootElement = document.getElementById("root");
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,9 @@
 import React from "react";
 import ReactDOM from "react-dom";
 import { Provider } from "react-redux";
-import { createStore } from "redux";
+import { applyMiddleware, combineReducers, createStore } from "redux";
+import thunk from "redux-thunk";
+import { composeWithDevTools } from "redux-devtools-extension";
 
 import "./scss/index.scss";
 
@@ -10,13 +12,20 @@ import rootReducer from "./reducers/root";
 import * as serviceWorker from "./serviceWorker";
 
 import { fetchModelList, loginWithBakery } from "./juju";
+import jujuReducers from "./juju/reducers";
 
-async function connectAndListModels() {
+const reduxStore = createStore(
+  combineReducers({
+    root: rootReducer,
+    juju: jujuReducers
+  }),
+  composeWithDevTools(applyMiddleware(thunk))
+);
+
+async function connectAndListModels(reduxStore) {
   try {
     const conn = await loginWithBakery();
-    const modelList = await fetchModelList(conn);
-    // Store models in the redux store.
-    console.log(modelList.userModels);
+    reduxStore.dispatch(fetchModelList(conn));
   } catch (error) {
     // XXX Surface error to UI.
     // XXX Send to sentry.
@@ -25,13 +34,12 @@ async function connectAndListModels() {
   }
 }
 
-connectAndListModels();
+connectAndListModels(reduxStore);
 
-const store = createStore(rootReducer);
 const rootElement = document.getElementById("root");
 
 ReactDOM.render(
-  <Provider store={store}>
+  <Provider store={reduxStore}>
     <React.StrictMode>
       <App />
     </React.StrictMode>
--- a/src/index.js
+++ b/src/index.js
@@ -11,8 +11,10 @@ import App from "./App";
 import rootReducer from "./reducers/root";
 import * as serviceWorker from "./serviceWorker";
 
-import { fetchModelList, loginWithBakery } from "./juju";
+import { loginWithBakery } from "./juju";
 import jujuReducers from "./juju/reducers";
+import { fetchModelList } from "./juju/actions";
+import { actionsList } from "./reducers/actions";
 
 const reduxStore = createStore(
   combineReducers({
@@ -25,7 +27,11 @@ const reduxStore = createStore(
 async function connectAndListModels(reduxStore) {
   try {
     const conn = await loginWithBakery();
-    reduxStore.dispatch(fetchModelList(conn));
+    reduxStore.dispatch({
+      type: actionsList.updateControllerConnection,
+      payload: conn
+    });
+    reduxStore.dispatch(fetchModelList());
   } catch (error) {
     // XXX Surface error to UI.
     // XXX Send to sentry.
--- a/src/index.js
+++ b/src/index.js
@@ -24,6 +24,7 @@ const reduxStore = createStore(
   composeWithDevTools(applyMiddleware(thunk))
 );
 
+// eslint-disable-next-line no-shadow
 async function connectAndListModels(reduxStore) {
   try {
     const conn = await loginWithBakery();
@@ -35,8 +36,8 @@ async function connectAndListModels(reduxStore) {
   } catch (error) {
     // XXX Surface error to UI.
     // XXX Send to sentry.
+    // eslint-disable-next-line no-console
     console.log("Something went wrong: ", error);
-    return;
   }
 }
 
--- a/src/juju/actions.js
+++ b/src/juju/actions.js
@@ -20,7 +20,7 @@ export function updateModelList(models) {
   @returns {Object} models The list of model objects under the key `userModels`.
 */
 export function fetchModelList() {
-  return async function(dispatch, getState) {
+  return async function thunk(dispatch, getState) {
     const conn = getState().root.controllerConnection;
     const modelManager = conn.facades.modelManager;
     const models = await modelManager.listModels({ tag: conn.info.identity });
--- a/src/juju/actions.js
+++ b/src/juju/actions.js
@@ -1,0 +1,12 @@
+// Action labels
+export const actionsList = {
+  updateModelList: "UPDATE_MODEL_LIST"
+};
+
+// Action creators
+export function updateModelList(models) {
+  return {
+    type: actionsList.updateModelList,
+    models
+  };
+}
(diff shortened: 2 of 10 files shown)


src/juju/index.js
```diff
@@ -15,18 +17,28 @@ const options = {
...
+  return async function(dispatch) {
     const modelManager = conn.facades.modelManager;
     const models = await modelManager.listModels({ tag: conn.info.identity });
-    console.log("models", models);
-  } catch (error) {
-    // XXX Surface error to UI.
-    console.log("unable to connect:", error);
-    return;
-  }
+    dispatch(actions.updateModelList(models.userModels));
```
hatched: I'm wondering if maybe the reducer should get the raw data from the controller? 

ie) `dispatch(actions.updateModelList(models));`

src/juju/actions.js
```diff
@@ -0,0 +1,12 @@
+// Action labels
+export const actionsList = {
+  updateModelList: "UPDATE_MODEL_LIST"
+};
+
+// Action creators
+export function updateModelList(models) {
```
hatched: Should action creators be tested?

src/juju/reducers.js
```diff
@@ -0,0 +1,24 @@
+import produce from "immer";
+
+import { actionsList } from "./actions";
+
+/* eslint-disable default-case */
+// immer handles the default case so one isn't needed.
+export default function jujuReducers(state = {}, action) {
```
hatched: Does this seem like a good architecture for this? I'll add the test if this approach is 👍

src/juju/index.js
```diff
@@ -15,18 +17,28 @@ const options = {
 
 const controllerURL = process.env.REACT_APP_CONTROLLER_URL;
 
+/**
+  Connects to the controller at the url defined in the REACT_APP_CONTROLLER_URL
+  environment variable.
+  @returns {Object} conn The controller connection instance.
+*/
```
barrymcgee: As per team practices on [commenting code](https://canonical-web-and-design.github.io/practices/coding/javascript.html#commenting-code), we should not use JSdocs formatting unless for library code. I know you're probably not going to be happy about this but using it here and not elsewhere will leave the codebase inconsistent.

hatched: I wholly disagree with that statement. But if you'd like, this is effectively a library that the main application uses to communicate with jujulib via redux mechanisms 😄

src/juju/actions.js
squidsoup: Yep, there are [some examples](https://redux.js.org/recipes/writing-tests#example) in the docs.

src/juju/index.js
squidsoup: > As per team practices on commenting code, we should not use JSdocs formatting unless for library code

That's not what practices says - it says JSDoc should be used for library code (an important requirement for generating API docs), but doesn't stipulate whether it can or should be used elsewhere. We use JSdoc all over the place in maas-ui and crbs-ui for any sufficiently complex function. I don't see how *more* documentation could ever be a bad thing.

src/juju/index.js
```diff
@@ -15,18 +17,28 @@ const options = {
...
+
+/**
+  Fetches the model list from the supplied Juju controller.
+  @param {Object} conn The controller connection instance.
+  @returns {Object} models The list of model objects under the key `userModels`.
+ */
+function fetchModelList(conn) {
+  return async function(dispatch) {
     const modelManager = conn.facades.modelManager;
     const models = await modelManager.listModels({ tag: conn.info.identity });
```
squidsoup: If you're able to keep your "main" react app ignorant of the modelManager, that would probably be ideal, using redux actions as your interface boundary. Could you instead dispatch a fetch/list action which calls `modelManager.listModels` (ideally in middleware), with a corresponding reducer to update state?

src/juju/reducers.js
squidsoup: I think my only advice around modeling state in redux is to try to keep state as flat and close to the API as possible, and do all your transformations/clever stuff in selectors.

src/juju/index.js
squidsoup: Not sure what you mean here, maybe let's chat about it on irc.

src/juju/reducers.js
```diff
@@ -0,0 +1,24 @@
...
+        const modelList = action.models.map(model => {
+          return {
+            lastConnection: model.lastConnection,
+            name: model.model.name,
+            ownerTag: model.model.ownerTag,
+            type: model.model.type,
+            uuid: model.model.uuid
+          };
+        });
+        draftState.models = modelList;
```
squidsoup: I would call this `draftState.models.items` as you'll often want properties on the top level of state like `loading`, `loaded`, `error` etc.

src/index.js
```diff
@@ -1,23 +1,45 @@
...
+    root: rootReducer,
+    juju: jujuReducers
+  }),
+  composeWithDevTools(applyMiddleware(thunk))
+);
+
+async function connectAndListModels(reduxStore) {
+  try {
+    const conn = await loginWithBakery();
+    reduxStore.dispatch(fetchModelList(conn));
```
squidsoup: `dispatch` is typically called with an action creator which returns a standard flux action. It is a bit unusual to dispatch a regular function here which in turn dispatches another redux action. Could you instead dispatch an action from fetchModelList which performs the API call in middleware or in the action creator, and then responds with a SUCCESS action of some sort (settings some state so the UI knows the fetch has completed).

src/juju/index.js
```diff
@@ -15,18 +17,28 @@ const options = {
...
+  const conn = await juju.login({});
+  return conn;
+}
+
+/**
+  Fetches the model list from the supplied Juju controller.
+  @param {Object} conn The controller connection instance.
+  @returns {Object} models The list of model objects under the key `userModels`.
+ */
+function fetchModelList(conn) {
```
squidsoup: Should this live in `actions`?

src/index.js
```diff
@@ -1,23 +1,45 @@
...
+  } catch (error) {
```
squidsoup: Related comment below, but rather than wrapping `dispatch` in a try catch, you should probably do your error handing in your thunk.

src/juju/actions.js
```diff
@@ -0,0 +1,12 @@
...
+  return {
+    type: actionsList.updateModelList,
+    models
```
squidsoup: In standard flux actions this would be:
```
{
   type: actionsList.updateModelList,
   payload: models
}

File: src/juju/actions.js [hunk 11]
```diff
@@ -0,0 +1,12 @@
+// Action labels
+export const actionsList = {
+  updateModelList: "UPDATE_MODEL_LIST"
+};
```
squidsoup: nice

src/index.js
hatched: I suppose I could move the dispatch out of the try/catch but the try/catch is actually for the `await`'d function.

src/juju/actions.js
hatched: done

src/juju/reducers.js
hatched: done

src/juju/reducers.js
hatched: I suppose it would depend on how you'll be accessing that data. If you're always going to be accessing it in a specific format, you might as well store it in that format instead of re-creating that format on every request.

In this case we will be presenting this model list data to the tables to render in its entirety. I feel that a selector would be used if it was a subset of data or not the primary use case for it.

src/juju/index.js
barrymcgee: @squidsoup That discussion came about as it was pointed out that 'JSdoc all over the place ' bear no relation to the code that follows them as they were not being kept updated.

I don't disagree with "more documentation is always good" but imo using JSDocs format anywhere gives the impression that comprehensive documentation could be generated when that is not the case.

src/juju/actions.js
```diff
@@ -0,0 +1,29 @@
...
+  };
+}
+
+/**
+  Fetches the model list from the supplied Juju controller.
+  @returns {Object} models The list of model objects under the key `userModels`.
+*/
+export function fetchModelList() {
+  return async function(dispatch, getState) {
+    const conn = getState().root.controllerConnection;
```
squidsoup: You might find these posts interesting:

* https://blog.isquaredsoftware.com/2017/01/idiomatic-redux-thoughts-on-thunks-sagas-abstraction-and-reusability/
* https://stackoverflow.com/questions/35667249/accessing-redux-state-in-an-action-creator/35674575#35674575

Generally using getState in thunks is considered an anti-pattern apparently (and certainly is in an action creator), but I think this meets the criteria Dan mentions in the SO post, injecting some dependent cached data, in this case a connection.

src/juju/reducers.js
```diff
@@ -0,0 +1,24 @@
...
+        const modelList = action.payload.userModels.map(model => {
+          return {
+            lastConnection: model.lastConnection,
+            name: model.model.name,
+            ownerTag: model.model.ownerTag,
+            type: model.model.type,
+            uuid: model.model.uuid
+          };
+        });
+        draftState.models = { items: modelList };
```
squidsoup: Might be worth providing a default for draftState.models.items (you can do this by passing an object to the second param of `produce`, e.g.

```
produce(
  (draft, action) => {
    switch (action.type) {
     ...
    },
  },
  {
    items: []
  }
);
```

then you could just say `draftState.models.items = modelList` here.

src/juju/actions.js
hatched: Yep I agree, thanks for the links, some good content there!

src/index.js
```diff
@@ -10,14 +12,41 @@ import rootReducer from "./reducers/root";
...
-loginWithBakery();
+const reduxStore = createStore(
+  combineReducers({
+    root: rootReducer,
+    juju: jujuReducers
+  }),
+  composeWithDevTools(applyMiddleware(thunk))
+);
+
+// eslint-disable-next-line no-shadow
```
barrymcgee: What's the reason for this exception?

src/juju/index.js
```diff
@@ -15,19 +15,15 @@ const options = {
...
-    const models = await modelManagerFacades.listModels({
-      tag: conn.info.identity
-    });
-    console.log("models", models); // eslint-disable-line no-console
-  } catch (error) {
-    // XXX Surface error to UI.
-    console.log("unable to connect:", error); // eslint-disable-line no-console
-  }
+  const juju = await jujulib.connect(controllerURL, options);
+  const conn = await juju.login({});
```
barrymcgee: Shouldn't the result of `jujulib.connect()` be assigned to a variable called `conn` (assuming it's short for connection?) ...and the result of `juju.login({})` might be better described by assigning to `const auth` or similar?

src/index.js
hatched: Ahh, because I have a variable called `reduxStore` in the `connectAndListModels` and one in the app scope. It's not really a shadow because the app scoped one is defined by const and cannot be overridden without an error so I added the exception.

src/juju/index.js
hatched: Yeah that would make a bit more sense, however the instance returned by `login` is the 'connection' instance so that's why it's named that way 🤷‍♂

src/juju/reducers.js
hatched: This didn't work in practice, we can revisit this in a follow-up as it would be a nice feature to have

hatched:
Thanks all for the reviews/qa!

PR #34, opened 2019-08-21
Fetch model statuses and store in Redux

hatched:
## Done

Fetch the model status for the list of models returned by modelList and then store that data in the Redux under the model.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click the link to log into the controller and then switch to redux devtools
- You will see the model status values being updated.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,7 @@ import App from "./components/App/App";
 import rootReducer from "./reducers/root";
 import * as serviceWorker from "./serviceWorker";
 
-import { loginWithBakery } from "./juju";
+import { connectAndFetchModelStatus, loginWithBakery } from "./juju";
 import jujuReducers from "./juju/reducers";
 import { fetchModelList } from "./juju/actions";
 import { actionsList } from "./reducers/actions";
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -27,12 +27,40 @@ const reduxStore = createStore(
 // eslint-disable-next-line no-shadow
 async function connectAndListModels(reduxStore) {
   try {
+    // eslint-disable-next-line no-console
+    console.log("Logging into the Juju controller.");
```
hatched: These console logs will go away once we have the UI updating.

src/index.js
```diff
@@ -27,12 +27,40 @@ const reduxStore = createStore(
...
+    // listModels to update the model lists.
+    const modelList = reduxStore.getState().juju.models.items;
+    // eslint-disable-next-line no-console
+    console.log("Fetching model statuses");
+
+    // Using a for loop here for performance reasons for users with many models.
+    // eslint-disable-next-line no-restricted-syntax
+    for (const modelUUID in modelList) {
+      if (!Object.prototype.hasOwnProperty.call(modelList, modelUUID)) {
+        // eslint-disable-next-line no-continue
```
hatched: I'm not sure why `no-continue` is enabled??

src/juju/reducers.js
```diff
@@ -2,21 +2,31 @@ import produce from "immer";
 
 import { actionsList } from "./actions";
 
-export default function jujuReducers(state = {}, action) {
-  return produce(state, draftState => {
+export default produce(
+  (draftState, action) => {
+    // The below is required because CRA ignores the eslint configuration file
+    // during development and adds warnings in the dev console.
+    // eslint-disable-next-line default-case
```
hatched: I'm open to suggestions on how to avoid those warnings without resorting to weird hacks like this.

barrymcgee: Add a blank default?

```
      case actionsList.updateModelStatus:
        const payload = action.payload;
        draftState.models.items[payload.modelUUID].status = payload.status;
        break;
      default: // do nothing.
```

hatched: It was then complaining about an empty default haha.

hatched: I could then add a console log and then an exception for that 😆

hatched:
Thanks!

PR #40, opened 2019-08-23
Use Maas <MainTable> component instead of <TableSet>

barrymcgee:
## Done

Use Maas `<MainTable>` component instead of `<TableSet>`

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify table is sortable by clicking column headers


src/testing/utils.js
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+/**
+ * Assert that some JSX from Enzyme is equal to some provided JSX.
+ * @param {Object} actual - Some JSX from Enzyme.
+ * @param {Object} expected - Some JSX provided in the test.
+ */
+export const compareJSX = (actual, expected) => {
```
hatched: Were these files just copied over? I am not sure why we'd use this vs snapshots?

src/testing/utils.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Replace objects in an array with objects that have new values, given a match
+ * criteria.
+ * @param {Array} array - Array to be reduced.
+ * @param {String} key - Object key to compare the match criteria e.g. "name".
+ * @param {String} match - Match criteria e.g. "Bob".
+ * @param {Object} newValues - Values to insert or update in the object.
+ * @returns {Array} The reduced array.
+ */
+export const reduceInitialState = (array, key, match, newValues) => {
```
hatched: I don't see this used anywhere...

src/testing/utils.js
barrymcgee: 17:30:37 <barrymcgee> Hatch yes, those tests were just copied over verbatim.. I spent a while trying to cherry pick only the bits I needed then figured if we’re gonna hopefully abstract to a lib later down the line, I shouldn’t diverge without good reason

PR #42, opened 2019-08-28
Display model data from Redux store.

hatched:
## Done

Display the model data in the table format from the data retrieved from the controller stored in the Redux store.

## QA

- You'll want a number of juju models with a various number of applications/units/machines before starting to QA.
- Pull code
- Open http://0.0.0.0:8036/
- Log in via the link in the browser console. 
- Now your model data should populate as it becomes available.

### DIFF
--- a/src/components/TableList/TableList.js
+++ b/src/components/TableList/TableList.js
@@ -1,0 +1,109 @@
+import React from "react";
+import { connect } from "react-redux";
+import PropTypes from "prop-types";
+
+import MainTable from "../MainTable/MainTable";
+
+function generateTableHeaders() {
+  return [
+    { content: "Name", sortKey: "name" },
+    { content: "Owner", sortKey: "owner" },
+    { content: "Summary", sortKey: "summary" },
+    { content: "Cloud", sortKey: "cloud" },
+    { content: "Region", sortKey: "region" },
+    { content: "Credential", sortKey: "credential" },
+    { content: "Controller", sortKey: "controller" },
+    { content: "Last Updated", sortKey: "last-updated" }
+  ];
+}
+
+function TableList({ modelData }) {
+  return (
+    <MainTable headers={generateTableHeaders()} rows={modelData} sortable />
+  );
+}
+
+TableList.propTypes = {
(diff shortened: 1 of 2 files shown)


src/components/TableList/TableList.js
```diff
@@ -0,0 +1,94 @@
+import React from "react";
+import { useSelector } from "react-redux";
+
+import MainTable from "../MainTable/MainTable";
+
+function generateTableHeaders() {
```
barrymcgee: If there is no way to generate these table headers, can you pass them as a prop? Hard coding them makes this component unusable for the Model Detail view.

src/components/TableList/TableList.js
```diff
@@ -0,0 +1,94 @@
...
+/**
+  Used to fetch the values from status as it won't be defined when the
+  modelInfo data is.
+  @param {Object|undefined} status The status for the model.
+  @param {String} key The key to fetch.
+  @returns {String} The computed value for the requested field if defined, or
+    an empty string.
+*/
+function getStatusValue(status, key) {
+  let returnValue = "";
```
barrymcgee:
```suggestion
  let returnValue;
```

src/components/TableList/TableList.js
```diff
@@ -0,0 +1,94 @@
...
+          " machines, " +
+          unitCount +
+          " units";
+        break;
+      case "cloudTag":
+        returnValue = status.model.cloudTag.replace("cloud-", "");
+        break;
+      case "region":
+        returnValue = status.model.region;
+        break;
```
barrymcgee:
```suggestion
        break;
      default:
        returnValue = "Err!";
```

Eslint is flagging these missing default cases, we should add them.

src/components/TableList/TableList.js
barrymcgee: Good to know.

src/components/TableList/TableList.js
barrymcgee: It's passing but throwing warnings?

Screenshot 2019-08-28 at 13 33 23

PR #49, opened 2019-09-05
Add tests for model list table

hatched:
## Done

Add tests for the model list table and add mock data for testing with the redux store.

## QA

- CI should pass

### DIFF
--- a/src/testing/complete-redux-store-dump.js
+++ b/src/testing/complete-redux-store-dump.js
@@ -1,0 +1,2811 @@
+// Internally consistent mock data.
+// When generating additional mock data make sure that it's internally consistent
+// across the top level keys.
+
+export default {
+  root: {
+    controllerConnection: {
+      transport: {
+        _ws: {},
+        _counter: 39,
+        _callbacks: {},
+        _debug: true
+      },
+      info: {
+        controllerTag: 'controller-a030329a-941f-4760-8fcd-3063b41a04e7',
+        serverVersion: '2.6.5',
+        servers: [],
+        user: {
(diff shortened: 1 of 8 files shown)


src/testing/complete-redux-store-dump.js
```diff
@@ -0,0 +1,2811 @@
...
+            instanceStatus: {
+              status: 'running',
+              info: 'RUNNING',
+              data: {},
+              since: '2017-03-09T20:40:24.303597554Z',
+              kind: '',
+              version: '',
+              life: ''
+            },
+            dnsName: '104.196.135.3',
```
ajbeattie: 104.196.135.3 happens to be a publicly visible Grafana :)

hatched: Hah, thanks!

src/testing/complete-redux-store-dump.js
```diff
@@ -0,0 +1,2811 @@
...
+            instanceStatus: {
+              status: 'running',
+              info: 'RUNNING',
+              data: {},
+              since: '2016-12-12T17:36:03.774497893Z',
+              kind: '',
+              version: '',
+              life: ''
+            },
+            dnsName: '104.196.205.152',
```
ajbeattie: FWIW this seems to resolve but throws a 404 🤷‍♀

hatched: Hah, yeah I just randomly typed in numbers for the uuid's and ip's. I suppose there is always bound to be something real on the other end (especially for ipv4)

PR #54, opened 2019-09-11
Link models through to details pages

barrymcgee:
## Done

- Clean up `window.location` mocking
- Wrap models details in `<Layout>`
- Link models through to details pages

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code


src/components/TableList/TableList.test.js
```diff
@@ -31,9 +35,11 @@ describe("TableList", () => {
   it("displays all data from redux store", () => {
     const store = mockStore(dataDump);
     const wrapper = mount(
-      <Provider store={store}>
-        <TableList />
-      </Provider>
+      <Router history={newHistory}>
```
hatched: What's going on with this change? Does this have to be here for a valid test? If so then it'll need to go on every test, meaning we will need to create a 'const wrapper = generateComponent(options)` method that can be shared between tests.

PR #55, opened 2019-09-12
Add terminal component to model details view.

hatched:
## Done

The model details view is now gated on logging in.
After logging in the details view is rendered along with the terminal component which auto-connects to the JAAS limited terminal.

## QA

- Pull code
- Open http://0.0.0.0:8036/models/foo
- You should get a 'log in message'
- Open the browser console and log in
- The model details page should render
- Wait for the terminal to connect and become functional.

## Details

The terminal is a small box in the lower left corner of the page because the markup and styling needs to be adjusted. That's out of scope for this branch and will be done as part of https://github.com/canonical-web-and-design/juju-squad/issues/889

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/861
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/863


package.json
```diff
@@ -41,9 +41,13 @@
   "jest": {
     "snapshotSerializers": [
       "enzyme-to-json/serializer"
+    ],
+    "transformIgnorePatterns": [
```
hatched: The `juju-react-components` package doesn't export a commonjs build so this needs to tell Jest to build them. We can remove this when we bring the Terminal component in to the project.

src/pages/Models/ModelDetails/ModelDetails.test.js
```diff
@@ -1,11 +1,35 @@
...
+    const store = mockStore({
+      root: {},
+      juju: dataDump.juju
+    });
+    const wrapper = mount(
+      <Provider store={store}>
+        <ModelDetail />
+      </Provider>
+    );
+    expect(wrapper).toMatchSnapshot();
```
huwshimi: These snapshots are probably not that useful when using `mount()`. Probably better to check that a few things exist.

At the very least you should do something like `expect(wrapper.find("ModelDetail")).toMatchSnapshot();` so that you're not including the provider in the snapshot.

hatched:
Thanks @huwshimi

PR #56, opened 2019-09-12
Add <Suspense> around main content area

barrymcgee:
## Done

- Add `<Suspense>` around main content area. This will display a loading component until all lazy-loaded components in the tree beneath it have loaded. 
- Wrap `<ErrorBoundary>` around the main content area so that if a component fails, it'll not lock up the entire app but leave the navigation still visible so other sections can still be accessed.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Refresh manically to verify a loading spinner

### DIFF
--- a/src/components/App/App.js
+++ b/src/components/App/App.js
@@ -1,30 +1,31 @@
-import React from "react";
+import React, { Suspense } from "react";
 import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
 
-// Components
-import ErrorBoundary from "components/ErrorBoundary/ErrorBoundary";
+import Loading from "components/Loading/Loading";
 
 // Pages
-import Controllers from "pages/Controllers/Controllers";
-import Logs from "pages/Logs/Logs";
-import Models from "pages/Models/Models";
-import ModelDetails from "pages/Models/Details/ModelDetails";
-import NotFound from "pages/NotFound/NotFound";
-import Usage from "pages/Usage/Usage";
+const Controllers = React.lazy(() => import("pages/Controllers/Controllers"));
+const Logs = React.lazy(() => import("pages/Logs/Logs"));
+const Models = React.lazy(() => import("pages/Models/Models"));
+const ModelDetails = React.lazy(() =>
+  import("pages/Models/Details/ModelDetails")
+);
+const NotFound = React.lazy(() => import("pages/NotFound/NotFound"));
+const Usage = React.lazy(() => import("pages/Usage/Usage"));
 
 function App() {
   return (
     <Router>
-      <ErrorBoundary>
-        <Switch>
+      <Switch>
+        <Suspense fallback={<Loading />}>
           <Route path="/" exact component={Models} />
           <Route path="/models/:id" exact component={ModelDetails} />
           <Route path="/controllers" exact component={Controllers} />
           <Route path="/usage" exact component={Usage} />
           <Route path="/logs" exact component={Logs} />
           <Route component={NotFound} />
-        </Switch>
-      </ErrorBoundary>
+        </Suspense>
+      </Switch>
     </Router>
   );
 }
--- a/src/components/App/App.js
+++ b/src/components/App/App.js
@@ -16,16 +16,16 @@ const Usage = React.lazy(() => import("pages/Usage/Usage"));
 function App() {
   return (
     <Router>
-      <Switch>
-        <Suspense fallback={<Loading />}>
+      <Suspense fallback={<Loading />}>
+        <Switch>
           <Route path="/" exact component={Models} />
           <Route path="/models/:id" exact component={ModelDetails} />
           <Route path="/controllers" exact component={Controllers} />
           <Route path="/usage" exact component={Usage} />
           <Route path="/logs" exact component={Logs} />
           <Route component={NotFound} />
-        </Suspense>
-      </Switch>
+        </Switch>
+      </Suspense>
     </Router>
   );
 }
--- a/src/components/App/App.js
+++ b/src/components/App/App.js
@@ -1,6 +1,7 @@
 import React, { Suspense } from "react";
 import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
 
+import Layout from "components/Layout/Layout";
 import Loading from "components/Loading/Loading";
 
 // Pages
@@ -16,16 +17,18 @@ const Usage = React.lazy(() => import("pages/Usage/Usage"));
 function App() {
   return (
     <Router>
-      <Suspense fallback={<Loading />}>
-        <Switch>
-          <Route path="/" exact component={Models} />
-          <Route path="/models/:id" exact component={ModelDetails} />
-          <Route path="/controllers" exact component={Controllers} />
-          <Route path="/usage" exact component={Usage} />
-          <Route path="/logs" exact component={Logs} />
-          <Route component={NotFound} />
-        </Switch>
-      </Suspense>
+      <Layout>
+        <Suspense fallback={<Loading />}>
+          <Switch>
+            <Route path="/" exact component={Models} />
+            <Route path="/models/:id" exact component={ModelDetails} />
+            <Route path="/controllers" exact component={Controllers} />
+            <Route path="/usage" exact component={Usage} />
+            <Route path="/logs" exact component={Logs} />
+            <Route component={NotFound} />
+          </Switch>
+        </Suspense>
+      </Layout>
     </Router>
   );
 }
--- a/src/components/App/__snapshots__/App.test.js.snap
+++ b/src/components/App/__snapshots__/App.test.js.snap
@@ -2,81 +2,85 @@
 
 exports[`App renders without crashing and matches snapshot 1`] = `
 <BrowserRouter>
-  <Switch>
+  <Layout
+    sidebar={true}
+  >
     <Suspense
       fallback={<Loading />}
     >
-      <Route
-        component={
-          Object {
-            "$$typeof": Symbol(react.lazy),
-            "_ctor": [Function],
-            "_result": null,
-            "_status": -1,
+      <Switch>
+        <Route
+          component={
+            Object {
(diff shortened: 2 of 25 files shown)


hatched:
@anthonydillon sure thing!

After lots of thought about this I’ve come to the following conclusion. Sorry, it’s long :-)

#### Network Performance Status

We don’t currently have any performance issues related to fetching the JS assets over the wire as it stands. The largest time waiting, by a large margin, is the communications with the controller. Introducing this Suspense feature now does not improve performance and actually increases the variance in the time-to-load because it has to load additional files. This increases the chance of a slow network connection blocking the application from loading as seen by the following tests.

Current: 17 requests1.2 MB transferred4.6 MB resourcesFinish: [767, 731, 717, 742, 718] ms
Suspense: 23 requests1.2 MB transferred3.8 MB resourcesFinish: [730, 710, 830, 973, 888] ms

In these tests the time-to-interactive carried similar deltas. 

The data transferred is the same after gzip but the unzipped value is quite large as these are unoptimized dev builds. So that means the potential increase in network performance will be even smaller. Keeping in mind that the suspense library code will add some (yet undetermined) amount of code that it will have to load.

#### User experience

It’s important to know whether you want to optimize for the initial page load, or for a more fluid experience while you’re in the application. A quick way to answer this is by how you expect the user to be using the application (ideally you’d use real data for this decision). Are they going to be loading the application up and then sitting there using it for some time, so the load time is an insignificant amount of time based on the total use, or are they effectively shopping around and will quickly lose patience if the information isn’t available right away. I feel like we will be strongly in the former position.

With this in mind, the change to use Suspense introduces a regression to the goal of a fluid application experience. By requiring that it load every top level component when the user clicks to view it the first time means that, no matter how fast the connection, it still flashes the loading spinner between each section as it loads instead of just showing the content immediately. 

#### Login & controller connection spinner

Suspense currently only supports showing the loading spinner while requesting components over the network. It does not provide a mechanism by which we can trigger it while we’re waiting to connect to and log into a controller or fetching data. We cannot programatically trigger it while refreshing stale data, or if the user has logged out and is in the process of logging back in. These types of things are currently under discussion but have no concrete release date, and may never support our use case, so will have to implement those our selves. 

#### Restricted UI/UX

Suspense requires a replacement of the component wholesale with some alternative. This means that it has the potential to limit our design space to provide feedback to the user about the status of a request or pending data loading, etc.

We’ll have to show two spinners, one for requesting the component over the wire, then another one once the component loads requesting the data from the component.

#### Added Complexity

It’s adding a layer of complexity to the application for, currently, no measurable gain.

Component authors will now have to consider the effects of lazy loading on their component lifecycles. One example, if a component needs to request additional data then that data cannot be fetched until the component is loaded over the wire the component will then have to trigger another spinner. This potentially means that the component is ready even later than if it was all bundled into one.

#### Juju / Third Party Server

Whenever we modify how assets are going to be served or data fetched for this application we have to consider how this may impact being served by other web servers outside of our direct control, such as the Juju GUI server. I haven’t yet investigated how this type of lazy loading would work with the existing server and what changes may need to be made. 

#### A Panacea

Assuming that in the future we find that certain sections of the application are going mostly unused, or that one section acts as the primary landing page for most users. We can, and should, implement some type of a lazy loading (Suspense) mechanism by which we ship the most important content to the user in the initial bundle(s) but then push the lesser used content when the network connection becomes idle and the application is operational, avoiding the dreaded “I clicked something but nothing is happening” feeling. This will give us the best of both worlds. The initial page loads will be kept in check while pre-loading lesser used content to keep the application experience as fluid as possible. This is not currently possible with Suspense and there hasn’t been any documentation alluding to this functionality being in the pipeline.

#### Summary

I appreciate the time and effort @barrymcgee has put into trying to make this work and resolve my objections. However, it currently doesn’t provide a decrease in page load time, it creates a more jarring application experience, we still have to implement our own loading UI to indicate to the user when we’re connecting to the controller, logging in, fetching data, etc, and we don’t have enough data to make a knowledgable decision as to if this will increase user satisfaction while it guarantees additional complexity in the application.

While I’m not against this approach in general to increase perceived performance (See `A Panacea`) I feel that the move to realize any performance improvements is premature as is the tooling.

anthonydillon:
@hatched thank you for taking the time to explain the pros and cons from your point of view and they are good points.  

I was under the impression that Suspense would provide a programmatic way of waiting for the data to be loaded. Therefore, giving us a common way to switch view and load the data. As this appears to not be the case (@barrymcgee please clarify) then the only benefit is a smaller initial payload.

The speed in which the user can navigate the application is paramount. I think we should keep the codebase as simple as possible until we have a problem to solve. Solving future problems tends to lead to the wrong solution or not resulting in an issue at all.

Personally my favourite loading mechanism is displaying a mock UI as it improves perceived performance. These mock UIs would need to be baked into each component that lists data. Therefore Suspense would restrict us from using the method of UX.

@barrymcgee please provide you pros and cons and let's discuss further if required.

barrymcgee:
> I was under the impression that Suspense would provide a programmatic way of waiting for the data to be loaded. Therefore, giving us a common way to switch view and load the data.

It will eventually but does not provide a fallback when loading data right now.  Suspense right now only provides code-splitting. Data-fetching is currently due to ship [before the end of the year.](https://reactjs.org/blog/2019/08/08/react-v16.9.0.html#an-update-to-the-roadmap)

> Solving future problems tends to lead to the wrong solution or not resulting in an issue at all.

We currently have an app that toggles one table and 4 empty divs. Benchmarking load times against the app as-is feels a little pointless. I agree with the sentiment of not prematurely solving problems we haven't got but given where we currently are, every architectural decision is made with one eye on where we're going to be. This sentiment is also outlined in the [official React docs on Suspense](https://reactjs.org/docs/react-api.html#suspense):  

Screenshot 2019-09-25 at 09 31 36

> These mock UIs would need to be baked into each component that lists data.

So, each component would have a custom loading component? That feels overkill to me. The loading fallback is a component like any other so can be styled, however. I'm sure we can think of an animation that's generic enough to replace all components while still conveying perceived performance.

> This means that it has the potential to limit our design space to provide feedback to the user about the status of a request or pending data loading, etc.

That _is_ solving a problem we don't have and afaik we have no way to send status info on a request back anyway. I imagine the most probable solution for that use case if it ever arises will be... data fetching with suspense.

> We can, and should, implement some type of a lazy loading (Suspense) mechanism by which we ship the most important content to the user in the initial bundle(s) but then push the lesser-used content when the network connection becomes idle and the application is operational, avoiding the dreaded “I clicked something but nothing is happening” feeling. This will give us the best of both worlds.

I think we have enough problems to solve without trying to better React. It's my view we should work with the framework where possible, utilise the tools as they become available and let the framework do the heavy lifting. It might not be perfect now but that's what version bumps are for.

hatched:
Please remember that the issue we’re looking to solve isn’t one based around application load time or network performance but of pending requests and fetching data which Suspense does not solve. We also do not know what the API for this feature will look like when it becomes available and as such cannot make a decision now that it'll meet our requirements. By your own admission this feature is premature and the statement you’ve snapshotted ignores the nuances in making such a decision.

> I think we have enough problems to solve without trying to better React. It's my view we should work with the framework where possible, utilise the tools as they become available and let the framework do the heavy lifting. It might not be perfect now but that's what version bumps are for.

This isn't about trying to 'better anything' it's simply about not using a tool that doesn't remedy the issue in question in hopes that a future version of it may.

I propose that we close this PR now to focus on pending indicators for data fetching and revisit its usefulness when the Suspense feature has been fully released.

barrymcgee:
> Please remember that the issue we’re looking to solve isn’t one based around application load time or network performance

Performance is and should be a continuous consideration.

> By your own admission this feature is premature 

The code splitting aspect of Suspense, the only aspect of Suspense this PR implements, is stable.

> the statement you’ve snapshotted ignores the nuances in making such a decision.

I included all three paragraphs from the docs related to code-splitting. I didn't write the React docs.

hatched:
> Performance is and should be a continuous consideration.

Certainly, however this has only shown to create a negative impact on performance in this project at this time.

squidsoup:
Our apps (jaas, maas etc.) are in the unusual position of typically having a relatively small number of _regular_, private users, in contrast to public facing applications like Facebook, Instagram etc. which I think does mean initial time to load is arguably not the most important metric for performance for us. While @barrymcgee is absolutely right that we should always be considering performance, this does feel like premature optimisation at this stage in the project's life, and @hatched has raised some valid concerns.

Given the Suspense API is still very young, perhaps it would be prudent to reassess this at a later date if we feel we have performance issues to merit its consideration?

Lukewh:
This is an interesting one. On the one hand I agree that Suspense isn't quite there yet (it seems the most valuable part of Suspense for all our projects will be the async loading - when it arrives), on the other hand I feel like getting familiar with the feature isn't a bad thing. 
This change seems so minimal that it could easily be backed out if it was deemed unnecessary. 

Another argument is - maybe having it lazy load components so early (routing) is overkill at this point, are there any parts of the app where there's some secondary information that can be loaded later? This approach would get the best of both worlds IMO, immediate loading of the routes with key information followed by lazy loading additional components.
Looking through the Zepelin some components come to mind: Juju Shell, Graphs, Filter component (the one @charlesjquinn et al. explored in Paris)?

anthonydillon:
@barrymcgee thank you for taking the time raise this discussion and it is something we should talk about as a team to produce guidelines on when to introduce Suspense into a project. 

This project is too premature to add code splitting. The difference between the chunks are very small. We also want to reuse a lot of the components such as tables, buttons, etc. Which would have to be in the initial chunk. Once we introduce large dependencies such as d3 the reward for code splitting will be worth revisiting. 

Unfortunately at this time I don't think this is the right approach. But, please do raise this topic on the mailing list to discuss the parameters in which Suspense should be introduced.

barrymcgee:
@Lukewh Lazy with preloading: https://blog.maximeheckel.com/posts/preloading-views-with-react

PR #67, opened 2019-10-02
Add auth wrapper components & Fetch model UUID on details page

hatched:
## Done

- When trying to view an authenticated page, if you've not already authenticated it'll now render a very basic text link for you to authenticate with usso.
- Get the model UUID and pass it to the Terminal component for use later.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Open / it should request that you log in.
- After logging in, click one of the models in your model list to show the details page.
- The page with dummy data should load as expected.

### DIFF
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -1,5 +1,6 @@
 /**
   Checks state to see if the user is logged in.
+  @param {Object} state The application state.
   @returns {Boolean} If the user is logged in.
 */
 export const isLoggedIn = state =>
@@ -7,6 +8,7 @@ export const isLoggedIn = state =>
 
 /**
   Pull the users macaroon credentials from state.
+  @param {Object} state The application state.
   @returns {Object} The macaroons extracted from the bakery in state.
 */
 export const getUserCredentials = state => {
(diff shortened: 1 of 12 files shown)


src/app/selectors.js
```diff
@@ -36,3 +38,32 @@ export const getDecodedMacaroons = macaroons => {
...
+
+/**
+  Gets the model UUID from the supplied name.
+  @param {String} name The name of the model.
+  @param {Object} modelInfo The list of models.
+  @returns {Object|Null} The model UUID or null if none found.
+*/
+export const getModelUUIDByName = (name, modelInfo) => {
+  if (modelInfo) {
+    for (let uuid in modelInfo) {
```
squidsoup: `filter` might be nicer.

src/components/IsLoggedIn/IsLoggedIn.js
```diff
@@ -0,0 +1,29 @@
...
+        return root.visitURL;
+      }
+    }
+  });
+
+  if (!userIsLoggedIn) {
+    return (
+      <div>
+        Please{" "}
+        <a href={visitURL} rel="noopener noreferrer" target="_blank">
```
squidsoup: This should be a vanilla link component.

src/app/selectors.js
hatched: I want to bail out as soon as it's found so `some` would be preferred, however `modelInfo` is an Object so can't filter over it. In order to use any of the array utilities I'd have to first do something like `Object.keys().some(key => { ... })` which wouldn't be as performant in the event that there is a significant number of models.

src/components/IsLoggedIn/IsLoggedIn.js
hatched: This is only a temporary page to make qa'ing and developing easier. There is a design for this page that's not this :D

PR #68, opened 2019-10-02
Switch the CLI to the model you're viewing

hatched:
## Done

Switch the CLI to the model you're viewing the details of.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Pick a model **you own** from the model list and click on it (there is a bug if you don't own the model)
- The terminal should switch to that model.


anthonydillon:
QA worked well as I had a few models knocking around.

hatched:
Thanks @anthonydillon

PR #71, opened 2019-10-08
Show the model status using real data

hatched:
## Done

Show the model status details on the model details page. 

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click on a model you own and it should show your model status.
- Navigate back to the model listing and click on a model you don't own. It should also show you the model status.

## Details

- Requires #70 to land first (contains code from #70).
- This is unstyled, that will be able to be done as a follow-up.

### DIFF
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -1,5 +1,32 @@
 import { createSelector } from "reselect";
 
+// ---- Selectors for top level keys
+/**
+  Fetches the model list from state.
+  @param {Object} state The application state.
+  @returns {Object|Null} The list of models or null if none found.
+*/
+const getModelList = state => {
+  if (state.juju && state.juju.models) {
+    return state.juju.models;
+  }
+  return null;
+};
+
+/**
+  Fetches the model statuses from state.
+  @param {Object} state The application state.
+  @returns {Object|Null} The list of model statuses or null if none found.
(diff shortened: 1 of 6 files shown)


hatched:
Good find @barrymcgee I've resolved the issue, it's ready for another qa.

barrymcgee:
Testing locally:

This is my list of models:

Screenshot 2019-10-09 at 10 01 19

One of my models `mongodb-test` continues to be empty. Even if this was expected, I'd want to see an error/info message instead of empty tables?

Screenshot 2019-10-09 at 09 59 48

The `monitor` and `test-model` models populate the first two tables but not the third? Is this because the third table represents relational data as it is shared?

If so, I'd expect this table to disappear for those models rather than render empty?

Screenshot 2019-10-09 at 10 01 56

The shared models _do_ populate all three tables.

Screenshot 2019-10-09 at 10 04 59

anthonydillon:
Instead of hiding tables I would suggest we should display a message in the empty table to the user.

clagom:
> Instead of hiding tables I would suggest we should display a message in the empty table to the user.

+1

clagom:
Btw this is generally what we do also on RBAC and other products, I think it's safe to assume it should be the default behaviour whenever we can

PR #73, opened 2019-10-10
Refactor login flow

barrymcgee:
## Done

- Rename `isLoggedIn` to `LogIn` for consistency
- Add basic styles to login screen
- Wrap all routes internal routes behind login gateway

## QA

- Pull code
- Run `./run clean && ./run serve`
- Open http://0.0.0.0:8036/
- Click through login flow and sanity check

PR #75, opened 2019-10-10
Add real data to Model detail info panel

barrymcgee:
## Done

Add real data to Model detail info panel

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to a models detail page
- Verify the info in the models detail info panel is correct


hatched:
I don't think this code actually relies at all on #73 branch so if you could rebase it on master it can be done independently.

barrymcgee:
It does rely on #73.

src/components/InfoPanel/InfoPanel.js
```diff
@@ -42,4 +36,20 @@ const InfoPanel = () => {
   );
 };
 
+InfoPanel.defaultProps = {
```
hatched: Instead of passing props in I think we'll want to stick with the `redux` data model here. That is, fetching the necessary data from a selector. This will give us the ability to modify the data displayed in this component without worrying about its parent. I can help you out doing this if you'd like.

src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -4,8 +4,25 @@ import { shallow } from "enzyme";
 import InfoPanel from "./InfoPanel";
 
 describe("Info Panel", () => {
+  const modelStatus = {
```
hatched: After making the above change to use the `selector` you can use the real 'mock' data which will allow us to keep our tests more robust as they are all using the same data set.

src/pages/Models/Details/__snapshots__/ModelDetails.test.js.snap
```diff
@@ -4,7 +4,823 @@ exports[`ModelDetail Container renders the details pane 1`] = `
 <div
   className="model-details"
 >
-  <InfoPanel>
+  <InfoPanel
```
hatched: I understand that this is just an update to an existing test but we had previously discussed the limited usefulness of snapshots that are this large. Would you mind updating the assertions in this branch to be more focused? Thanks!

src/pages/Models/Details/ModelDetails.js
```diff
@@ -157,7 +157,7 @@ const ModelDetails = () => {
   return (
     <Layout sidebar={false}>
       <div className="model-details">
-        <InfoPanel />
+        <InfoPanel modelStatus={modelStatusData} />
```
hatched: With the `redux` data flow change then this data will no longer be passed in as each component is responsible for fetching/requesting its own data.

src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -1,11 +1,43 @@
...
+
 describe("Info Panel", () => {
   it("renders without crashing and matches snapshot", () => {
-    const wrapper = shallow(<InfoPanel />);
-    expect(wrapper).toMatchSnapshot();
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter initialEntries={["/models/group-test"]}>
+          <Route path="/models/*">
```
hatched: Now that we're doing this multiple places maybe we should also create a test scaffold for defining these test routes. https://github.com/canonical-web-and-design/jaas-dashboard/issues/86

src/pages/Models/Details/ModelDetails.js
hatched: Passing this prop in is no longer required.

src/pages/Models/Details/__snapshots__/ModelDetails.test.js.snap
hatched: I think you marked this as resolved but didn't update the assertions :)

src/components/InfoPanel/InfoPanel.js
```diff
@@ -1,39 +1,55 @@
...
+  const modelStatusData = useSelector(getModelStatusMemo);
+
   return (
     <div className="info-panel">
-      <div className="info-panel__grid">{gridItems}</div>
+      <div className="info-panel__grid">
+        <div className="info-panel__grid-item">
+          <h4 className="p-muted-heading">Model</h4>
+          <p data-name="name">
+            {modelStatusData ? modelStatusData.model.name : ""}
```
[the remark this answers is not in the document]
barrymcgee: Ok

PR #81, opened 2019-10-14
Restyle primary navigation

barrymcgee:
## Done

- Remove top bar
- Move links into sidebar under toggle 
- Apply dark theme

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code
- View nav

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/942


anthonydillon:
@barrymcgee looking good, do you have a link to the design to check the style?

anthonydillon:
I now see horizontal scrolling on the model details page.

src/components/PrimaryNav/PrimaryNav.test.js
```diff
@@ -13,4 +18,40 @@ describe("Primary Nav", () => {
...
+    global.window.location = {
+      pathname: "/logs"
+    };
+    const wrapper = mount(
+      <Router>
+        <PrimaryNav />
+      </Router>
+    );
+    expect(
+      wrapper.find(".internal .p-list__item:last-child").hasClass("is-selected")
```
hatched: This test isn't very resilient as if we change the order of the nav then the test could fail/pass erroneously. Probably best to instead loop through the items to make sure only a single one has `is-selected` and that its value is the label from `logs`.

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+    transition: max-height 0.25s ease-in-out;
+  }
+
+  &.ext-nav-open {
+    .external {
+      max-height: 200px;
+    }
+
+    .p-primary-nav__toggle {
+      transform: rotate(180deg);
```
hatched: In Chrome, clicking on the chevron 'selects' the element and because it's selected it also rotates the blue 'selection' border.

Firefox however has no such border and instead has a 'mouse down' background highlight which disappears when the animation starts. If you navigate to the element in FF and interact with it via the keyboard the blue border appears and rotates with the chevron.

In Chrome can we mirror the Firefox experience somehow? It's much cleaner. 👍

hatched:
I also saw the horizontal overflow on the model details page. You'll probably have to modify the width in the terminal component styles: https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/components/Terminal/_terminal.scss

anthonydillon:
It looks like the terminal dynamic sets the width as an online style. This is hopefully something that is configurable.

src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -1,72 +1,78 @@
...
+            className="p-primary-nav__logo"
+            src="https://assets.ubuntu.com/v1/a9e0ed4a-jaas-logo1.svg"
+            alt="JAAS logo"
+          />
+        </Link>
+        <button
+          className="p-primary-nav__toggle"
+          onClick={() => setExtNavOpen(!extNavOpen)}
+        >
+          <i className="p-icon--contextual-menu"></i>
```
anthonydillon: You can add text inside an icon for accessibility which Vanilla will hide.

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+    display: grid;
+    grid-template-columns: 1fr 40px;
+    margin-bottom: -1rem;
+    padding: 0.5rem;
+  }
+
+  &__toggle {
+    background: none;
+    border: 0;
+    transition: transform 0.25s ease-in-out;
```
anthonydillon: Please use the Vanilla animation mixin: https://docs.vanillaframework.io/settings/animation-settings/

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+.p-primary-nav {
+  a,
+  a:hover {
+    color: $color-light;
+  }
+
+  &__header {
+    display: grid;
+    grid-template-columns: 1fr 40px;
+    margin-bottom: -1rem;
```
anthonydillon: Why not set the height with `vh`?

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+    padding: 0.5rem;
+  }
+
+  &__toggle {
+    background: none;
+    border: 0;
+    transition: transform 0.25s ease-in-out;
+
+    &:hover {
+      background: 0;
```
anthonydillon: This should be `background: none` but why do you need it if the toggle already had it above?

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  .p-list__link {
+    color: $color-light;
+  }
+
+  .p-list__item {
+    padding-left: 2.5rem;
```
anthonydillon: It would be nicer to add the padding too the links so you can click the row and now just on the word.

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    &.is-selected {
+      background-color: #044f66;
+      border-left: 3px solid #3fc8f2;
+      margin-right: -0.5rem;
+      padding-left: calc(2.5rem - 3px);
```
anthonydillon: If all links have a transparent 3px border on the left then selected you only need to change the colour. Saves complexity and repaints

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  .external {
```
anthonydillon: Could you continue to use a BEM naming convention for this section please

src/components/PrimaryNav/PrimaryNav.js
barrymcgee: That's not mentioned in the docs - added an issue: https://github.com/canonical-web-and-design/vanilla-framework/issues/2573

src/components/PrimaryNav/_primary-nav.scss
barrymcgee: Didn't think of that - nice.

src/components/PrimaryNav/_primary-nav.scss
barrymcgee: Vanilla gives button hover a specific grey background which is not overridden by the toggle above.

src/components/PrimaryNav/_primary-nav.scss
barrymcgee: Viewport height cannot be relied upon here to give a consistent height for the navigation header.

src/components/PrimaryNav/_primary-nav.scss
barrymcgee: I did try but hit pain - issue raised here: https://github.com/canonical-web-and-design/vanilla-framework/issues/2576

barrymcgee:
Design in Zelin (circles are icon placeholders) - https://app.zeplin.io/project/5d2dfd8e5b67f14c5e1d2aef/screen/5da02df2ebc1b630a9a955f4

src/components/PrimaryNav/PrimaryNav.test.js
```diff
@@ -13,4 +18,40 @@ describe("Primary Nav", () => {
...
     expect(wrapper).toMatchSnapshot();
   });
+
+  it("toggles external nav menu", () => {
+    const wrapper = mount(
+      <Router>
+        <PrimaryNav />
+      </Router>
+    );
+    wrapper.find(".p-primary-nav__toggle").simulate("click");
```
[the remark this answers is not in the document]
barrymcgee: Isn't that implicit with `wrapper.find(".p-primary-nav__toggle")` - if that's can't be found, the test fails?

src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -1,72 +1,78 @@
-import React from "react";
+import React, { useState } from "react";
 import { Link } from "react-router-dom";
+import classNames from "classnames";
 
-function PrimaryNav() {
+import "./_primary-nav.scss";
+
+const pages = [
```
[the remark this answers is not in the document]
barrymcgee: Pls add as follow-up issue - doesn't appear trivial.

src/components/PrimaryNav/_primary-nav.scss
barrymcgee: You're talking about a browser level accessibility difference - something which it feels icky to mess with and even if we did, I'd prefer it was beyond the scope of this PR - please add a follow-up.

src/components/PrimaryNav/PrimaryNav.js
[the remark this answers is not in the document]
hatched: Issue created: https://github.com/canonical-web-and-design/jaas-dashboard/issues/82

src/components/PrimaryNav/PrimaryNav.test.js
```diff
@@ -13,4 +18,40 @@ describe("Primary Nav", () => {
...
+    expect(wrapper.find(".p-primary-nav").hasClass("ext-nav-open")).toEqual(
+      false
+    );
+  });
+
+  it("applies is-selected state correctly", () => {
+    delete global.window.location;
+    global.window = Object.create(window);
+    global.window.location = {
```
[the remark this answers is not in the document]
hatched: Issue created to address this in a follow-up: https://github.com/canonical-web-and-design/jaas-dashboard/issues/83

src/components/PrimaryNav/_primary-nav.scss
hatched: I may not have the background to understand why, but if it's not clear maybe a comment should be added here for future dev's who think to make the same change.

src/components/PrimaryNav/_primary-nav.scss
hatched: Hah, smart :D

src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -1,72 +1,104 @@
...
+            Discourse
           </a>
+        </li>
+        <li className="p-list__item">
           <a
-            href="#navigation-closed"
-            className="p-navigation__toggle--close"
-            title="close menu"
+            className="p-link--external p-list__link"
+            href="https://jaas.ai/docs/"
```
hatched: I think you missed the `target="_blank"` on this one.

src/components/PrimaryNav/PrimaryNav.test.js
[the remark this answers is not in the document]
hatched: I just mean that the element shouldn't have `ext-nav-open` on the initial page load. So this click and the subsequent assertion may not actually be testing anything. 

You'll want to assert first that it does not have `ext-nav-open`.

spencerbygraves:
@barrymcgee I made some styling updates yesterday after a meeting with @clagom, would you prefer to land this and come back to it, perhaps when you do the open/close etc?

barrymcgee:
Yes, land this and create new issue for new changes.

src/components/PrimaryNav/PrimaryNav.test.js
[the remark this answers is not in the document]
barrymcgee: Ok, I've added that check.

barrymcgee:
> I'm also assuming that you plan to address the open/close of the sidebar interaction in a follow-up? Could you please create an issue to track this.

This is already tracked elsewhere.

PR #88, opened 2019-10-16
Sort models by status and display them in separate tables

hatched:
## Done

Naively sort the models by status and then display them in separate tables. 

## QA

- Pull code
- Open http://0.0.0.0:8036/
- If you have models with units who's agents are `lost` then you'll see them in the topmost table.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/929

We don't have a spec for this so at the moment it's only splitting the models by applications that have units who's agents are `lost`. This will need a follow-up to enhance this sorting functionality but it's good to get the component output landed so we can style things.

I'll leave https://github.com/canonical-web-and-design/juju-squad/issues/930 open to track for the follow-up

## Screenshots

For those who may not have models in this state:

Screen Shot 2019-10-16 at 2 49 09 PM

### DIFF
--- a/src/components/TableList/TableList.js
+++ b/src/components/TableList/TableList.js
@@ -81,12 +81,13 @@ function determineModelStatus(modelStatus) {
   // Short circuit the loop. If something is in error then the model is in error.
   // Grab the first error message and display that.
   Object.values(modelStatus.applications).some(app => {
-    Object.values(app.units).some(unit => {
+    return Object.values(app.units).some(unit => {
       if (badStatuses.includes(unit.agentStatus.status)) {
         status = "blockedModelData";
         why = unit.agentStatus.info;
         return true;
       }
+      return false;
     });
   });
   return { status, why };
--- a/src/components/TableList/TableList.js
+++ b/src/components/TableList/TableList.js
@@ -4,6 +4,7 @@ import { useSelector } from "react-redux";
 
 import MainTable from "../MainTable/MainTable";
 
+import "./_table-list.scss";
 /**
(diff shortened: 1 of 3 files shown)


src/components/TableList/TableList.js
```diff
@@ -30,9 +36,19 @@ function generateModelTableData(state) {
       }`;
       return <Link to={sharedModelDetailsPath}>{modelListData.name}</Link>;
     };
-    return {
+    const generateModelNameCell = why => {
+      const link = generateModelDetailsLink();
+      return (
```
hatched: I fully expect that the markup here will change when styled.

src/components/TableList/TableList.js
```diff
@@ -44,8 +60,37 @@ function generateModelTableData(state) {
...
+  return modelData;
+}
+
+/**
+  Uses the model status to determine what the status of the model is and then
+  returns the string value of the table it's to be displayed in.
+  @param {Object} modelStatus The model status.
+  @returns {String} The status table for it to be displayed in.
+*/
+function determineModelStatus(modelStatus) {
```
hatched: I also expect the logic here to change once we've defined a spec.

src/components/TableList/__snapshots__/TableList.test.js.snap
```diff
@@ -5,7 +5,87 @@ Array [
   <th
     role="columnheader"
   >
-    Name
```
hatched: These snapshots aren't great. Let's plan to fix them during the styling/markup changes.

PR #90, opened 2019-10-17
Add static UI build for filtering of models

barrymcgee:
## Done

Add static UI build for filtering of models

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code
- Cross ref with designs

## Zeplin designs

https://zpl.io/VqkEBGl

### DIFF
--- a/src/components/ButtonToggle/ButtonToggle.js
+++ b/src/components/ButtonToggle/ButtonToggle.js
@@ -1,0 +1,18 @@
+import React from "react";
+
+import "./_button-toggle.scss";
+
+const ButtonToggle = () => {
+  return (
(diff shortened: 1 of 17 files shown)


src/components/ButtonToggle/ButtonToggle.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+
+import "./_button-toggle.scss";
+
+const Button = () => {
```
hatched: ButtonToggle :-)

barrymcgee: Good catch

PR #91, opened 2019-10-17
Add message if a model list table contains no models

hatched:
## Done

Add message if a model list table contains no models.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- You should see all three tables have a message on the initial load that they have no models, and then as your data comes in that message should disappear.

### DIFF
--- a/src/components/TableList/TableList.js
+++ b/src/components/TableList/TableList.js
@@ -155,6 +155,17 @@ function generateTableHeaders(label) {
   ];
 }
 
+function renderNoDataMessage(modelData) {
+  if (modelData.length === 0) {
+    return (
+      <div className="table-list_no-models-message">
+        No models currently in this status
+      </div>
+    );
+  }
+  return null;
+}
+
 function TableList() {
   const {
     blockedModelData,
@@ -168,16 +179,19 @@ function TableList() {
         headers={generateTableHeaders("Blocked")}
         rows={blockedModelData}
       />
+      {renderNoDataMessage(blockedModelData)}
       <MainTable
         className={"u-table-layout--auto"}
         headers={generateTableHeaders("Attention")}
         rows={attentionModelData}
       />
+      {renderNoDataMessage(attentionModelData)}
       <MainTable
         className={"u-table-layout--auto"}
         headers={generateTableHeaders("Running")}
         rows={runningModelData}
       />
+      {renderNoDataMessage(runningModelData)}
     </>
   );
 }
--- a/src/components/TableList/TableList.js
+++ b/src/components/TableList/TableList.js
@@ -41,7 +41,7 @@ function generateModelTableData(state) {
       return (
         <>
           {link}
-          <div className="table-list_error-message">{why}</div>
+          <div className="table-list__error-message">{why}</div>
         </>
       );
     };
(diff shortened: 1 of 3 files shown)


src/components/TableList/TableList.js
```diff
@@ -155,6 +155,17 @@ function generateTableHeaders(label) {
   ];
 }
 
+function renderNoDataMessage(modelData) {
+  if (modelData.length === 0) {
+    return (
+      <div className="table-list_no-models-message">
```
hatched: We dont' really have a design for this but I imagine it'll take on some of the styling from the overall design yet to come.

src/components/TableList/_table-list.scss
```diff
@@ -1,4 +1,11 @@
-.table-list_error-message {
-  color: red;
-  font-size: 0.5rem;
+.table-list {
+  &_error-message {
+    color: red;
+    font-size: 0.5rem;
+  }
+
+  &_no-models-message {
```
anthonydillon: Should be `&--no-models-message` at it's a modifier on the list.

src/components/TableList/_table-list.scss
```diff
@@ -1,4 +1,11 @@
-.table-list_error-message {
-  color: red;
-  font-size: 0.5rem;
+.table-list {
+  &_error-message {
```
anthonydillon: Should be `&--error-message` at it's a modifier on the list.

src/components/TableList/TableList.js
barrymcgee: Double underscore BEM notation

hatched:
@barrymcgee you'rs and @anthonydillon's comments conflict with eachother...

Maybe BEM isn't any good if people can't agree on the correct type of hyphen  😆 😆

hatched:
@anthonydillon Unfortunately the table, which is a shared component, doesn't support empty state so we have to handle it externally.

src/components/TableList/_table-list.scss
hatched: After reading the documentation I think that this needs to be `&__no-models-message` as it's an element not a modifier.

anthonydillon:
After thinking about it further in fact I think we may all have been wrong. I think the real issue is the banking of the classes. `error-message` and `no-models-message` are very specific and can not be reused widely. I would suggest using [a state class](https://canonical-web-and-design.github.io/practices/coding/stylesheets.html#state-classes) like `has-error` or `has-warning`.

clagom:
Looks good visually speaking

hatched:
> I would suggest using [a state class](https://canonical-web-and-design.github.io/practices/coding/stylesheets.html#state-classes) like `has-error` or `has-warning`.

@anthonydillon that would be fine but we still need some way to target styling at that specific element for the formatting.

hatched:
@barrymcgee can you please clarify what you'd like me to do? Shall I close this PR and instead propose a change to the shared component to add an 'empty' element state to the table?

barrymcgee:
> propose a change to the shared component to add an 'empty' element state to the table?

@hatched Preferably, yes.

barrymcgee:
@hatched Any update on this PR?

hatched:
@barrymcgee I'll be modifying the table to take an 'empty' state element when time permits.

spencerbygraves:
"No models currently **have** this status"?

cristinadresch:
@hatched any update on this?

hatched:
Nope

hatched:
Closing this as a different approach will be taken

PR #93, opened 2019-10-18
Remove superfluous role attributes

barrymcgee:
## Done

Remove superfluous role attributes

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code


hatched:
Looks good! Thanks @barrymcgee 

Before this lands, @squidsoup @huwshimi will making this change cause any issues on your end?

squidsoup:
@barrymcgee seems reasonable to me given we're using the semantic element - @barrymcgee would be good to get confirmation from @huwshimi that this won't have any unintended consequences however.

huwshimi:
@squidsoup as far as I know that should be fine, this was implemented from the examples the Vanilla docs so presuming aren't any Vanilla side affects this should be fine.

PR #100, opened 2019-10-23
Add sidebar icons

barrymcgee:
## Done

Add sidebar icons

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify sidebar icons are [as in design](https://zpl.io/adYzyrn)

## Details

Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/95

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.js
+++ b/src/components/PrimaryNav/PrimaryNav.js
@@ -5,10 +5,14 @@ import classNames from "classnames";
 import "./_primary-nav.scss";
 
 const pages = [
-  { label: "Models", path: "/" },
-  { label: "Controllers", path: "/controllers" },
-  { label: "Usage", path: "/usage" },
-  { label: "Logs", path: "/logs" }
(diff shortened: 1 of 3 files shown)


src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -94,6 +98,11 @@ const PrimaryNav = () => {
             })}
           >
             <Link className="p-list__link" to={navItem.path}>
+              <img
+                className="p-list__icon"
+                src={`https://assets.ubuntu.com/v1/${navItem.icon}`}
```
hatched: We'll need to make sure that this can work offline if it's going to be part of Juju so we can't access the assets from the asset server.

barrymcgee: Issue raised here: https://github.com/canonical-web-and-design/jaas-dashboard/issues/101

hatched:
QA OK

Just also making a note that we have to do a follow-up to replace the "models" and "controllers" labels with the updated icons in the future.

PR #103, opened 2019-10-23
Add model count summary header

hatched:
## Done

Added model count summary header to the model listings page.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The model count in the top should update as your model data comes in. 

## Details

The data used to generate this count and the data used to sort the table are currently different so you'll likely see different sorting. The immediate follow-up to this branch will be to bring those two in line and expand the logic used to determine how models are categorized. 

## Screenshots

Screen Shot 2019-10-23 at 10 33 58 AM


src/app/selectors.js
```diff
@@ -105,6 +105,92 @@ const getModelStatusByUUID = (modelUUID, modelStatuses) => {
...
 
+// Highest status to the right
+const statusOrder = ["running", "alert", "blocked"];
+
+/**
+  Returns a grouped collection of model statuses.
+  @param {Object} modelStatuses
+  @returns {Function} The grouped model statuses.
+*/
+const groupModelStatuses = modelStatuses => {
```
hatched: The logic used here is pretty basic. In the follow-up I'll expand it with additional entities to check to determine the model status. Note the empty `getUnitStatusGroup` function below.

src/app/selectors.js
```diff
@@ -148,11 +234,31 @@ export const isLoggedIn = state =>
...
     getModelStatuses,
     modelStatuses => getModelStatusByUUID(modelUUID, modelStatuses)
   );
 };
+
+/**
+  Returns the model statuses sorted by status.
+  @returns {Function} The memoized selector to return the sorted model statuses.
+*/
+export const getGroupedModelStatuses = createSelector(
```
hatched: This is exported because it'll be used as a selector for determining the sorting order of the table in the follow-up

src/pages/Models/Models.js
```diff
@@ -1,20 +1,38 @@
...
 export default function Models() {
+  const { blocked, alert, running } = useSelector(getGroupedModelStatusCounts);
+  const models = blocked + alert + running;
   return (
     <Layout>
       <Header>
         <div className="models__header">
           <div className="models__count">
-            258 models: 4 blocked, 1 alert, 253 running
+            {`${models} ${pluralize(
```
hatched: Sorry, prettier had a fit...

PR #111, opened 2019-10-25
Update Models listing styling

### LINKED ISSUE #937: Local Apps title should be Local App
In an applications units view there can only ever be a single app so the "Local Apps" table should not be plural. This pluralization should be based on the size of data sent to the table as the same table is used elsewhere as well.

Screen Shot 2021-03-29 at 2 58 33 PM

barrymcgee:
## Done

Update Models listing styling

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check user icon dropdown and logout button
- Sanity check code

Fixes #937

### DIFF
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -3,7 +3,7 @@ export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
-  LogOut: "LOG_OUT"
+  logOut: "LOG_OUT"
 };
 
 // Action creators
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -3,7 +3,7 @@ export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
-  LogOut: "FLUSH_LOCAL_STORAGE"
+  LogOut: "LOG_OUT"
 };
 
 // Action creators
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -2,7 +2,8 @@
 export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
-  updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION"
+  updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
+  LogOut: "FLUSH_LOCAL_STORAGE"
 };
 
 // Action creators
@@ -36,5 +37,18 @@ export function storeVisitURL(visitURL) {
   return {
     type: actionsList.storeVisitURL,
     payload: visitURL
+  };
+}
+
+/**
+  Flush localStorage login keys
+*/
+export function LogOut() {
+  console.log("logging ou");
+  localStorage.removeItem("identity");
+  localStorage.removeItem("https://api.jujucharms.com/identity");
+  return {
+    type: actionsList.LogOut,
+    payload: true
   };
 }
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -44,7 +44,6 @@ export function storeVisitURL(visitURL) {
   Flush localStorage login keys
 */
 export function LogOut() {
-  console.log("logging ou");
   localStorage.removeItem("identity");
   localStorage.removeItem("https://api.jujucharms.com/identity");
   return {
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -40,14 +40,20 @@ export function storeVisitURL(visitURL) {
   };
 }
 
+export function clearUserData() {
+  return {
+    type: actionsList.logOut
+  };
+}
+
+// Thunks
 /**
   Flush localStorage login keys
 */
-export function LogOut() {
-  localStorage.removeItem("identity");
-  localStorage.removeItem("https://api.jujucharms.com/identity");
-  return {
-    type: actionsList.LogOut,
-    payload: true
+export function logOut() {
+  return async function thunk(dispatch) {
+    localStorage.removeItem("identity");
+    localStorage.removeItem("https://api.jujucharms.com/identity");
+    dispatch(clearUserData());
   };
 }
--- a/src/app/root.js
+++ b/src/app/root.js
@@ -14,6 +14,9 @@ function rootReducer(state = {}, action) {
       case actionsList.storeVisitURL:
         draftState.visitURL = action.payload;
         break;
+      case actionsList.LogOut:
+        draftState.LogOut = action.payload;
+        break;
       default:
         // no default value, fall through.
         break;
--- a/src/app/root.js
+++ b/src/app/root.js
@@ -14,8 +14,8 @@ function rootReducer(state = {}, action) {
       case actionsList.storeVisitURL:
         draftState.visitURL = action.payload;
         break;
-      case actionsList.LogOut:
-        draftState.LogOut = action.payload;
+      case actionsList.logOut:
+        draftState.bakery = null;
         break;
       default:
         // no default value, fall through.
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -4,3 +4,7 @@ export const generateStatusIcon = status => {
   let statusClass = status ? `is-${status.toLowerCase()}` : "";
   return <span className={"status-icon " + statusClass}>{status}</span>;
 };
+
+export const generateSpanClass = (className, value) => {
+  return <span className={className}>{value}</span>;
+};
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -1,6 +1,6 @@
 import React from "react";
 
-export const assignStatusIcon = status => {
+export const generateStatusIcon = status => {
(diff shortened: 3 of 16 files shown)


hatched:
Is it intentional that there is no lower border on the headings when there are no records in the table?  

Screen Shot 2019-10-25 at 10 36 32 AM

src/app/utils.js
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+
+export const assignStatusIcon = status => {
```
hatched: `generateStatusIcon`

src/components/UserIcon/UserIcon.js
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from "react";
+import classNames from "classnames";
+
+import "./_user-icon.scss";
+
+export default function User(props) {
+  const [userPanelVisibility, setUserPanelVisibility] = useState(false);
+
+  console.log(props);
```
hatched: remove log

src/components/UserIcon/UserIcon.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const logOut = e => {
+    e.preventDefault();
+    console.log("logging out");
+    localStorage.removeItem("identity");
+    localStorage.removeItem("https://api.jujucharms.com/identity");
```
hatched: Ping me about this please.

clagom:
I don’t think this is by design, no

src/components/UserIcon/UserIcon.js
```diff
@@ -0,0 +1,50 @@
...
+    // TODO: Use React Router to redirect correctly
+    window.location.href = "/";
```
hatched: I'm not sure this is necessary - the application should just render the 'login UI' once this has cleared. If that's not happening then we have a different problem.

If your intention was to navigate to the `/` just by matter of course I would add a `<link>` which would use the router to navigate to / but then attach an event handler to the click on that element which would first clear the items from local storage.

src/components/UserIcon/UserIcon.js
```diff
@@ -0,0 +1,50 @@
...
+import "./_user-icon.scss";
+
+export default function User(props) {
+  const [userPanelVisibility, setUserPanelVisibility] = useState(false);
+
+  console.log(props);
+
+  const logOut = e => {
+    e.preventDefault();
+    console.log("logging out");
```
hatched: log

src/components/UserIcon/UserIcon.test.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import UserIcon from "./UserIcon";
+
+describe("User Icon", () => {
```
hatched: This should have a test for opening the element and performing a logout.

src/components/UserIcon/UserIcon.js
```diff
@@ -0,0 +1,50 @@
...
+    localStorage.removeItem("identity");
+    localStorage.removeItem("https://api.jujucharms.com/identity");
+    // TODO: Use React Router to redirect correctly
+    window.location.href = "/";
+  };
+
+  return (
+    <div className="user-icon">
+      <i
+        className="p-icon--user"
```
hatched: This icon needs a `cursor: pointer` style.

src/components/UserIcon/_user-icon.scss
```diff
@@ -0,0 +1,39 @@
...
+    &-arrow {
+      background: $color-x-light;
+      box-shadow: 3px 3px 6px -3px rgba(0, 0, 0, 0.2);
+      height: 1rem;
+      position: absolute;
+      right: 1.5rem;
+      top: -0.25rem;
+      transform: rotate(225deg);
+      width: 1rem;
+      z-index: 2;
```
hatched: We should probably implement a z-index stacking order in the application like we had in the old GUI.

https://github.com/canonical-web-and-design/jaas-dashboard/issues/112

src/scss/index.scss
```diff
@@ -1,9 +1,11 @@
...
 @import "node_modules/vanilla-framework/scss/vanilla";
 
-$grid-max-width: 100%;
+// Import custom patterns
+@import "status_icons";
 
 // Settings
+$grid-max-width: 100%;
 $color-navigation-background: $color-brand;
 // XXX remove when updating from Vanilla 2.3.0
```
hatched: https://github.com/canonical-web-and-design/jaas-dashboard/issues/113

src/components/UserIcon/UserIcon.js
barrymcgee: The first point - it does not just render login.

Second point - that approach doesn't work when you're already on `/`

hatched:
Spacing between the icons is too large compared to the design.
Screen Shot 2019-10-30 at 10 00 03 AM

src/app/actions.js
```diff
@@ -38,3 +39,15 @@ export function storeVisitURL(visitURL) {
     payload: visitURL
   };
 }
+
+/**
+  Flush localStorage login keys
+*/
+export function LogOut() {
```
hatched: In a basic `action creator` you don't really want any side effects and instead only try and generate the `message object`. To get around this limitation we use the `thunks` redux plugin which allows us to perform arbitrary side effects. It's something you want to try and limit as it can make the code really hard to reason about if abused. But in this case it's what you want.

A more idiomatic approach to this would be something like we've done in https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/juju/actions.js#L56

This code would become something like...

```JavaScript
export function clearUserData() {
  return {
    type: actionsList.clearUserData
  }
}

export function logOut() {
  return async function thunk(dispatch, getState) {
    localStorage.removeItem("identity");
    localStorage.removeItem("https://api.jujucharms.com/identity");;
    dispatch(clearUserData());
  };
}
```

src/app/actions.js
```diff
@@ -2,7 +2,8 @@
 export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
-  updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION"
+  updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
+  LogOut: "LOG_OUT"
```
hatched: Lets use something more descriptive `clearUserData`. This will allow us to compose actions in the future if needed.

src/app/root.js
```diff
@@ -14,6 +14,9 @@ function rootReducer(state = {}, action) {
       case actionsList.storeVisitURL:
         draftState.visitURL = action.payload;
         break;
+      case actionsList.LogOut:
+        draftState.LogOut = action.payload;
```
hatched: We don't actually want to store any payload but instead want to clear the stored user data.

You can see here https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/app/selectors.js#L245 that the `isLoggedIn` checks for the existence of the controller and bakery. So you'll want to invalidate this check. I'd recommend nulling out the `bakery` instance. This will cause the application to re-render and when the login check is made it'll return false, logging the user out.

The application doesn't actually support having the bakery ripped out from under it so both this reducer and how the application creates the bakery instances will need to be updated afterwards as a follow-up.

barrymcgee:
> Is it intentional that there is no lower border on the headings when there are no records in the table?

@hatched Two valid tables following each other will have the correct marhins applied by Vanilla. This issue because it's a table header without any subsequent table rows - an issue that will be addressed when your [PR here](https://github.com/canonical-web-and-design/jaas-dashboard/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc) lands.

hatched:
Thanks for the continued work here. Switching to `<Link>` kept the page from refreshing as I expected but we have to do the `onClick` a bit different than you've got here. You'll find the diff below.

This will correctly log the user out and render the login UI but you won't be able to log in again as per the comment somewhere in the comments above :-)

```diff
diff --git a/src/app/actions.js b/src/app/actions.js
index b989b77..d4ac46c 100644
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -3,7 +3,7 @@ export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
-  LogOut: "LOG_OUT"
+  logOut: "LOG_OUT"
 };

 // Action creators
diff --git a/src/components/UserIcon/UserIcon.js b/src/components/UserIcon/UserIcon.js
index e5194d4..8cd1b11 100644
--- a/src/components/UserIcon/UserIcon.js
+++ b/src/components/UserIcon/UserIcon.js
@@ -1,6 +1,7 @@
 import React, { useState } from "react";
 import classNames from "classnames";
 import { Link } from "react-router-dom";
+import { useDispatch } from "react-redux";

 import { logOut } from "app/actions";

@@ -8,7 +9,7 @@ import "./_user-icon.scss";

 export default function User(props) {
   const [userPanelVisibility, setUserPanelVisibility] = useState(false);
-
+  const dispatch = useDispatch();
   return (
     <div className="user-icon">
       <i
@@ -31,7 +32,7 @@ export default function User(props) {
             <a href="#_">Help</a>
           </li>
           <li className="p-list__item">
-            <Link to="/" onClick={() => logOut()}>
+            <Link to="/" onClick={() => dispatch(logOut())}>
               Log out
             </Link>
           </li>
```

PR #116, opened 2019-10-28
Group model list models using new selectors

hatched:
## Done

The model list is now grouped using the same selector as the model summary in the header.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Your models should be grouped between the three sections and those should match the values in the header summary.

## Details

A new section of model data is created in the Redux store, `modelData` this is a collection of all the data we get from all the different API calls. This is done because the multiple locations for data was becoming unwieldy to work with. This will also mirror more closely how the data will be structured when we have a single API for all of the model data.

Unfortunately by making this change the test scaffolds no longer work properly as they do not support the `modelData` top level key. A follow-up [issue](https://github.com/canonical-web-and-design/jaas-dashboard/issues/115) has been created to track this.

While this groups them, we'll still need to decide exactly where to draw the line between blocked and alert.

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/965
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/930

### DIFF
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -246,6 +246,15 @@ export const isLoggedIn = state =>
   state.root.controllerConnection && state.root.bakery;
 
 /**
+  Returns the users current controller logged in identity
+  @param {Object} state The application state.
+  @returns {String} The users userTag.
+*/
+export const getActiveUserTag = state =>
+  state.root.controllerConnection &&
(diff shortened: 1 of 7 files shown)


hatched:
Thanks guys!

PR #134, opened 2019-10-30
Models details collapsible sidebar

barrymcgee:
## Done

Make the models details sidebar collapsible

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to models details page and see the sidebar is collapsed. 
- Hover over to expand and interact with it.
- Hover off to see it collapse again.
- No other page should have this behavior.

## Details

#129 

## Known issues

- I tried to test for the existence of the `has-collapsible-sidebar` on the `<ModelDetails>` component but it's not there when Jest loads it - is this before useEffect onDomReady?

- When you expand the sidenav, open the external nav and hover off - the external nav does not close. Will follow up later.

- The icons in the collapsed sidebar will need proper aligning and pixel pushing. I will follow up with @spencerbygraves on this.

### DIFF
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -3,8 +3,7 @@ export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
-  logOut: "LOG_OUT"
-  LogOut: "LOG_OUT",
+  logOut: "LOG_OUT",
   collapsibleSidebar: "TOGGLE_COLLAPSIBLE_SIDEBAR"
 };
 
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -4,6 +4,8 @@ export const actionsList = {
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
   logOut: "LOG_OUT"
+  LogOut: "LOG_OUT",
+  collapsibleSidebar: "TOGGLE_COLLAPSIBLE_SIDEBAR"
 };
 
 // Action creators
@@ -55,5 +57,15 @@ export function logOut() {
     localStorage.removeItem("identity");
     localStorage.removeItem("https://api.jujucharms.com/identity");
     dispatch(clearUserData());
+  };
+}
+
+/**
+  Toggle collapsible sidebar
+*/
+export function collapsibleSidebar(toggle) {
+  return {
+    type: actionsList.collapsibleSidebar,
+    payload: toggle
   };
 }
--- a/src/app/root.js
+++ b/src/app/root.js
@@ -17,6 +17,9 @@ function rootReducer(state = {}, action) {
       case actionsList.logOut:
         draftState.bakery = null;
         break;
+      case actionsList.collapsibleSidebar:
+        draftState.collapsibleSidebar = action.payload;
+        break;
       default:
         // no default value, fall through.
         break;
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -206,6 +206,20 @@ const countModelStatusGroups = groupedModelStatuses => {
 // ----- Exported functions
 
 /**
+  Checks state to see if the sidebar is collapsible.
+  Usage:
+    const isSidebarCollapsible = useSelector(isSidebarCollapsible);
+
+  @param {Object} state The application state.
+  @returns {Boolean} If the sidebar is collapsible.
+*/
+export const isSidebarCollapsible = state => {
+  if (state && state.root) {
+    return state.root.collapsibleSidebar;
+  }
+};
+
+/**
   Gets the model UUID from the supplied name using a memoized selector
   Usage:
     const getModelUUIDMemo = useMemo(getModelUUID.bind(null, modelName), [
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -1,12 +1,21 @@
 import React from "react";
+import { useSelector } from "react-redux";
 import PrimaryNav from "components/PrimaryNav/PrimaryNav";
+import classNames from "classnames";
+
+import { isSidebarCollapsible } from "app/selectors";
 
 import "./_layout.scss";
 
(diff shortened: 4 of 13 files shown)


hatched:
QA Notes:

> When you expand the sidenav, open the external nav and hover off - the external nav does not close. Will follow up later.

I'm not sure this is actually a bug - I've found that the target area to expand the sidebar feels quite small in Firefox (see below) so you don't necessarily want to close that. I'd probably leave it as-is for now but I don't feel strongly either way.

In Firefox if you pass through the collapsed sidebar so that your cursor is to resize the window the sidebar stays collapsed, in Chrome it does not. While this is outside of our control, it makes the target hover area effectively a lot bigger and more useful in Chrome than in Firefox. I'd probably suggest a wider collapsed sidebar because of this. cc @clagom 

Also in Firefox, the expanding is not very smooth (drops down to 13fps) vs in Chrome. What do you think about instead of the expansion pushing the content over it just expands over top? See the ZenHub sidebar for an example of what I mean. cc @clagom

src/components/Layout/Layout.js
```diff
@@ -1,12 +1,23 @@
...
+import classNames from "classnames";
+import useHover from "hooks/useHover";
+
+import { isSidebarCollapsible } from "app/selectors";
 
 import "./_layout.scss";
 
 const Layout = ({ children }) => {
+  const [sidebarRef, isSidebarHovered] = useHover();
+  const sidebarCollapsible = useSelector(isSidebarCollapsible);
```
hatched: For the implemented use case I'd say that this should be done using a prop passed from the parent page component being rendered instead of storing this in the redux state. This is because the sidebar on `Models` doesn't care what the sidebar on `Details` did. 

However if we want to implement some sidebar pinning that would persist through pages this implementation makes it quite trivial to do so. It also gives us a great way for different parts of the application to have knowledge available about the UI state of the application.

src/app/actions.js
```diff
@@ -57,3 +58,13 @@ export function logOut() {
     dispatch(clearUserData());
   };
 }
+
+/**
+  Toggle collapsible sidebar
+*/
+export function collapsibleSidebar(toggle) {
```
hatched: In an effort to support the 'pin-able' sidebar you could convert this to accept an enum. That way it could be set to various states like... `['closed', 'open', 'floating', 'pushing']`

src/app/actions.js
```diff
@@ -3,7 +3,8 @@ export const actionsList = {
   storeBakery: "STORE_BAKERY",
   storeVisitURL: "STORE_VISIT_URL",
   updateControllerConnection: "UPDATE_CONTROLLER_CONNECTION",
-  logOut: "LOG_OUT"
+  logOut: "LOG_OUT",
+  collapsibleSidebar: "TOGGLE_COLLAPSIBLE_SIDEBAR"
```
hatched: To support multiple states you could change this to be `SET_SIDEBAR_VISIBILITY` or the like...

The variable names throughout would probably do for an update in that case then.

src/app/root.js
```diff
@@ -17,6 +17,9 @@ function rootReducer(state = {}, action) {
       case actionsList.logOut:
         draftState.bakery = null;
         break;
+      case actionsList.collapsibleSidebar:
+        draftState.collapsibleSidebar = action.payload;
```
hatched: At first I thought we should put UI state in it's own top level key but the `root` state doesn't change often outside of these sidebar changes so we can look at that as a performance optimization later on if necessary.

src/app/selectors.js
```diff
@@ -205,6 +205,20 @@ const countModelStatusGroups = groupedModelStatuses => {
...
 
+/**
+  Checks state to see if the sidebar is collapsible.
+  Usage:
+    const isSidebarCollapsible = useSelector(isSidebarCollapsible);
+
+  @param {Object} state The application state.
+  @returns {Boolean} If the sidebar is collapsible.
+*/
+export const isSidebarCollapsible = state => {
```
hatched: If you were to change this to support an enum this would be changed to something like `getSidebarVisibility`

src/components/Layout/Layout.test.js
```diff
@@ -1,21 +1,46 @@
...
 
+const mockStore = configureStore([]);
 describe("Layout", () => {
   it("renders without crashing and matches snapshot", () => {
-    const wrapper = shallow(<Layout />);
-    expect(wrapper).toMatchSnapshot();
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <Router>
```
hatched: Should we always be using `MemoryRouter` in tests to avoid any potential issues with routes changing? 🤷‍♂

hatched:
> I tried to test for the existence of the has-collapsible-sidebar on the <ModelDetails> component but it's not there when Jest loads it - is this before useEffect onDomReady?

`useEffect` will be called after the component is first rendered to the page. With that said the `redux-mock-store` doesn't handle setting/updating see https://github.com/dmitry-zaets/redux-mock-store#redux-mock-store

So to test these you'll want to pre-seed the mock store with the state that you want to test. Here is an example making a 'empty' state test https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/components/TableList/TableList.test.js#L15

src/components/Layout/Layout.test.js
barrymcgee: Haven't seen any issues so far - if so, we can make the change then.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -142,6 +143,16 @@ const ModelDetails = () => {
   ]);
   const modelStatusData = useSelector(getModelStatusMemo);
 
+  // Toggle collapsible sidebar when component mounts/unmounts
+  useEffect(() => {
+    // This gets called after every render
+    dispatch(collapsibleSidebar(true));
+    // Cleanup when unmounting
+    return () => {
```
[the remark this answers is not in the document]
barrymcgee: I don't know what you mean here - the default state as used by all other pages is not to have a collapsible sidebar, therefore this component needs to invert that?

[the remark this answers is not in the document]
hatched: That's what I mean, each component would set what it expects the sidebar state to be.

PR #156, opened 2019-10-31
Controller connection reliability improvements

### LINKED ISSUE #105: Controller connection disconnects before subsequent modelInfo calls can be made
**Describe the bug**
The controller connection appears to disconnect after a short period of time. If you leave the application running you'll eventually see it time out. It's possible we have to enable the pinger to keep the connection alive. With a sufficiently large list of models the model status requests take too long and the next controller request falls outside the 'keep alive' time limit.

**To Reproduce**
1. Open application and have it list the models
2. Open the devtools and wait
3. The controller connection will time out.

**Screenshots**
Screen Shot 2019-10-23 at 12 22 37 PM

hatched:
## Done

- When connecting to a controller, start the pinger on 20s increments to keep the controller connection alive. This is required because the controller interactions can be spaced out longer than the 1m timeout in the event the user has a large number of models we're polling against.
- If a model connection takes longer than 5s then abandon the connection. This was necessary because sometimes model connections hang, which prevents the poller from starting over again as something is always stuck in the queue.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The model list should work as expected
- To test the failure case, change the timeout to 50ms from 5000ms and reload. The console should have a number of console errors that an item has timed out.

## Details

Fixes #105

### DIFF
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -100,7 +100,7 @@ async function loginWithBakery(visitPage, macaroonStore) {
 }
 
 /**
-  Commects and logs in to the supplied modelURL. If the connection takes longer
+  Connects and logs in to the supplied modelURL. If the connection takes longer
   than the allowed timeout it gives up.
   @param {String} modelURL The fully qualified url of the model api.
   @param {Object} options The options for the connection.
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -154,6 +154,9 @@ async function fetchModelStatus(modelUUID) {
  */
 async function fetchAndStoreModelStatus(modelUUID, dispatch) {
   const status = await fetchModelStatus(modelUUID);
+  if (status === null) {
+    return;
+  }
   dispatch(updateModelStatus(modelUUID, status));
 }
 
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -93,7 +93,7 @@ async function loginWithBakery(visitPage, macaroonStore) {
   // Ping to keep the connection alive.
   conn.facades.pinger.pingForever(20000, err => {
     if (err) {
-      console.log("unable to ping:", err);
+      console.error("unable to ping:", err);
     }
   });
   return { bakery, conn };
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -132,13 +132,17 @@ async function connectAndLoginWithTimeout(modelURL, options, duration = 5000) {
 */
 async function fetchModelStatus(modelUUID) {
   const modelURL = controllerURL.replace("/api", `/model/${modelUUID}/api`);
-  const { conn, logout } = await jujulib.connectAndLogin(
-    modelURL,
-    {},
-    generateConnectionOptions()
-  );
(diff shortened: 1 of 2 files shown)


src/juju/index.js
```diff
@@ -80,12 +87,42 @@ async function loginWithBakery(visitPage, macaroonStore) {
   }
   const juju = await jujulib.connect(
     controllerURL,
-    generateConnectionOptions()
+    generateConnectionOptions(true)
   );
   const conn = await juju.login({});
+  // Ping to keep the connection alive.
+  conn.facades.pinger.pingForever(20000, err => {
```
anthonydillon: Is it worth checking if pinger is a member of facades here for cases when usePinger is false?

src/juju/index.js
```diff
@@ -80,12 +87,42 @@ async function loginWithBakery(visitPage, macaroonStore) {
...
+    if (err) {
+      console.log("unable to ping:", err);
```
anthonydillon: Should this be a console.error so we get reports from sentry?

src/juju/index.js
```diff
@@ -80,12 +87,42 @@ async function loginWithBakery(visitPage, macaroonStore) {
...
+    }
+  });
   return { bakery, conn };
 }
 
+/**
+  Commects and logs in to the supplied modelURL. If the connection takes longer
```
anthonydillon: Typo: Connects

src/juju/reducers.js
```diff
@@ -53,7 +53,13 @@ export default produce(
         // There don't appear to be any irrelevent data in the modelInfo so
         // we overwrite the whole object every time it changes even though
         // mostly that'll just be status timestamps.
-        draftState.modelData[modelInfo.uuid].info = modelInfo;
+        const modelData = draftState.modelData[modelInfo.uuid];
+        // If any of the status requests timeout then it's possible the data
+        // won't be available. Just abandon saving any data in that case.
+        // This will go away with the new API.
+        if (modelData) {
+          draftState.modelData[modelInfo.uuid].info = modelInfo;
```
anthonydillon: Why not use `modelData.infro = modelInfo;` instead of accessing the full object again? It made me reread this conditional to make sense of it.

src/juju/index.js
hatched: For this connection `pinger` should never be undefined as it's included in the generated options above. 

I think I'd prefer it to fail dramatically then silently fail when the pinger doesn't keep the connection alive.

src/juju/index.js
hatched: We don't have sentry yet but yep I can change that.

```suggestion
      console.error("unable to ping:", err);
```

src/juju/index.js
hatched: Thanks!
```suggestion
  Connects and logs in to the supplied modelURL. If the connection takes longer
```

src/juju/reducers.js
hatched: I was being explicit because `draftState` is a proxy supplied by the library `immer` that we use for immutable state.

PR #160, opened 2019-11-01
Use vf-animations mixin across all stylesheets

barrymcgee:
## Done

Use vf-animations mixin across all stylesheets

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Ensure no errors

### DIFF
--- a/src/components/Layout/_layout.scss
+++ b/src/components/Layout/_layout.scss
@@ -1,5 +1,5 @@
-@import "vanilla-framework/scss/vanilla";
 @import "../../scss/settings";
+@import "vanilla-framework/scss/vanilla";
 
 .l-container {
   display: flex;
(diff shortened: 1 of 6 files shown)


src/components/Layout/_layout.scss
```diff
@@ -1,4 +1,4 @@
-@import "vanilla-framework/scss/settings_colors";
+@import "vanilla-framework/scss/vanilla";
 @import "../../scss/settings";
```
anthonydillon: Don't you want settings imported before vanilla so it can override variables.

barrymcgee: They don't currently override any Vanilla settings but I suppose they may in future - I've now swapped them around.

src/components/Terminal/_terminal.scss
```diff
@@ -36,7 +40,6 @@ $terminal-header-height: 40px;
 
     .is-visible & {
       transform: rotate(0);
```
anthonydillon: Doesn't this need a animation?

barrymcgee: There's another issue to remove this anyway.

Issue #112, opened 2019-10-25
Create z-index stacking variable 


hatched:
In the old GUI we would stack elements in one common place so that we don't end up with weird stacking issues as the application grows in complexity.

https://github.com/juju/juju-gui/blob/d01429be10395b6dc7e627ada71347cf807bdee9/gui/src/app/assets/css/_variables.scss#L85

use:

https://github.com/juju/juju-gui/blob/b309c95462c2c523bd209cc1d0351fbb1b714c01/gui/src/app/assets/css/_modal.scss#L8

PR #164, opened 2019-11-01
Use a z-index scale

barrymcgee:
## Done

- Use a z-index scale
- Mapped values to the greek alphabet? Why the greek alphabet? Because it's arbitrary and doesn't couple the scale to the elements they are being applied to.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check that all page elements look correct across the board

## Details

Fixes: #112


barrymcgee:
@anthonydillon We already use the same z-index across multiple panels and that's fine, not everything needs to stack

If you are proposing we give each component that needs a z-index it's own distinct value, you end up maintaining the bonkers system [we had previously in the GUI](https://github.com/juju/juju-gui/blob/d01429be10395b6dc7e627ada71347cf807bdee9/gui/src/app/assets/css/_variables.scss#L85).

hatched:
This is an interesting approach to the problem. As someone who used the old system for a long time when we had some complex layering I prefer it, but I'll leave this one up to you guys :-)

anthonydillon:
I feel the same as @hatched but happy to disagree and commit to this approach to get us moving.

PR #177, opened 2019-11-04
Reduce info panel width on smaller screens

### LINKED ISSUE #166: Detail page: Adjust width of the topology column
Viewing the details page on my 1440 screen, the topology is similar in width to the [1680 large screen design](https://zpl.io/a8R7QeQ). The design for 1440 has a [narrower column](https://zpl.io/VOpxL7r).

We may not be making the dashboard responsive for Vancouver, but we don't want to be seen to waste space that could be used for the table, especially if the demo screen doesn't have a big resolution.

So I think we should make the column narrower for Vancouver. @clagom?

**Currently**

image

**Narrower design**

image

barrymcgee:
## Done

Reduce info panel width on smaller screens

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to model details
- Start with screen width less than 1681px then increase screen width and view changes with info panel and topology

## Details

Fixes #166

### DIFF
--- a/src/components/InfoPanel/_info-panel.scss
+++ b/src/components/InfoPanel/_info-panel.scss
@@ -12,7 +12,11 @@
 
   &__grid {
     display: grid;
-    grid-template-columns: 1fr 1fr;
+    grid-template-columns: 1fr;
+
+    @media (min-width: $breakpoint-x-large) {
+      grid-template-columns: 1fr 1fr;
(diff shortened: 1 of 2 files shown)


hatched:
QA Note:
The headings are not in the same order as they are in the design. This wasn't introduced here just thought I'd mention it.

PR #186, opened 2019-11-06
Truncate machine error messages

### LINKED ISSUE #173: Details page: Truncate long messages
Truncate long 'messages' in the details page and put the long message in the `title` attribute.

hatched:
## Done

Truncate long machine error messages.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a model with long machine error messages. Contact me if you need one shared with you.

## Details

Fixes #173

### DIFF
--- a/run
+++ b/run
@@ -43,7 +43,7 @@ Commands
 ##
 
 # Define docker images versions
-dev_image="canonicalwebteam/dev:v1.6.4"
+dev_image="canonicalwebteam/dev:v1.6.5"
 if [ -n "${DOCKER_REGISTRY:-}" ]; then
     dev_image="${DOCKER_REGISTRY}/${dev_image}"
 fi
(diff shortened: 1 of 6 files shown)


run
```diff
@@ -43,7 +43,7 @@ Commands
 ##
 
 # Define docker images versions
-dev_image="canonicalwebteam/dev:v1.6.4"
+dev_image="canonicalwebteam/dev:v1.6.5"
```
hatched: The code uses `Object.fromEntries` which requires Node 12. This wasn't an issue before because the test wasn't exercising that code path.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -199,10 +199,12 @@ const generateMachineRows = modelStatusData => {
         { content: splitParts(machine.hardware)["availability-zone"] },
         { content: machine.instanceId },
         {
-          content: generateSpanClass(
-            "u-capitalise",
-            machine.instanceStatus.info.toLowerCase()
```
hatched: This was not valid because these can be long form messages, not just "RUNNING"

PR #197, opened 2019-11-08
Use npm react components

huwshimi:
## Done

- Remove MainTable components.
- Use MainTable from @canonical/react-components.
- Mock imported scss files.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check that the model list and model details pages appear exactly as they did before.


package.json
```diff
@@ -39,6 +39,9 @@
     ]
   },
   "jest": {
+    "moduleNameMapper": {
+      "\\.(css|scss)$": "<rootDir>/__mocks__/styleMock.js"
```
hatched: We haven't had to do this so far, can we investigate why this is necessary and if there are any work arounds before this lands.

src/pages/Models/Details/__snapshots__/ModelDetails.test.js.snap
```diff
@@ -582,13 +582,16 @@ exports[`ModelDetail Container renders the details pane 1`] = `
         >
           <thead>
             <TableRow>
-              <tr>
+              <tr
+                role="row"
```
hatched: It looks like the table in the repository doesn't contain the role removal updates

huwshimi: I didn't remove them because the docs suggest they're required, it's possible that'll change in the next release, in which case we can move them then.

"Using the class p-table--sortable and assigning role="columnheader" and aria-sort to each <th> element will show each table column to be sortable..."

https://docs.vanillaframework.io/base/tables/#sortable

PR #203, opened 2019-11-12
Subordinate units

barrymcgee:
## Done

Added row for subordinate units 

## QA

- Sanity check code
- Check model with known subordinates and verify the correct units are nested


hatched:
QA Note:
The angle element here doesn't appear to be the correct colour or shape. @spencerbygraves do we have an asset for this image?

App: 
Screen Shot 2019-11-13 at 8 58 38 AM

Design: 
Screen Shot 2019-11-13 at 8 59 09 AM

src/pages/Models/Details/ModelDetails.js
```diff
@@ -141,8 +139,38 @@ const generateUnitRows = modelStatusData => {
           { content: unit.workloadStatus.info }
         ]
       });
+
+      const subordinates = unit.subordinates;
+
+      if (subordinates) {
+        Object.keys(subordinates).forEach(key => {
```
hatched: `Object.entries` might be a good method to use here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

spencerbygraves:
@barrymcgee @hatched svg attached.

[unit-tree.svg.zip](https://github.com/canonical-web-and-design/jaas-dashboard/files/3845575/unit-tree.svg.zip)

spencerbygraves:
Can the svg be nudged down a bit when it's added please, so it's vertically centred to the icon on the right:

image

Thanks.

barrymcgee:
I mentioned in my email that we can no way to target the subordinate table row with CSS to override the border-top property as it's a shared component. 

I've created an issue here: https://github.com/canonical-web-and-design/react-components/issues/28

barrymcgee:
The unit tree icon will still look off when viewing the `sub-test` model as the `ubuntu/0` icon is off-centre. The unit tree icon however is dead centre - see:

Screenshot 2019-11-14 at 13 29 43

spencerbygraves:
@barrymcgee looks good, thanks.

hatched:
Looks like this has introduced an issue where some unit icons do not render.

Screen Shot 2019-11-14 at 8 36 04 AM

hatched:
Close :)

in the `hadoopspark` model

Screen Shot 2019-11-14 at 9 26 13 AM

hatched:
It concerned me that there was no changes to the tests here...

I've created another PR to add the data from sub-test to the test data https://github.com/canonical-web-and-design/jaas-dashboard/pull/215

Once this lands you can add a test for the subordinates section

src/static/images/unit-tree.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hatched: This should probably be run through svgo

PR #206, opened 2019-11-13
Offline images

### LINKED ISSUE #101: Assets being served from Assets server should be moved into codebase directly
The dashboard will need to run in a resticted environment and will not be able to talk to third party URLs.

barrymcgee:
## Done

The dashboard could be used in a restricted environment so we shouldn't use the assets server for static assets.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Ensure all images and icons loads as expected

## Details

Fixes #101


src/components/LogIn/LogIn.js
```diff
@@ -2,6 +2,8 @@ import React from "react";
 import { useSelector } from "react-redux";
 import { isLoggedIn } from "app/selectors";
 
+import logo from "static/images/logo/logo-black-on-white.svg";
```
hatched: Curious, why `import` the svg instead of using it just as the link in `src`?

barrymcgee: - Importing images that are less than 10,000 bytes returns a data URI instead of a path (excludes SVGs).
- If the file is renamed or deleted, the app will throw a compilation error
- If the file content changes in the future, Webpack will give it a different name in production so you don’t need to worry about long-term caching of assets.

Ref: https://create-react-app.dev/docs/adding-images-fonts-and-files

PR #216, opened 2019-11-15
Update logout mechanism to allow re-log-in

### LINKED ISSUE #121: Navigation: After logging out you can't log back in
When logging out we clear the bakery to remove the user data but the application doesn't create a new bakery when logging in so you cannot log back in after logging out.

The login/out mechanisms need to be improved to make for easier clearing of user data and more robust checks for logins.

Introduced https://github.com/canonical-web-and-design/jaas-dashboard/pull/111

hatched:
## Done

When logging out clear the identity item out of the bakery instead of removing the bakery. This allows the user to re-log in after logging out.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Log in, log out, navigate around, log in, log out, etc. Try and break it.

## Details

Fixes #121

### DIFF
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -41,7 +41,7 @@ export function storeVisitURL(visitURL) {
   };
 }
 
-export function clearUserData() {
+export function clearBakeryIdentity() {
   return {
     type: actionsList.logOut
   };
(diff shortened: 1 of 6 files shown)


src/app/actions.js
```diff
@@ -51,11 +51,11 @@ export function clearUserData() {
 /**
   Flush localStorage login keys
```
barrymcgee: Comment needs updated.

barrymcgee:
```suggestion
  Flush bakery from redux store
```

PR #217, opened 2019-11-18
Update the login UI to include a connecting state

### LINKED ISSUE #76: Login box jumps left/right in Chrome when changing page resolutions
Seems like a browser bug in Chrome, will check if still an issue before release.

Screen Recording 2019-10-10 at 05 52 pm

### LINKED ISSUE #79: There is a flash of login screen when refreshing UI
See example: https://cl.ly/954187927ca2

hatched:
## Done

Update the login UI to show a "connecting" state when we're connecting to the controller waiting to know if we've logged in yet or not. This always render the application under a mask to reduce the jarring effect of the 'login screen flash' in #79. 

@spencerbygraves The UX has been OK'd by @clagom but I just winged the design. Feel free to throw together some comments here or we can easily update this after it lands.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- It should show 'connecting' as it's waiting for a `visitURL` from the controller. 
- If you're already logged in it should then disappear and you should be logged in.
- If not it'll show a green button to log in. 
- Log in and then log out.
- You should get a green button again to log in.
- Log in and it should work

## Details

Fixes #79 
Fixes #76


src/components/LogIn/LogIn.js
```diff
@@ -17,21 +28,25 @@ const LogIn = ({ children }) => {
     }
   });
 
+  const button = !useSelector(isConnecting)
+    ? generateButton("p-button", "_#", "Connecting...")
```
anthonydillon: What about disabling the button rather then changing the URL to a hashbang

src/components/LogIn/LogIn.js
```diff
@@ -17,21 +28,25 @@ const LogIn = ({ children }) => {
...
-            href={visitURL}
-            rel="noopener noreferrer"
-            target="_blank"
-          >
-            Log in to view your models
-          </a>
+      <>
+        <div className="login">
+          <div className="login__inner">
+            <img className="login__logo" src={logo} alt="JAAS logo"></img>
```
anthonydillon: Images are self closing elements. You don't need the `</img>`

src/components/LogIn/LogIn.js
```diff
@@ -17,21 +28,25 @@ const LogIn = ({ children }) => {
...
-          <a
-            className="p-button--positive"
-            href={visitURL}
-            rel="noopener noreferrer"
-            target="_blank"
-          >
-            Log in to view your models
-          </a>
+      <>
+        <div className="login">
```
anthonydillon: I would prefer if this used `p-card--highlighted` with some positionally utilities but for a follow up.

anthonydillon:
You need to add padding to the login box as you can see it in small screens overlaying the sidebar. Also I don't think we want to see the words. I would suggest added a `is-loading` class to the containers of the app which hides the navigation elements and tables. For extra points you could even at the logo to gray-scale. 

Let me know if you need any help understanding any of this.

src/components/LogIn/LogIn.js
hatched: I've applied this class and made a few other tweaks to make it look better.

src/components/LogIn/LogIn.js
hatched: You can't disable an anchor tag. I've switched it to use a `<button>` when no `visitURL` is present, and the anchor tag when it is.

hatched:
> You need to add padding to the login box as you can see it in small screens overlaying the sidebar.

I'm not really clear what you mean here can you use the newest updates and add a screenshot and highlight the issue.

> Also I don't think we want to see the words. I would suggest added a `is-loading` class to the containers of the app which hides the navigation elements and tables. For extra points you could even at the logo to gray-scale.

There has been some work done towards this design but aiui it's still in flux as it still utilizes a modal to indicate the status (which would never be "loading models"). Until all of the various loading states have been addressed in the design we can use this PR, else I can close it and we can wait.
see: https://app.zeplin.io/project/5d2dfd8e5b67f14c5e1d2aef/screen/5db07eed7b5ffe22b0153c3d

hatched:
> You need to add padding to the login box as you can see it in small screens overlaying the sidebar.

I'm not really clear what you mean here can you use the newest updates and add a screenshot and highlight the issue.

> Also I don't think we want to see the words. I would suggest added a `is-loading` class to the containers of the app which hides the navigation elements and tables. For extra points you could even at the logo to gray-scale.

There has been some work done towards this design but aiui it's still in flux as it still utilizes a modal to indicate the status (which would never be "loading models"). Until all of the various loading states have been addressed in the design we can use this PR, else I can close it and we can wait.
see: https://app.zeplin.io/project/5d2dfd8e5b67f14c5e1d2aef/screen/5db07eed7b5ffe22b0153c3d

clagom:
I wonder about the label "log in to view your models"... Can it be simpler like "login" or "login to the dashboard" or similar? To 'view your models' sounds limited.

Screenshot 2019-11-18 at 15 22 09

hatched:
@clagom I updated the label to be "Log in to the dashboard"

@spencerbygraves I've made the requested changes.

hatched:
@anthonydillon thanks - lets get get it on the design roadmap 👍

hatched:
I need to land this to make a subsequent PR readable.

@spencerbygraves please let me know if there are any further changes you'd like me to make by filing issues 👍

PR #228, opened 2019-11-21
Add relation icons

### LINKED ISSUE #150: Icons also for relation provider - Model detail page
Not sure the API returns this info, but ideally we would have icons also for the relation provider and requirer:

barrymcgee:
## Done

Added icons to the model details relations table and tweaked some of the columns to better fit the real data.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a model that has relations to make sure that the icons for the applications show up.

## Details

Fixes #150

## Screenshots

Screen Shot 2019-12-17 at 3 34 18 PM

### DIFF
--- a/src/pages/Models/Details/ModelDetails.js
+++ b/src/pages/Models/Details/ModelDetails.js
@@ -229,29 +229,27 @@ const generateRelationRows = modelStatusData => {
         {
           content: (
             <>
-              <span>
-                {generateRelationIconImage(
-                  providerApplicationName || peerApplicationName,
-                  modelStatusData
-                )}
-              </span>
-              <span>{provider || peer || "-"}</span>
+              {generateRelationIconImage(
+                providerApplicationName || peerApplicationName,
+                modelStatusData
+              )}
+              {provider || peer || "-"}
             </>
-          )
+          ),
+          className: "u-display--flex"
         },
         {
           content: (
             <>
-              <span>
-                {generateRelationIconImage(
-                  requirerApplicationName,
-                  modelStatusData
-                )}
-              </span>
-              <span>{requirer || "-"}</span>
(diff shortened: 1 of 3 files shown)


hatched:
I'm not totally sure I understand the problem you're running into as the same `modelStatusData` is passed to each table generator...but here is how I'd approach the task:

The relation relation data contains a key `key: "kibana:rest elasticsearch:client"`
this key is in the format of <applicationName>:<interface> and it uses a space as delimiter to determine if it's a `global` or `peer` relation ex) `key: "elasticsearch:peer"`.

So I'd extract the necessary application name for the position where I'd need the icon. Then to find the appropriate `name` for the icon; you would do the same thing that was done with the units.

https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/pages/Models/Details/ModelDetails.js#L88

I'd then abstract the icon generation into its own function and call it from both `generateEntityLink` and from however you're generating the content for the row. 'DRY' ;-)

https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/pages/Models/Details/ModelDetails.js#L62

Hope this helps!

barrymcgee:
@hatch Dumping the relations, I see neither "peer" nor "client" in any of the keys - they all seem to list two relationships, delimited by a `:` between two.. charms? But in some cases the values are duplicated;

e.g. `plugin:resourcemanager resourcemanager:resourcemanager`

Is there any documentation for this API at all?

barrymcgee:
By pulling the first value from each key, I can return about 50% of the icons but the rest 404.

hatched:
No problem @barrymcgee I've added a few commits here which generate the icon and add it to the tables. You can take this now layout and style the tables so that it matches the designs.

PR #257, opened 2019-12-10
Patch the optimization key of Webpack in CRA

### LINKED ISSUE #225: Production build doesn't work
When building the production dist the application doesn't boot.

hatched:
## Done

The mangling being done by Terser in the default CRA Webpack config was causing issues in a few files. To resolve this we use `craco` to patch the webpack config to include the `reserved` configuration value to skip mangling on a few property names. 

Disabling mangling here increases the finished bundle by only 62B.

### Development

- Pull code
- Open http://0.0.0.0:8036/

It should work as before

### Production

- Pull code
- Run `./run exec yarn run build`
- use your webserver of choice to serve the build files. I use `http-server`
  - `yarn global add http-server`
  - `cd build && http-server`

It should work as expected. Try logging out, refreshing, logging back in, etc.

## Details

Fixes #225

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,0 +1,116 @@
+const TerserPlugin = require("terser-webpack-plugin");
+const OptimizeCSSAssetsPlugin = require("optimize-css-assets-webpack-plugin");
+const safePostCssParser = require("postcss-safe-parser");
+const isWsl = require("is-wsl");
+
+module.exports = function({ env }) {
+  // The values below are copied wholesale from the webpack.config.js that
+  // CRA uses to generate it's builds.
+  // The following values have been modified to tweak the minification steps
+  // to support jujulib and the check-auth redux middleware for thunks.
+  // Both jujulib and the check-auth middleware require that the function names
+  // remain intact so we need to tell Terser to not mangle those names.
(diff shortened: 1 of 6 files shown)


craco.config.js
```diff
@@ -0,0 +1,116 @@
...
+                    comparisons: false,
+                    // Disabled because of an issue with Terser breaking valid code:
+                    // https://github.com/facebook/create-react-app/issues/5250
+                    // Pending further investigation:
+                    // https://github.com/terser-js/terser/issues/120
+                    inline: 2
+                  },
+                  mangle: {
+                    safari10: true,
+                    reserved: [
```
hatched: This is the only value changed from the CRA config so it will make for easy CRA upgrades in the future by copying the top level `optimization` key wholesale and then adding this key.

squidsoup: > This is the only value changed from the CRA config so it will make for easy CRA upgrades in the future by copying the top level `optimization` key wholesale and then adding this key.

Would it be worth documenting the process for upgrading CRA in the project's readme?

hatched: Good call, Done.

PR #263, opened 2019-12-20
Create API endpoint for fetching the controller list

### LINKED ISSUE #237: Create jaas http library

hatched:
## Done

Create API endpoint for fetching the controller list.

## Note

This does not work in Firefox, you'll have to use Chrome/Safari to QA. It appears to be around the `OPTIONS` request not having the correct `CORS` headers for redirects. I'll still need to investigate this further.

In Firefox it'll just fail silently with a `CORS` warning message in the console.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- After logging in, watch the redux store. `juju/controllers` should contain the controller list (if you have permissions to do so).

## Details

Fixes #237

### DIFF
--- a/src/app/model-poller.js
+++ b/src/app/model-poller.js
@@ -1,4 +1,8 @@
-import { fetchAllModelStatuses, loginWithBakery } from "juju";
+import {
+  fetchAllModelStatuses,
+  fetchControllerList,
+  loginWithBakery
+} from "juju";
 import { fetchModelList } from "juju/actions";
 import {
   updateControllerConnection,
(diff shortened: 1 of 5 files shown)


src/app/model-poller.js
```diff
@@ -13,6 +17,7 @@ export default async function connectAndListModels(reduxStore, bakery) {
     reduxStore.dispatch(updateControllerConnection(conn));
     reduxStore.dispatch(updateJujuAPIInstance(juju));
     reduxStore.dispatch(updatePingerIntervalId(intervalId));
+    fetchControllerList(reduxStore);
```
hatched: This may be a temporary location for this request.

We may want to gate this on access level so we don't have a silently failing request but we'll also want to use these values in the overall model list for users who have access at the controller level.

anthonydillon:
I get a `UPDATE_CONTROLLER_LIST` event but the model event lifecycle continues to poll updates? Should we pause model updates when on a different view?

hatched:
In order to populate and keep the other columns in the controller view up to date we need to continue to poll for the model data. That will change with the new API as it'll receive updates over the websockets when they happen.

hatched:
Thanks for the review!

PR #265, opened 2019-12-22
Add a basic table and populate it on the Controller listing page

hatched:
## Done

Added a basic table to the controller page populated with real data from a controller listing.

## Note

This does not work in Firefox, you'll have to use Chrome/Safari to QA. It appears to be around the `OPTIONS` request not having the correct `CORS` headers for redirects. 

In Firefox it'll just fail silently with a `CORS` warning message in the console.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View the controllers page, it should list the controllers in your environment if you have permissions to view them.


hatched:
Good call!, Thanks @anthonydillon, done!

PR #270, opened 2020-01-08
Update docker dev image version.

hatched:
## Done

Update the docker image version in the run script.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- It should build and run as expected.

### DIFF
--- a/run
+++ b/run
@@ -43,7 +43,7 @@ Commands
 ##
 
 # Define docker images versions
-dev_image="canonicalwebteam/dev:v1.6.5"
+dev_image="canonicalwebteam/dev:v1.6.7"
 if [ -n "${DOCKER_REGISTRY:-}" ]; then
     dev_image="${DOCKER_REGISTRY}/${dev_image}"
 fi


barrymcgee:
This did work for me but the whole process of deleting the old image, and pulling/downloading and installing all the deps took well over 15 minutes. This was the same when we first started this project but I have since totally nuked my machine to install Catalina. for this reason, I use yar for this project instead of the run script. @nottrobin

hatched:
@nottrobin I'm not sure I understand your request here. You'd just like us to add the gitignore values? I think that's the only thing from that generator that we don't have here. As we don't need the sass-lint. We don't actually need all of those extra gitignore values either...

nottrobin:
@hatched if you actually ran `yo canonical-webteam:run` you'd see what I mean.

The run script has a version you can inspect with `./run version` (or maybe `./run --version`. This version is tied to the version of https://www.npmjs.com/package/generator-canonical-webteam that it was generated from.

So I at least expect the version values in the `run` script to be updated to a version where the docker image is of the version you're trying to change to. Cos we would use the version of the run script to know if the project is up to date with the latest stuff like the docker image version.

hatched:
Ahh I see, yeah I just updated the file from the github UI.

Here is a new PR with the proper update: https://github.com/canonical-web-and-design/jaas-dashboard/pull/274

PR #271, opened 2020-01-08
Add model data counts to top of the controller view

anthonydillon:
## Done
- Add model data counts to top of the controller view
- Capitalise the True or False for public column to match the design

## QA
- Pull code
- Open http://0.0.0.0:8036/
- Go to controllers
- See the counts of machines, applications and units appear at the top 

## Details
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1073


src/app/selectors.js
```diff
@@ -188,6 +188,43 @@ const groupModelsByOwner = modelData => {
...
+      if (machines.hasOwnProperty(machineID)) {
+        machinesCount++;
+      }
+    }
+  }
+  return {
+    machinesCount: machinesCount,
+    applicationCount: applicationCount,
+    unitCount: unitCount
+  };
```
hatched: This can just be 
```javascript
return {
  applicationCount,
  machinesCount,
  unitCount
};
```

src/app/selectors.js
```diff
@@ -188,6 +188,43 @@ const groupModelsByOwner = modelData => {
...
+  let applicationCount = 0;
+  let unitCount = 0;
+
+  for (const modelUUID in modelData) {
+    const applications = modelData[modelUUID].applications;
+    for (const applicationID in applications) {
+      if (applications.hasOwnProperty(applicationID)) {
+        applicationCount++;
+      }
+      const units = applications[applicationID].units;
```
hatched: There isn't really any need to loop through the units to get this value. You can instead do the following and remove the loop. You can do the same for the machines below.

```suggestion
      unitCount = Object.keys(applications[applicationID].units).length;
```

src/app/selectors.js
```diff
@@ -188,6 +188,43 @@ const groupModelsByOwner = modelData => {
...
+      for (const unitID in units) {
+        if (units.hasOwnProperty(unitID)) {
+          unitCount++;
+        }
+      }
+    }
+    const machines = modelData[modelUUID].machines;
```
hatched:
```suggestion
    machinesCount = Object.keys(modelData[modelUUID].machines).length;
```

src/app/selectors.js
```diff
@@ -188,6 +188,43 @@ const groupModelsByOwner = modelData => {
...
+*/
+const countModelData = modelData => {
+  let machinesCount = 0;
+  let applicationCount = 0;
+  let unitCount = 0;
+
+  for (const modelUUID in modelData) {
+    const applications = modelData[modelUUID].applications;
+    for (const applicationID in applications) {
+      if (applications.hasOwnProperty(applicationID)) {
```
hatched: See the comment below for the details about this change.

```javascript
applicationCount = Object.keys(applications).length
```

src/app/selectors.js
```diff
@@ -188,6 +188,43 @@ const groupModelsByOwner = modelData => {
   return grouped;
 };
 
+/**
+  Returns the number of models, applications and units.
```
hatched:
```suggestion
  Returns an object containing the number of models, applications and units.
```

anthonydillon:
@hatched @barrymcgee thanks I've addressed the comments.

anthonydillon:
@hatched sure, I didn't see many test so wasnt sure but I have added some.

src/pages/Controllers/Controllers.js
```diff
@@ -37,19 +40,34 @@ export default function Controllers() {
...
+        { content: `${c.Public}`, className: "u-align--right u-capitalise" }
       ]
     }));
 
   return (
     <Layout>
       <div className="l-content">
+        <div className="overview">
+          <div className="row">
+            <div className="col-4 overview__machines">
```
anthonydillon: I'm not a fan of using class names to target elements in for tests. I believe class names should be reserved for styling. Is there another way to do this? Maybe a data attr for example: `data-test="overview__machines"`.

barrymcgee: I'd prefer this approach also, although think we should use `data-testid` to be explicit. 

Note, however, those who have given this much more thought [0], recommend that targeting via a data attribute is used as a last resort. [1]

However, I like the simplicity of using it consistently everywhere.  

[0] https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change
[1] https://testing-library.com/docs/guide-which-query

src/pages/Controllers/Controllers.test.js
```diff
@@ -37,3 +37,43 @@ describe("Controllers page", () => {
...
+  it("displays the applications count", () => {
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <MemoryRouter>
+        <Provider store={store}>
+          <Controllers />
+        </Provider>
+      </MemoryRouter>
+    );
+    expect(wrapper.find(".overview__applications strong").text()).toBe(
```
hatched: Thanks for the tests 👍 

You can probably just merge these assertions into the single test case :-)

src/pages/Controllers/Controllers.js
hatched: I agree, while I'm not against using classes for targeting if you have already added a specific class to style the target, in this case there isn't really a clean way of accessing it. I came up with something like...

```js
  it("displays the correct counts", () => {
    const wrapper = mount(
      <MemoryRouter>
        <Provider store={store}>
          <Controllers />
        </Provider>
      </MemoryRouter>
    );
    const values = [
      "56 machines",
      "59 applications",
      "58 units"
    ];
    wrapper.find('.overview strong').forEach((ele, index) => expect(ele.text()).toBe(values[index]));
  });
```

Which isn't super resilient to markup changes, which is kind of the whole point of not using classes to query for elements :-D But maybe it's ok in this case...however, maybe in this case using `data-testid` (or whatever string we'd like to standardize on) is more appropriate.

anthonydillon:
@barrymcgee done.

PR #272, opened 2020-01-08
Persist model grouping via URL

### LINKED ISSUE #260: Persist model grouping selection
When selecting a model grouping it should persist between pageviews. I believe the best way to do this would be via query parameter as it may be useful to share filters/groups.

barrymcgee:
## Done

Persist model grouping via URL

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click different grouping toggles and observe URL changes
- Refresh the page and filter should persist

## Details

Fixes #260

### DIFF
--- a/src/components/ModelGroupToggle/ModelGroupToggle.js
+++ b/src/components/ModelGroupToggle/ModelGroupToggle.js
@@ -12,17 +12,11 @@ const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
   const queryStrings = queryString.parse(window.location.search);
   queryStrings.groupedby = groupedBy;
   useEffect(() => {
-    if (groupedBy === "status") {
-      history.push({
-        pathname: "/models",
-        search: null
-      });
-    } else {
-      history.push({
-        pathname: "/models",
-        search: queryString.stringify(queryStrings)
-      });
-    }
+    history.push({
+      pathname: "/models",
+      search:
+        groupedBy === "status" ? null : queryString.stringify(queryStrings)
+    });
     // Including queryStrings to the dependency array will trigger an infinite loop
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [history, groupedBy]);
--- a/src/components/ModelGroupToggle/ModelGroupToggle.js
+++ b/src/components/ModelGroupToggle/ModelGroupToggle.js
@@ -24,7 +24,7 @@ const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
       });
     }
     // Including queryStrings to the dependency array will trigger an infinite loop
-    // eslint-disable-next-line
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [history, groupedBy]);
 
   return (
--- a/src/components/ModelGroupToggle/ModelGroupToggle.js
+++ b/src/components/ModelGroupToggle/ModelGroupToggle.js
@@ -1,24 +1,11 @@
-import React, { useEffect } from "react";
-import { useHistory } from "react-router-dom";
+import React from "react";
 import classNames from "classnames";
-import queryString from "query-string";
 
 import "./_model-group-toggle.scss";
 
 const buttons = ["status", "cloud", "owner"];
 
 const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
-  const history = useHistory();
-  const queryStrings = queryString.parse(history.location.search);
-  queryStrings.groupedby = groupedBy;
-  const newQs = queryString.stringify(queryStrings);
-  useEffect(() => {
-    history.push({
-      pathname: "/models",
-      search: groupedBy === "status" ? null : newQs
-    });
-  }, [history, groupedBy, newQs]);
-
   return (
     <div className="p-model-group-toggle">
       <div className="p-model-group-toggle__inner">
--- a/src/components/ModelGroupToggle/ModelGroupToggle.js
+++ b/src/components/ModelGroupToggle/ModelGroupToggle.js
@@ -1,6 +1,7 @@
 import React, { useEffect } from "react";
 import { useHistory } from "react-router-dom";
 import classNames from "classnames";
+import queryString from "query-string";
 
 import "./_model-group-toggle.scss";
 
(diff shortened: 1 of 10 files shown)


src/components/ModelGroupToggle/ModelGroupToggle.js
```diff
@@ -1,11 +1,18 @@
...
 
 import "./_model-group-toggle.scss";
 
 const buttons = ["status", "cloud", "owner"];
 
 const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
+  const history = useHistory();
+
+  useEffect(() => {
+    history.push(`/models?groupby=${groupedBy}`);
```
hatched: This isn't scalable and won't work with other components interacting with the url. You'll want to instead have it just update the `groupby` query parameter. 

We almost always have a `groupedby` value but we don't need to define a query param if it's grouped by the default value.

src/pages/Models/Models.js
```diff
@@ -21,7 +22,19 @@ function pluralize(value, string) {
...
+
+  const location = useLocation();
+  // Grab filter from 'groupby' query in URL and assign to variable
+  // If it doesn't exist, fall back to grouping by status
+  const groupFilterFromQueryString =
+    location.search.split("groupby=")[1] || "status";
+  // Set variable as default state value
+  const [groupedBy, setGroupedBy] = useState(groupFilterFromQueryString);
+  // Listen for back button click and update groupedBy state accordingly
+  window.onpopstate = function() {
```
hatched: You shouldn't redefine a global in a component as everywhere will use this method now for `onpopstate`.

src/components/ModelGroupToggle/ModelGroupToggle.js
```diff
@@ -1,11 +1,21 @@
...
 
 const buttons = ["status", "cloud", "owner"];
 
 const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
+  const history = useHistory();
+
+  useEffect(() => {
+    history.push({
+      pathname: "/models",
+      search: `?groupby=${groupedBy}`
```
hatched: I think this will have the same problem...what happens if your url is `/models?groupby=owner&filterby=gce` ? 

What you'll need to do is fetch the query param, update the specific element you're interested in then push the updated value... You may find something like https://www.npmjs.com/package/query-string to be helpful - note I did not inspect that package, just adding it as a push off point.

We also don't need to set a groupby query param for the default grouping

anthonydillon: You probably just been to prepend the existing search query?

barrymcgee:
@hatched I've refactored using query string and it works for my use case but seems to hit an infinite loop when hitting multiple query strings e.g. `/models?groupby=owner&filterby=gce`

Can you spot why?

barrymcgee:
@anthonydillon @hatched Fixed that bug with @Caleb-Ellis's help.

Now ready for re-review.

src/pages/Models/Models.js
```diff
@@ -21,7 +23,20 @@ function pluralize(value, string) {
 
 export default function Models() {
   const { blocked, alert, running } = useSelector(getGroupedModelStatusCounts);
-  const [groupedBy, setGroupedBy] = useState("status");
+
+  // Grab filter from 'groupedby' query in URL and assign to variable
+  // If it doesn't exist, fall back to grouping by status
+  const location = useLocation();
+  const queryStrings = queryString.parse(location.search);
+  const getGroupedByFilter = queryStrings.groupedby || "status";
```
hatched: the name `getGroupedByFilter` seems like it's a method...maybe this should just be `groupedBy` as the value here and the value of state on 34 should be the same.

src/components/ModelGroupToggle/ModelGroupToggle.test.js
```diff
@@ -1,23 +1,34 @@
...
-import { shallow } from "enzyme";
+import { mount } from "enzyme";
+import { MemoryRouter } from "react-router-dom";
 
 import ModelGroupToggle from "./ModelGroupToggle";
 
 describe("Model group toggle", () => {
   it("renders without crashing and matches snapshot", () => {
-    const wrapper = shallow(<ModelGroupToggle />);
+    const wrapper = mount(
```
hatched: These updated tests create unusable snapshot updates....

https://github.com/canonical-web-and-design/jaas-dashboard/issues/277

src/pages/Models/Models.test.js
```diff
@@ -36,4 +37,28 @@ describe("Models page", () => {
       "17 models: 2 blocked, 4 alerts, 11 running"
     );
   });
+
+  it("displays correct grouping view", () => {
+    global.window = { location: { pathname: "/models?groupedby=owner" } };
```
hatched: Because you're using the `useLocation` hook you should be able to use the `MemoryRouter` here instead with an `initialEntries` defined

https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/components/InfoPanel/InfoPanel.test.js#L19

src/pages/Models/Models.test.js
```diff
@@ -36,4 +37,28 @@ describe("Models page", () => {
...
+      wrapper.find(".p-model-group-toggle__button.is-selected").text()
+    ).toBe("status");
+    wrapper.find("button[value='owner']").simulate("click", {
+      target: { value: "owner" }
+    });
+    wrapper.update();
+    expect(
+      wrapper.find(".p-model-group-toggle__button.is-selected").text()
+    ).toBe("owner");
+    expect(global.window.location.search).toEqual("?groupedby=owner");
```
hatched: Using the `MemoryRouter` above I'm not sure how you would inspect that this value gets updated... maybe you can inspect `MemoryRouter` and it'll have a `location` object?

src/components/ModelGroupToggle/ModelGroupToggle.js
```diff
@@ -1,11 +1,32 @@
...
+import { useHistory } from "react-router-dom";
 import classNames from "classnames";
+import queryString from "query-string";
 
 import "./_model-group-toggle.scss";
 
 const buttons = ["status", "cloud", "owner"];
 
 const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
+  const history = useHistory();
```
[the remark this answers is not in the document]
anthonydillon: Yeah I agree, but this could happen in a follow up branch. This branch is getting old now.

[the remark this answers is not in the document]
hatched: That's fine, if that's the case just create an issue so we don't forget.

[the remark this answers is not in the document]
barrymcgee: I was already 2/3 there with this refactor so just finished it.

src/pages/Models/Models.test.js
```diff
@@ -36,4 +37,29 @@ describe("Models page", () => {
...
+      wrapper.find(".p-model-group-toggle__button.is-selected").text()
+    ).toBe("status");
+    wrapper.find("button[value='owner']").simulate("click", {
+      target: { value: "owner" }
+    });
+    wrapper.update();
+    expect(
+      wrapper.find(".p-model-group-toggle__button.is-selected").text()
+    ).toBe("owner");
+    const searchParams = new URLSearchParams(history.location.search);
```
hatched: Cool, I didn't know about this one!

PR #292, opened 2020-01-17
Populate filter panel with real data from models

### LINKED ISSUE #124: Generate filter UI elements from model data
The items in the filter UI are static but should instead be generated by the values in the `modelData`.

barrymcgee:
## Done

Populate filter panel with real data from models

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click filter box in the header
- Verify all those real filters

## Details

Fixes #124

### DIFF
--- a/src/components/FilterTags/FilterTags.js
+++ b/src/components/FilterTags/FilterTags.js
@@ -15,15 +15,16 @@ const FilterTags = () => {
   const [filterPanelVisibility, setfilterPanelVisibility] = useState(false);
   const node = useRef();
 
-  const filters = {
-    cloud: [],
-    owner: [],
-    region: [],
-    credential: [],
-    controller: []
-  };
+  const filters = {};
 
   const modelData = useSelector(getModelData);
+
+  const addFilter = function(type, value) {
+    filters[type] = filters[type] || [];
+    if (!filters[type].includes(value)) {
+      filters[type].push(value);
+    }
+  };
 
   // Loop the model data and pull out the available filters
   Object.values(modelData).forEach(model => {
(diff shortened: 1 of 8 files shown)


src/components/FilterTags/FilterTags.js
```diff
@@ -39,6 +87,10 @@ const FilterTags = () => {
     };
   }, []);
 
+  const handleFilterClick = () => {
+    console.log("Filter clicked");
```
barrymcgee: I know this is left here, the next ticket is to add the functionality.

src/components/FilterTags/FilterTags.js
```diff
@@ -54,42 +106,32 @@ const FilterTags = () => {
...
-          </ul>
-        </div>
-        <div className="p-filter-panel__section">
-          <h4 className="p-filter-panel__heading">Controller</h4>
-          <ul className="p-list p-filter-panel__list">
-            <li className="p-filter-panel__item">prodstack-1</li>
-            <li className="p-filter-panel__item">prodstack-2</li>
-          </ul>
-        </div>
+        {Object.keys(filters).map(filterBy => {
```
hatched: It would be nice if this had an 'empty' message while the model data is loading.

src/components/FilterTags/FilterTags.js
```diff
@@ -1,12 +1,60 @@
...
+*/
+  const addFilter = function(type, value) {
+    filters[type] = filters[type] || [];
+    if (!filters[type].includes(value)) {
+      filters[type].push(value);
+    }
+  };
+
+  // Loop the model data and pull out the available filters
+  Object.values(modelData).forEach(model => {
```
hatched: What do you think about moving this filter list generation into a selector? 🤷‍♂

src/pages/Models/Models.js
```diff
@@ -52,7 +55,7 @@ export default function Models() {
             )}, ${running} running`}
           </div>
           <ModelGroupToggle setGroupedBy={setGroupedBy} groupedBy={groupedBy} />
-          <FilterTags />
+          <FilterTags modelData={modelData} />
```
hatched: This isn't used.

hatched:
Can you align the filter box to the input box on the right side?

Screen Shot 2020-01-17 at 3 45 44 PM

src/components/FilterTags/FilterTags.js
barrymcgee: Can you explain the benefit?

hatched: I suppose it would be easier to cache the output. But I'm not convinced, hence the shrug :):

PR #300, opened 2020-01-21
Render application icons to the Topology

### LINKED ISSUE #238: Render applications onto topology

hatched:
## Done

- Application and subordinate icons are rendered onto the Topology.
- Annotation data has been requested and stored in the model status.
- The application icons have not yet been positioned, that'll come in a follow-up.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a model and 'inspect' the canvas, you should see a number of `g` elements. One for each application.

## Details

Fixes #238


src/components/Topology/Topology.js
```diff
@@ -0,0 +1,65 @@
...
+  useEffect(() => {
+    const zoom = d3.zoom();
+
+    const topo = d3
+      .select(ref.current)
+      .attr("width", width)
+      .attr("height", height)
+      .append("g")
+      .call(zoom)
+      // 2 lines below copied....why do they work?
```
hatched: magnets, how do they work?

barrymcgee: https://developer.mozilla.org/en-US/docs/Web/CSS/transform

src/components/Topology/Topology.js
```diff
@@ -0,0 +1,65 @@
...
+      .attr("cx", d => (isSubordinate(d) ? 65 : 95))
+      .attr("cy", d => (isSubordinate(d) ? 65 : 95))
+      .attr("r", d => (isSubordinate(d) ? 60 : 90))
+      .attr("fill", "#f5f5f5")
+      .attr("stroke-width", 1)
+      .attr("stroke", "#888888");
+
+    appIcon
+      .append("image")
+      .attr("xlink:href", d => generateIconPath(d.charm.replace("cs:", "")))
```
barrymcgee: This replace is also used in four different places in ModelDetails.js - can you pull it out to a util function please?

src/components/Topology/Topology.js
```diff
@@ -0,0 +1,65 @@
...
+      .attr("width", 96)
+      .attr("height", 96)
+      .attr("transform", d =>
+        isSubordinate(d) ? "translate(17, 17)" : "translate(47, 47)"
+      );
+
+    // console.log(topo.selectAll(".application").nodes());
```
barrymcgee: Remove console

src/juju/index.js
```diff
@@ -109,6 +110,22 @@ async function fetchModelStatus(modelUUID, getState) {
       if (isLoggedIn(getState())) {
         status = await conn.facades.client.fullStatus();
       }
+      if (isLoggedIn(getState())) {
+        const entities = Object.keys(status.applications).map(name => ({
+          tag: `application-${name}`
+        }));
+        const response = await conn.facades.annotations.get({ entities });
+        // It will return an entry for every entity even if there are no
+        // annotations so we have to inspect them and strip out the empty.
```
barrymcgee: empty... what?

src/components/Topology/Topology.js
hatched: It was a comment to remind me to investigate how `zoom.transform` and `d3.zoomIdentity` work :)

The second line is also a transform attribute on the svg itself not a css transform.

src/components/Topology/Topology.js
hatched: Good call, I've expanded `generateIconPath` to support stripping it if it's provided.

src/juju/index.js
hatched: entries :)

barrymcgee:
@hatched Why didn't you pull the `.replace("cs", "")`  function out into a util function so it can be reused in ModelDetails as I suggested?

hatched:
I made the link generator more resilient so that you can pass the id's with or without the `cs:` prefix no util is needed for such a simple action. Looks like I missed removing `.replace("cs", "")` in a few places. I'll do that.

PR #316, opened 2020-01-28
Add filtering input interactions

barrymcgee:
## Done

Add filtering input interactions

- Still need to add tests

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click the filter panel and add and remove filters

### DIFF
--- a/src/components/FilterTags/FilterTags.js
+++ b/src/components/FilterTags/FilterTags.js
@@ -14,6 +14,7 @@ import "./_filter-tags.scss";
 const FilterTags = () => {
   const [filterPanelVisibility, setFilterPanelVisibility] = useState(false);
   const [activeFilters, setActiveFilters] = useState({});
+
   const node = useRef();
   const filters = {};
   const modelData = useSelector(getModelData);
@@ -87,6 +88,12 @@ const FilterTags = () => {
     };
   }, []);
 
+  /**
+  Apply a given filter
+  @param {object} e The event object
+  @param {string} filter The name of the filter
+   @param {string} filterBy The name of the filter group
+*/
   const addActiveFilter = (e, filter, filterBy) => {
     setActiveFilters(filters => {
       const updatedFilters = { ...filters };
@@ -99,6 +106,12 @@ const FilterTags = () => {
     e.currentTarget.classList.add("is-selected");
   };
 
+  /**
+  Remove a given filter
+  @param {object} e The event object
+  @param {string} filter The name of the filter
+   @param {string} filterBy The name of the filter group
+*/
   const removeActiveFilter = (e, filter, filterBy) => {
     e.stopPropagation();
     setActiveFilters(filters => {
@@ -107,12 +120,12 @@ const FilterTags = () => {
       };
       if (updatedFilters[filterBy].includes(filter)) {
         const index = updatedFilters[filterBy].indexOf(filter);
-        delete updatedFilters[filterBy][index];
+        updatedFilters[filterBy].splice(index, 1);
       }
       return updatedFilters;
     });
 
-    // Remove is-selected from filter panel
+    // Remove is-selected state from relevant filter from filter panel
     const panelFilterItems = document.querySelectorAll(
       ".p-filter-panel__button"
     );
@@ -123,7 +136,10 @@ const FilterTags = () => {
     });
   };
 
-  console.log(activeFilters);
+  let activeFiltersCount = 0;
+  Object.values(activeFilters).forEach(filterGroup => {
+    activeFiltersCount = activeFiltersCount + filterGroup.length;
+  });
 
   return (
     <div className="p-filter-tags" ref={node}>
@@ -153,8 +169,11 @@ const FilterTags = () => {
               );
             });
           })}
-        {Object.entries(activeFilters).length < 1 && (
-          <span>Filter models:</span>
+        {activeFiltersCount === 0 && <span>Filter models:</span>}
+        {activeFiltersCount > 0 && (
+          <span className="p-filter-tags__active-count">
+            +{activeFiltersCount}
+          </span>
         )}
       </div>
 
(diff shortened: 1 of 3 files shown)


src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
-    console.log("Filter clicked");
+  /**
+  Apply a given filter
+  @param {object} e The event object
+  @param {string} filter The name of the filter
+   @param {string} filterBy The name of the filter group
+*/
+  const addActiveFilter = (e, filter, filterBy) => {
+    setActiveFilters(filters => {
+      const updatedFilters = { ...filters };
```
hatched: Using a `set` might be more appropriate here.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

src/components/FilterTags/FilterTags.js
```diff
@@ -121,7 +197,7 @@ const FilterTags = () => {
                   {filters[filterBy].map(filter => (
                     <li key={filter} className="p-filter-panel__item">
                       <button
-                        onClick={e => handleFilterClick(e)}
+                        onClick={e => addActiveFilter(e, filter, filterBy)}
```
hatched: I'd expect to be able to click the filter again to remove it from the set so the original name might be better.

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
+                    Remove
+                  </i>
+                </span>
+              );
+            });
+          })}
+        {activeFiltersCount === 0 && <span>Filter models:</span>}
+        {activeFiltersCount > 0 && (
+          <span className="p-filter-tags__active-count">
+            +{activeFiltersCount}
```
hatched: IIRC one of the designs have this being clickable to view the overflowed filters? I imagine that's coming later which may by why you're looping through all the filters now?

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
-          className="p-filter-tags__input"
-          onFocus={() => setfilterPanelVisibility(true)}
-        />
-      </form>
+      <div
+        type="text"
+        className="p-filter-tags__input"
+        onClick={() => setFilterPanelVisibility(!filterPanelVisibility)}
+      >
+        {Object.entries(activeFilters).length > 0 &&
```
hatched: If `activeFilters` is a set or an array then this can be simplified a lot.

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
+          Object.entries(activeFilters).map(activeFilterObj => {
+            return Object.values(activeFilterObj[1]).map(activeFilter => {
+              return (
+                <span
+                  className="p-filter-tags__active-filter"
+                  key={activeFilterObj[0] + activeFilter}
+                >
+                  {activeFilterObj[0]}: {activeFilter}{" "}
+                  <i
+                    className="p-icon--close"
```
hatched: QA Issue:
Screen Shot 2020-01-27 at 11 22 34 PM

The close button can end up under the overflow count. The 'input' should probably end before the count element to avoid this.

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
-        />
-      </form>
+      <div
+        type="text"
+        className="p-filter-tags__input"
+        onClick={() => setFilterPanelVisibility(!filterPanelVisibility)}
+      >
+        {Object.entries(activeFilters).length > 0 &&
+          Object.entries(activeFilters).map(activeFilterObj => {
+            return Object.values(activeFilterObj[1]).map(activeFilter => {
```
hatched: If we're only going to show a single element then you don't really need to loop at all, just take the last one.

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
+      ".p-filter-panel__button"
+    );
+    panelFilterItems.forEach(panelFilterItem => {
+      if (panelFilterItem.innerHTML === filter) {
+        panelFilterItem.classList.remove("is-selected");
+      }
+    });
+  };
+
+  let activeFiltersCount = 0;
```
hatched: If you simplify the storage of the filters this can be simplified to just `activeFilters.size`

src/components/FilterTags/FilterTags.js
```diff
@@ -87,20 +88,95 @@ const FilterTags = () => {
...
+    e.currentTarget.classList.add("is-selected");
   };
 
+  /**
+  Remove a given filter
+  @param {object} e The event object
+  @param {string} filter The name of the filter
+   @param {string} filterBy The name of the filter group
+*/
+  const removeActiveFilter = (e, filter, filterBy) => {
```
hatched: if the filter value contained the group and the value then it would simplify this logic because you could just push them into a set/array and then extract the group/value when rendering or parsing.

For example 
```js
const filters = new Set(['cloud:google', 'credential:secret', ...]);
``` 

A set guarantees unique entries and it's iterable so you can loop through the set to generate the elements and set selections to it without worrying about duplicates.

src/components/FilterTags/FilterTags.js
barrymcgee: The filter dropdown panel shows all selected filters via the selected state.

src/components/FilterTags/FilterTags.js
barrymcgee: In some cases, we can show two and in theory, with more responsive refinements, we should be able to expand the input on wider screens to show three or four.

src/components/FilterTags/FilterTags.js
hatched: Ahh ok, that's not how the design shows, but understandable.

PR #318, opened 2020-01-28
Add eslint and fix config

### LINKED ISSUE #317: Fix eslint config
When doing the Yarn2 spike I noticed that our eslint config was misconfigured and we were relying on eslint being installed as a side effect of another package.

hatched:
## Done

Eslint was being used as a side effect of another package install. This explicitly defines it in the package.json and updates the config file to work correctly.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The application should work as expected and CI should pass.

## Details

Fixes #317

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,24 @@
 {
-  "extends": ["react-app", "prettier/react"],
-  "plugins": ["prettier", "jest"],
+  "extends": [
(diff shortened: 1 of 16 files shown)


hatched:
I'm going to close this now and propose a config to `practices` where we can have a single common shared config and discuss changes to it there.

Issue #321, opened 2020-01-29
t.annotations is undefined


hatched:
On staging (and only staging) you're not able to view the model details page because of an error t.annotations is undefined.

It appears that we need to be more resilient in our check here: 
https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/components/Topology/Topology.js#L117

However the question is why it only happens on the staging build.

PR #322, opened 2020-01-29
Set browserslist production to latest

hatched:
## Done

There is an issue on staging #321 which I'm not able to debug because the build transpiles the async await code making it not possible to step through the code. 

This can be reverted or modified later once #321 has been resolved.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The app should work as expected, but staging after landing will show the change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,9 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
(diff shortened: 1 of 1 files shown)


hatched:
This change reduced the total bundle size by 5KB.

PR #331, opened 2020-01-30
Fix back button navigation with model grouping

### LINKED ISSUE #330: Handle browser history
When on the models listing the jump through the grouping then use the back button in the broswer and the UI should change to reflect the URL:

You can use the popstate method to detect those URL changes and make UI changes as needed.

```js
window.addEventListener('popstate', function (event) {
	// The URL changed...
});
```
You can access that property on the event object.

```js
window.addEventListener('popstate', function (event) {
	// Log the state data to the console
	console.log(event.state);
});
```

hatched:
## Done

When using the back button after toggling the model grouping it now correctly updates the UI.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Play with the model grouping buttons, they should update the UI as expected
- Hit the back button, it should navigate back through the groupings.
- Refresh with a grouping query string in the URL, it should render that grouping.

## Details

Fixes #330

### DIFF
--- a/src/pages/Models/Models.js
+++ b/src/pages/Models/Models.js
@@ -22,23 +22,22 @@ export default function Models() {
   const queryStrings = queryString.parse(location.search);
   // If it doesn't exist, fall back to grouping by status
   const groupedByFilter = queryStrings.groupedby || "status";
-  // Set initial state using filter from URL
-  const [groupedBy, setGroupedBy] = useState(groupedByFilter);
-  // Assign updated filter back to queryStrings obj
-  queryStrings.groupedby = groupedBy;
-  // Stringify the obj again to push back to URL
-  const updatedQs = queryString.stringify(queryStrings);
-  // Add as an effect so UI is updated on each component render (e.g. Back button)
-  useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/pages/Models/Models.js
```diff
@@ -22,19 +22,21 @@ export default function Models() {
...
-  // Set initial state using filter from URL
-  const [groupedBy, setGroupedBy] = useState(groupedByFilter);
-  // Assign updated filter back to queryStrings obj
-  queryStrings.groupedby = groupedBy;
-  // Stringify the obj again to push back to URL
-  const updatedQs = queryString.stringify(queryStrings);
-  // Add as an effect so UI is updated on each component render (e.g. Back button)
-  useEffect(() => {
+
+  const [groupModelsBy, setGroupModelsBy] = useState(groupedByFilter);
```
hatched: @barrymcgee FYI Even though `groupedByFilter` was changing with the back button interactions, because the initial state of the state was already set it didn't update the `groupModelsBy` state. So the following 3 lines are used to do just that.

PR #349, opened 2020-02-05
Filter entities by application on model details

### LINKED ISSUE #148: Filter functionality when clicking on an app / unit / machine / relation - Model detail view
Not sure it's for the POC, but the functionality on click should be as design:

hatched:
## Done

When clicking an application on the model details page it'll now filter down the units, machines, and relations that are that application.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a model details and click one of the applications. It should filter the units, machines, and relations to only show ones for that application.
- View a model details with subordinates, be sure that it also highlights subordinates properly.
- View a model details with subordinates that aren't on each parent unit, be sure that it only shows units with subordinates.

Contact me if you don't have models that match these criteria.

## Details

Fixes #148 

## Screenshots

Screen Shot 2020-02-05 at 2 29 55 PM


src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -35,7 +35,9 @@ describe("Info Panel", () => {
     const store = mockStore(dataDump);
     const wrapper = mount(
       <Provider store={store}>
-        <MemoryRouter initialEntries={["/models/group-test"]}>
+        <MemoryRouter
+          initialEntries={["/models/activedev@external/group-test"]}
```
hatched: Driveby to fix an error output in the test console output.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -14,317 +15,94 @@ import { getModelUUID, getModelStatus } from "app/selectors";
...
+  relationTableHeaders,
+  generateApplicationRows,
+  generateMachineRows,
+  generateRelationRows,
+  generateUnitRows
+} from "./generators";
 
 import "./_model-details.scss";
 
-const applicationTableHeaders = [
```
hatched: All of this removed code was moved to the `generators.js` file wholesale.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -14,317 +15,94 @@ import { getModelUUID, getModelStatus } from "app/selectors";
...
-
-const generateApplicationRows = modelStatusData => {
-  if (!modelStatusData) {
-    return [];
+/**
+  Returns the modelStatusData filtered by the supplied values.
+  @param {Object} modelStatusData The model status data to filter
+  @param {String} appName The name of the application to filter the data by.
+*/
+const filterModelStatusData = (modelStatusData, appName) => {
```
hatched: This is the newly added function, unfortunately the diff makes it very hard to read, it might be best to view the new file instead.

src/pages/Models/Details/_model-details.scss
```diff
@@ -4,34 +4,31 @@
 .model-details {
   display: grid;
   grid-template-columns: 230px 1fr;
+  margin-bottom: 2rem;
   padding-bottom: 14rem;
 
-  @media (min-width: $breakpoint-m-large) {
```
hatched: The stylelint kind of went crazy with this file. It might be best to view the commits for the changes directly. They are limited to some background colours and a hover.

src/pages/Models/Details/generators.js
```diff
@@ -0,0 +1,318 @@
...
+          className: "u-capitalise"
+        },
+        { content: "-", className: "u-align--right" },
+        { content: "-", className: "u-align--right" },
+        { content: "CharmHub" },
+        { content: key.split("-")[-1] || "-", className: "u-align--right" },
+        { content: "Ubuntu" },
+        { content: "-" }
+      ],
+      className: filterByApp === key ? "is-selected" : "",
```
hatched: These functions were all copied over from the other file with the exception of these few lines on each table generation to add the necessary data for the filtering selection.

src/pages/Models/Details/_model-details.scss
```diff
@@ -4,34 +4,31 @@
...
-    text-align: center;
-  }
+  tbody tr {
+    @include vf-animation(background, brisk, ease-in-out);
 
-  tr {
-    display: grid;
+    &.is-selected,
+    &.is-selected:hover {
+      background-color: #cef3ff;
```
barrymcgee: Any good reason we don't use a Vanilla color here?

hatched: This colour wasn't defined in the settings overrides and I didn't see it here: https://docs.vanillaframework.io/settings/color-settings so I left it. Which variable should I define/override for it?

hatched: I merged this, we can address this fix in a follow-up.

Issue #125, opened 2019-10-30
Filter model list by the item selected in the filter UI


hatched:
When clicking an item in the filter UI then filter the model list.

Blocked by https://github.com/canonical-web-and-design/jaas-dashboard/issues/124

hatched:
@cristinadresch why was this closed? It hasn't been implemented yet.

clagom:
Is this merged, can be closed?

PR #351, opened 2020-02-07
Add filtering interactions to drop-down panel

barrymcgee:
## Done

Add filtering interactions to drop-down panel

## QA

- Pull code
- Open http://0.0.0.0:8036/

## Details

Related #125

### DIFF
--- a/src/components/FilterTags/FilterTags.js
+++ b/src/components/FilterTags/FilterTags.js
@@ -136,18 +136,23 @@ const FilterTags = () => {
     });
   };
 
-  let activeFiltersCount = 0;
-  Object.values(activeFilters).forEach(filterGroup => {
-    activeFiltersCount = activeFiltersCount + filterGroup.length;
-  });
-
   return (
     <div className="p-filter-tags" ref={node}>
-      <div
+      <input
         type="text"
         className="p-filter-tags__input"
         onClick={() => setFilterPanelVisibility(!filterPanelVisibility)}
+        placeholder="Filter models"
+      />
+
+      <div
+        className={classNames("p-card--highlighted p-filter-panel", {
+          "is-visible": filterPanelVisibility
+        })}
       >
+
+        {Object.entries(filters).length <= 0 && <p>Loading filters...</p>}
+
         {Object.entries(activeFilters).length > 0 &&
           Object.entries(activeFilters).map(activeFilterObj => {
             return Object.values(activeFilterObj[1]).map(activeFilter => {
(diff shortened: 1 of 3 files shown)


src/components/FilterTags/FilterTags.js
```diff
@@ -87,30 +88,89 @@ const FilterTags = () => {
...
+      />
+
       <div
         className={classNames("p-card--highlighted p-filter-panel", {
           "is-visible": filterPanelVisibility
         })}
       >
         {Object.entries(filters).length <= 0 && <p>Loading filters...</p>}
+
+        {activeFilters.length > 0 && (
```
hatched: I feel like this section should always be visible as it jumps quite a bit the first time it becomes active. It also acts as a hint to the user what to do. It could even have a message when empty "Click a filter" or something. @clagom

src/components/FilterTags/FilterTags.js
```diff
@@ -87,30 +88,89 @@ const FilterTags = () => {
...
-          type="text"
-          placeholder="Filter models"
-          className="p-filter-tags__input"
-          onFocus={() => setfilterPanelVisibility(true)}
-        />
-      </form>
+      <input
+        type="text"
+        className="p-filter-tags__input"
+        onClick={() => setFilterPanelVisibility(!filterPanelVisibility)}
```
hatched: With this change when you click the input again once it's opened it's closed...Was that intentional? Or should it always be open when the input is in focus? (I feel like it should be the latter)

barrymcgee: Makes sense that it no longer toggles now that clicking anywhere else plus using the Escape key will close the panel.

hatched: So lets say I click the input to open the dropdown, I select a button, then want to edit the input, when I click the input again it closes the dropdown. I know this doesn't work right now, but I'd imagine this would be a valid usecase when it does :)

src/components/FilterTags/FilterTags.js
barrymcgee: Ok, I'll add it in all the time and @clagom can review on staging..

src/components/FilterTags/FilterTags.js
barrymcgee: Re-read the above, I agree with you ;)

hatched: Oh, yes.... :-D how embarrassing

PR #352, opened 2020-02-10
Add query params for active filters on the models view

barrymcgee:
## Done

Add query params for active filters on the models view

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Add and remove filters and verify that URL reflects correct changes.

## Details

Related #148

### DIFF
--- a/src/components/FilterTags/FilterTags.js
+++ b/src/components/FilterTags/FilterTags.js
@@ -155,24 +155,21 @@ const FilterTags = () => {
       >
         {Object.entries(filters).length <= 0 && <p>Loading filters...</p>}
         <div className="p-filter-panel__section" data-test="selected">
-          <h4 className="p-filter-panel__heading">Selected</h4>
+          {Object.entries(filters).length > 0 && (
+            <h4 className="p-filter-panel__heading">Selected</h4>
+          )}
           {activeFilters.length > 0 &&
-            activeFilters.map(activeFilter => {
-              return (
-                <span
-                  className="p-filter-tags__active-filter"
-                  key={activeFilter}
+            activeFilters.map(activeFilter => (
+              <span className="p-filter-tags__active-filter" key={activeFilter}>
+                {activeFilter}
+                <i
+                  className="p-icon--close"
+                  onClick={() => removeActiveFilter(activeFilter)}
                 >
-                  {activeFilter}
-                  <i
-                    className="p-icon--close"
-                    onClick={() => removeActiveFilter(activeFilter)}
-                  >
-                    Remove
-                  </i>
-                </span>
-              );
-            })}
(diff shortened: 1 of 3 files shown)


src/components/FilterTags/FilterTags.test.js
```diff
@@ -132,4 +133,31 @@ describe("Filter pills", () => {
...
+    );
+    const firstFilterButton = ".p-filter-panel__button";
+    const selectedActiveFilterClose =
+      ".p-filter-tags__active-filter .p-icon--close";
+    wrapper
+      .find(firstFilterButton)
+      .first()
+      .simulate("click");
+    const searchParams = new URLSearchParams(history.location.search);
+    expect(searchParams.get("activeFilters")).toEqual("cloud: google");
```
hatched: The URL would be a bit nicer if there weren't spaces in it...

hatched:
The filters aren't being applied from the URL on a refresh. Is this going to be a follow-up task? It'll be interesting because it's possible the URL will contain filters that are invalid but we don't know they are invalid until we've pulled the status for the models.

hatched:
QA Issue:
I think this is caused by the removal of the space....maybe just revert that change for now and we can clean it up later. The fix is to make sure you add the space back in after the `:` and then ensure you have the proper format of `<group>:<space><value>`.

Anyways, if you click a few filters then refresh the page, the filters will show as selected but the associating button won't be selected because it's missing a space and as such won't match.

barrymcgee:
@hatched I've reverted that change.

src/components/FilterTags/FilterTags.js
```diff
@@ -88,6 +98,15 @@ const FilterTags = () => {
...
   }, []);
 
+  // Update query params when adding and removing filters
+  useEffect(() => {
+    const queryParams = queryString.parse(location.search);
+    queryParams.activeFilters = activeFilters;
+    history.push({
+      search: queryString.stringify(queryParams)
+    });
+  }, [activeFilters, history, location.search]);
```
hatched: It's a bit weird here that the `location.search` is a 'read-dependency' but you're also setting it which should create an infinite loop, it doesn't though so that's good :) We also don't actually care if the history object changes here. The only variable we actually care about is the `activeFilters`. 

Instead of using `useEffect` here, it might be more appropriate to create a wrapper function around `setActiveFilters` which handles updating the query string like here: https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/pages/Models/Models.js#L31

src/components/FilterTags/FilterTags.test.js
hatched: Can you expand this to include a case where there are two params?

PR #356, opened 2020-02-12
Add google analytics tracking

### LINKED ISSUE #341: Integrate GA

barrymcgee:
## Done

- Add Google Analytics tracking
- Add GA event for filter panel visibility
- Add GA event for user panel visibility
- Add GA event for user expanding topology

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code
-  Remove the `if (process.env.NODE_ENV === "production")` conditional
- Open demo link and verify yourself in realtime in GA dashboard

## Details

Screenshot 2020-02-12 at 12 32 27

Fixes #341


src/components/App/App.js
```diff
@@ -12,6 +13,10 @@ import NotFound from "pages/NotFound/NotFound";
 const baseURL = process.env.REACT_APP_BASE_APP_URL;
 
 function App() {
+  if (process.env.NODE_ENV === "production") {
+    ReactGA.initialize("UA-1018242-68");
+    ReactGA.pageview(window.location.pathname + window.location.search);
```
hatched: We probably also want the hash so a more resilient approach might be to do...Unless we don't, in which case ignore this :-)
```suggestion
    ReactGA.pageview(window.location.href.replace(window.location.origin, ''));
```

src/components/FilterTags/FilterTags.js
```diff
@@ -145,7 +146,15 @@ const FilterTags = () => {
       <input
         type="text"
         className="p-filter-tags__input"
-        onClick={() => setFilterPanelVisibility(true)}
+        onClick={() => {
+          setFilterPanelVisibility(true);
+          if (filterPanelVisibility) {
+            ReactGA.event({
+              category: "User",
```
hatched: We should have a common place where we list the available categories.

src/components/FilterTags/FilterTags.js
```diff
@@ -145,7 +146,15 @@ const FilterTags = () => {
       <input
         type="text"
         className="p-filter-tags__input"
-        onClick={() => setFilterPanelVisibility(true)}
+        onClick={() => {
+          setFilterPanelVisibility(true);
+          if (filterPanelVisibility) {
```
hatched: The previous line you're setting the `filterPanelVisibility` to `true` so this check is not necessary as it should always be called.

src/components/InfoPanel/InfoPanel.js
```diff
@@ -53,7 +54,15 @@ const InfoPanel = () => {
           <Topology width={300} height={300} modelData={modelStatusData} />
           <i
             className="p-icon--expand"
-            onClick={() => setShowExpandedTopology(!showExpandedTopology)}
+            onClick={() => {
+              setShowExpandedTopology(!showExpandedTopology);
+              if (showExpandedTopology) {
```
hatched: This should be the same value as being set on 58, yes? As at this point in the code the `showExpandedTopology` will be the old value.

src/components/UserIcon/UserIcon.js
```diff
@@ -48,7 +49,15 @@ export default function User() {
   return (
     <div
       className="user-icon"
-      onClick={() => setUserPanelVisibility(!userPanelVisibility)}
+      onClick={() => {
+        setUserPanelVisibility(!userPanelVisibility);
+        if (userPanelVisibility) {
```
hatched: See above comment about this value.

src/components/FilterTags/FilterTags.js
barrymcgee: A task for further down the line

src/components/InfoPanel/InfoPanel.js
barrymcgee: Because it's not in the callback, good catch.

src/components/InfoPanel/InfoPanel.js
```diff
@@ -53,7 +54,13 @@ const InfoPanel = () => {
           <Topology width={300} height={300} modelData={modelStatusData} />
           <i
             className="p-icon--expand"
-            onClick={() => setShowExpandedTopology(!showExpandedTopology)}
+            onClick={() => {
+              setShowExpandedTopology(!showExpandedTopology);
+              ReactGA.event({
```
hatched: This is triggering that it expanded even if it's closing. You'll want to add a conditional here.

src/components/UserIcon/UserIcon.js
```diff
@@ -45,10 +46,21 @@ export default function User() {
...
+        action: "Opened user panel"
+      });
+    }
+  }, [userPanelVisibility]);
+
   return (
     <div
       className="user-icon"
-      onClick={() => setUserPanelVisibility(!userPanelVisibility)}
+      onClick={() => {
```
hatched: You can remove the brackets here.

src/components/InfoPanel/InfoPanel.js
barrymcgee: Negative - the close button is different.

PR #359, opened 2020-02-13
⚠️ WIP: Filter on click

barrymcgee:
Why isn't the filtered model data clone being returned?


hatched:
Discussed offline and proposed a solution.

PR #361, opened 2020-02-13
Add Settings component

### LINKED ISSUE #342: create Settings component
This component will be used for toggling on and off settings like GA tracking and more configuration options in the future.

hatched:
## Done

- Add Settings page.
- Add option to disable analytics.

## QA

- Pull code
- Run `./run clean && ./run exec yarn run build`
- Run `cd build && serve` (You may have to install this `npm i serve`.
- Open the path serve returns.
- Navigate around a bit.
- You notice analytics events being sent in the devtools. You can filter the requests by 'analytic'
- Click the "Settings" link in the user drop down and it should take you to the settings page.
- Toggle the option to disable analytics
- `window.localStorage.getItem('disableAnalytics')` should be `"true"`
- Refresh
- The analytics should no longer be sent
- Go to settings and re-enable
- `window.localStorage.getItem('disableAnalytics')` should be `"false"`
- Refresh
- You should see analytics again.
- Click the demo link below and view devtools network pane and see that no analytics events are sent in dev builds.

## Details

Fixes #342

## Screenshots

Screen Shot 2020-02-13 at 4 12 24 PM

### DIFF
--- a/src/app/send-analytics-hook.js
+++ b/src/app/send-analytics-hook.js
@@ -1,0 +1,22 @@
+import ReactGA from "react-ga";
+
+export default function useSendAnalytics() {
+  return ({ path, category, action }) => {
+    const disableAnalytics = localStorage.getItem("disableAnalytics");
+    if (
+      process.env.NODE_ENV !== "production" ||
+      (disableAnalytics !== "false" && disableAnalytics !== undefined)
+    ) {
+      return;
+    }
+
+    if (path) {
+      ReactGA.pageview(path);
+    } else {
+      ReactGA.event({
+        category,
+        action
+      });
+    }
+  };
+}
--- a/src/components/UserIcon/UserIcon.js
+++ b/src/components/UserIcon/UserIcon.js
@@ -2,9 +2,10 @@ import React, { useState, useRef, useEffect } from "react";
 import classNames from "classnames";
 import { Link } from "react-router-dom";
 import { useDispatch, useStore } from "react-redux";
-import ReactGA from "react-ga";
 
 import { logOut } from "app/actions";
+
+import useSendAnalytics from "app/send-analytics-hook";
 
 import "./_user-icon.scss";
 
@@ -12,6 +13,8 @@ export default function User() {
   const [userPanelVisibility, setUserPanelVisibility] = useState(false);
   const dispatch = useDispatch();
   const node = useRef();
+
+  const sendAnalytics = useSendAnalytics();
 
   useEffect(() => {
     const closePanel = () => {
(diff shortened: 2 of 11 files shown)


src/components/UserIcon/UserIcon.js
```diff
@@ -47,12 +50,13 @@ export default function User() {
   const getState = useStore().getState;
 
   useEffect(() => {
-    if (userPanelVisibility)
-      ReactGA.event({
+    if (userPanelVisibility) {
```
barrymcgee: Screenshot 2020-02-14 at 15 47 21

src/pages/Settings/Settings.js
```diff
@@ -0,0 +1,43 @@
...
+                  localStorage.setItem(
+                    "disableAnalytics",
+                    e.currentTarget.checked
+                  );
+                }}
+              />
+              <div className="p-switch__slider"></div>
+            </div>
+          </label>
+          <Notification>
```
barrymcgee: Can we move this notification above the toggle?

src/pages/Settings/Settings.js
```diff
@@ -0,0 +1,43 @@
...
+import Header from "components/Header/Header";
+import Layout from "components/Layout/Layout";
+
+import "./settings.scss";
+
+export default function Settings() {
+  return (
+    <Layout>
+      <Header>
+        <span className="l-content settings__header">Settings</span>
```
barrymcgee: This should be a H1 element

src/app/send-analytics-hook.js
```diff
@@ -0,0 +1,22 @@
+import ReactGA from "react-ga";
+
+export default function useSendAnalytics() {
+  return ({ path, category, action }) => {
+    const disableAnalytics = localStorage.getItem("disableAnalytics");
+    if (
+      process.env.NODE_ENV !== "production" ||
+      (disableAnalytics !== "false" && disableAnalytics !== undefined)
```
barrymcgee: Aren't these both falsy? Could this be simplified to;

```
      if(!disableAnalytics){ }
```

barrymcgee:
QA issue: https://share.getcloudapp.com/7KuRJe5j

src/pages/Settings/Settings.js
hatched: I have to override almost every style on the H1 to make it look good. Without it's like...

Screen Shot 2020-02-14 at 11 20 49 AM

src/pages/Settings/Settings.js
hatched: done

src/components/UserIcon/UserIcon.js
hatched: 🤣 They are different :D

hatched:
As discussed offline the QA issue is caused by the browser rendering of the css animations getting confused as the underlying hidden toggle is correctly selected. This only happens if you rapidly toggle it on and off so we may be able to mitigate it a bit by shortening the animation in vanilla but I don't really expect this to cause any issues in the real world as it's not consistent even if you try and trigger it.

src/app/send-analytics-hook.js
hatched: Technically yes however I wanted it to be explicit because it's not obvious that `localStorage` values are always strings.

PR #372, opened 2020-02-26
Add active user name to settings header

### LINKED ISSUE #365: Account in User Menu should be replaced with user name

barrymcgee:
## Done

Add active user name to settings header

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Settings header should display your username

## Details

Fixes #365

### DIFF
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -438,8 +438,7 @@ export const isConnecting = state => !!state.root.visitURL;
   @returns {String} The users userTag.
 */
 export const getActiveUserTag = state =>
-  state.root.controllerConnection &&
-  state.root.controllerConnection.info.user.identity;
+  state?.root?.controllerConnection?.info.user.identity;
 
 /**
   Returns a model status for the supplied modelUUID.
--- a/src/components/UserMenu/UserMenu.js
+++ b/src/components/UserMenu/UserMenu.js
@@ -33,36 +33,34 @@ const UserMenu = () => {
         "is-expanded": isExpanded
       })}
     >
-      {activeUser && (
-        <>
-          <div
-            className="user-menu__header"
-            onClick={() => setIsExpanded(!isExpanded)}
-          >
-            <i className="p-icon--user"></i>
-            <span className="user-menu__name">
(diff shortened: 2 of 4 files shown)


src/app/selectors.js
```diff
@@ -438,8 +438,7 @@ export const isConnecting = state => !!state.root.visitURL;
   @returns {String} The users userTag.
 */
 export const getActiveUserTag = state =>
-  state.root.controllerConnection &&
-  state.root.controllerConnection.info.user.identity;
+  state?.root?.controllerConnection?.info.user.identity;
```
barrymcgee: Ooooh.. optional chaining 🌟

hatched: Too bad the output version of this is huge compared to the previous version :D

src/components/UserMenu/UserMenu.js
```diff
@@ -29,31 +33,36 @@ const UserMenu = () => {
...
-          <Link className="user-menu__link" to="/settings">
-            Settings
-          </Link>
-        </li>
-        <li>
-          <Link
-            className="user-menu__link"
-            to="/"
-            onClick={() => dispatch(logOut(getState))}
+      {activeUser && (
```
hatched: I'm not sure it's a great idea to gate on this value. I don't think we want to not render the entire user menu if their name isn't defined. A better approach might be to always render it but then have `activeUser` fall back to "Account" or something.

barrymcgee: The problem with that is you then get a flicker of "Account" on page load. When would a user login and not have a username?

hatched: Ahh I see, the issue is that it might be possible for the `identity` for the user to not be defined depending on the identity provider. It's not likely, but if so then they are prevented from logging out in this case. 

If you'd like you can just leave it an empty string then, that way no name will be shown and no flash, but the menu will still work.

src/app/selectors.js
barrymcgee: Party pooper! 👎🏻

PR #378, opened 2020-03-04
Add external config

### LINKED ISSUE #353: Create config.js template for configuration values

hatched:
## Done

Converted the config values that were defined in the `.env` file into a `config.js` file so that it can be generated or modified by other hosts, specifically Juju core.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The app should work as normal

## Details

Fixes #353


public/config.js
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line no-unused-vars
```
hatched: eslint disable line required as CRA doesn't respect the `.eslintrc` configs.

public/config.js
```diff
@@ -0,0 +1,9 @@
...
+var jaasDashboardConfig = {
```
hatched: It's a `var global` so that it can be re-declared later if need be.

PR #379, opened 2020-03-12
Add Usabilia link to primary nav

barrymcgee:
## Done

Add Usabilia to primary nav

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click the link in the sidebar and verify pop-up

## Details

Fixes: https://github.com/canonical-web-and-design/juju-squad/issues/1076


src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -168,10 +181,23 @@ const PrimaryNav = () => {
...
           </li>
+          <li className="p-list__item">
+            <a
+              rel="noopener noreferrer"
+              target="_blank"
+              className="p-list__link"
+              href="#_"
+              onClick={e => {
+                e.preventDefault();
+                window.usabilla_live("click");
```
anthonydillon: Is there a way to check if `window.usabilla_live` exists before displaying the link? The function was blocked by ad-blocker.

spencerbygraves:
@anthonydillon we discussed the font sizes in the [other issue](url), I thought it made sense to wait for the vanilla side nav. But as both you and @clagom have mentioned it, I don't mind if we make those 3 links small, as long as @barrymcgee is happy to do it.

anthonydillon:
@spencerbygraves I was thinking the navigation would look better if all links were the larger size and the horizontal rule would be enough to indicate the secondary level. (Just my thought)

src/components/PrimaryNav/PrimaryNav.js
hatched: This is an interesting problem as I wonder the percentage of users that will block this link but will also want to provide feedback. Maybe usabilla has a link with query params that we can use instead of their script?

barrymcgee: investigated this and couldn't find a timely resolution. 

@anthonydillon If you want to investigate further, can you raise a ticket?

hatched: I created the issue as we should definitely address this as it's just a broken link for many. 

https://github.com/canonical-web-and-design/jaas-dashboard/issues/383

PR #386, opened 2020-03-19
Hide status filters in Model Details

barrymcgee:
## Done

Hide status filters in Model Details

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View model details and verify status filters are hidden


src/pages/Models/Details/ModelDetails.js
```diff
@@ -168,6 +168,7 @@ const ModelDetails = () => {
               filterToggle={viewFilterToggle}
             />
             <Filter
```
hatched: I'm not sure this approach is any better than just removing this component here, even if temporary.

PR #394, opened 2020-03-26
Resize terminal

barrymcgee:
## Done

Added ability to drag to resize terminal on model details pages

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to model details page and drag the top of terminal window

### DIFF
--- a/src/components/Terminal/Terminal.js
+++ b/src/components/Terminal/Terminal.js
@@ -2,7 +2,7 @@ import React, { useEffect, useRef, useState } from "react";
 import { useHistory } from "react-router-dom";
 import { useSelector } from "react-redux";
 import { getMacaroons } from "app/selectors";
-import classNames from "classnames";
+import { FitAddon } from "xterm-addon-fit";
 
 import cleanUpTerminal from "./cleanup-terminal";
 import setupTerminal from "./setup-terminal";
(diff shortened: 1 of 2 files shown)


anthonydillon:
@barrymcgee pushed a fix remove the event listener on mouseup. The issue was the listeners did not match. The way js knows it's the same is the combination of the target and the listener function signature.

src/components/Terminal/Terminal.js
```diff
@@ -23,40 +31,79 @@ const Terminal = ({ address, modelName }) => {
...
+        document.addEventListener("mousemove", resize);
+      }
+    };
+
+    const removeResizeListener = () => {
+      document.removeEventListener("mousemove", resize);
+    };
+
+    // set resize listeners
+    window.addEventListener("mousedown", addResizeListener);
```
hatched: Why are these events not registered using the react event system? That way it should handle attaching and detaching and simplify this event handling code.

```
<div className="p-terminal__header" onMouseUp={...} onMouseDown={...}>
```

src/components/Terminal/Terminal.js
```diff
@@ -23,40 +31,79 @@ const Terminal = ({ address, modelName }) => {
...
+    };
     // Because the user can switch the model details UI from within the
     // component the `modelName` passed above will change as the UI
     // updates. This causes the terminal to log in again which is
     // undesirable.
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const [terminalVisible, setTerminalVisible] = useState(false);
+  useEffect(() => {
```
hatched: Can you add a comment that this `useEffect` order is important?

hatched:
QA Notes:

When you click on the handle to drag it, it jumps to the bottom of the header. It would be nice if we could enhance the logic here so that it would drag from where the user clicks instead of jumping. Can you create a follow-up issue for this please.

src/components/Terminal/Terminal.js
barrymcgee: Issue raised: https://github.com/canonical-web-and-design/jaas-dashboard/issues/397

PR #395, opened 2020-03-27
Support URL basename

huwshimi:
## Done
- Replace the low level Router with BrowserRouter to support basename.

## QA

- Build and run using Juju, the app should load and the URL should be prefixed by the basename.
- To test this locally: 
- - Open App.js and change line 40 from `<Router basename={baseAppURL} ...` to `<Router basename="/dashboard" ..`.
- - Load the app, it should load and the URL should start with /dashboard/...

### DIFF
--- a/src/components/Routes/Routes.js
+++ b/src/components/Routes/Routes.js
@@ -1,5 +1,5 @@
-import React from "react";
-import { Route, Redirect } from "react-router-dom";
+import React, { useEffect } from "react";
+import { Route, Redirect, useLocation } from "react-router-dom";
 
 import Controllers from "pages/Controllers/Controllers";
 import Logs from "pages/Logs/Logs";
@@ -7,6 +7,7 @@ import Models from "pages/Models/Models";
 import ModelDetails from "pages/Models/Details/ModelDetails";
 import Settings from "pages/Settings/Settings";
 import Usage from "pages/Usage/Usage";
+import useSendAnalytics from "app/send-analytics-hook";
 
 export const paths = {
   "/": { redirect: "/models" },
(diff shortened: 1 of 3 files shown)


src/components/Routes/Routes.js
```diff
@@ -19,6 +20,16 @@ export const paths = {
 };
 
 export function Routes() {
+  const sendAnalytics = useSendAnalytics();
+  let location = useLocation();
+
+  useEffect(() => {
+    // Send an analytics event when the URL changes.
+    sendAnalytics({
+      path: window.location.href.replace(window.location.origin, "")
```
hatched: Above location isn't used, did you mean to use it here instead of `window.location`?

src/components/Routes/Routes.js
```diff
@@ -19,6 +20,16 @@ export const paths = {
 };
 
 export function Routes() {
+  const sendAnalytics = useSendAnalytics();
+  let location = useLocation();
```
hatched:
```suggestion
  const location = useLocation();
```

src/components/Routes/Routes.js
huwshimi: This was copied over from the history listener in App.js. I presume it was done this way because you want the analytics to get the full location, not the location that the router returns (which from memory excludes the basename).

src/components/Routes/Routes.js
huwshimi: Whoops, thanks.

src/components/Routes/Routes.js
hatched: Ahh yes of course, thanks for reminding me! :)

PR #407, opened 2020-04-01
Remove the Prettier overrides and auto-update code

hatched:
## Done

Remove the Prettier overrides and auto-update code.
This added trailing commas to all arrays and objects as well as wrapping the arguments in arrow functions with parens.

This is their default and it's how we do it in the other projects so just keeping it consistent.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- It should just work as usual


hatched:
> More parens!

giphy

PR #409, opened 2020-04-02
Close topology on Escape key click

barrymcgee:
## Done

- Close topology on Escape key click
- Add error text to title of topology as I had instances when refreshing where model would not load but topology modal was still onscreen
- Add conditional so that expand button would not appear unless modelName exists

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to models details page 
- Open topology
- Hit Escape key
- Verify topology closes

### DIFF
--- a/src/components/InfoPanel/InfoPanel.js
+++ b/src/components/InfoPanel/InfoPanel.js
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useMemo, useState, useEffect } from "react";
 import { useSelector } from "react-redux";
 import { useParams } from "react-router-dom";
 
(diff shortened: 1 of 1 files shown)


src/components/InfoPanel/InfoPanel.js
```diff
@@ -49,12 +49,25 @@ const InfoPanel = () => {
 
   const sendAnalytics = useAnalytics();
 
+  // Close topology, if open, on Escape key press
+  useEffect(() => {
+    const closeOnEscape = function (e) {
```
hatched, line 54: "function" what is this? 2018?!? 😆

src/components/InfoPanel/InfoPanel.js
```diff
@@ -49,12 +49,25 @@ const InfoPanel = () => {
...
+      if (e.key === "Escape" && showExpandedTopology) {
```
hatched, line 55: in `FilterTags.js` we use `e.code === "Escape"` can we standardize plz.

src/components/InfoPanel/InfoPanel.js
```diff
@@ -49,12 +49,25 @@ const InfoPanel = () => {
...
+    };
+  });
+
   return (
     <div className="info-panel">
       {showExpandedTopology ? (
         <Modal
           close={() => setShowExpandedTopology(false)}
-          title={modelName.split("/")[1]}
+          title={modelName.split("/")[1] || "Error: model not loading"}
```
hatched, line 70: This looks good but maybe we should also put a conditional around the expand button so that the modal can't even be seen if we don't have the model data.

PR #416, opened 2020-04-03
Set maximum terminal height on drag

### LINKED ISSUE #415: Dragging Terminal up can result in it slipping under the page header

barrymcgee:
## Done

Set maximum terminal height on drag

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to model details page
- Drag terminal up
- See that it doesn't obscure page header

## Details

Fix #415

### DIFF
--- a/src/components/Terminal/Terminal.js
+++ b/src/components/Terminal/Terminal.js
@@ -15,12 +15,13 @@ const Terminal = ({ address, modelName }) => {
   const history = useHistory();
 
   // save default terminal height in the state object
-  let [terminalHeight, setTerminalHeight] = useState(200);
+  const [terminalHeight, setTerminalHeight] = useState(200);
 
   const fitAddonRef = useRef({ current: null });
 
   const terminalHeaderHeight = 40; // px
   const minimumTerminalHeight = 84; // px
+  const modelDetailHeaderHeight = 42; // px
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


src/components/Terminal/Terminal.js
```diff
@@ -47,7 +48,12 @@ const Terminal = ({ address, modelName }) => {
       const mousePosition = e.clientY;
       const newTerminalHeight =
         viewPortHeight - mousePosition + terminalHeaderHeight;
-      if (newTerminalHeight >= minimumTerminalHeight) {
+
+      const maximumTerminalHeight = viewPortHeight - modelDetailHeaderHeight;
+      if (
+        newTerminalHeight < maximumTerminalHeight &&
+        newTerminalHeight >= minimumTerminalHeight
```
anthonydillon, line 55: Instead of this you could use `max` and `min` when setting the `newTerminalHeight` variable. For example:

``` js
const proposedTerminalHeight = viewPortHeight - mousePosition + terminalHeaderHeight;
const newTerminalHeight = Math.min(Math.max(proposedTerminalHeight, minimumTerminalHeight), maximumTerminalHeight);
```

barrymcgee: @anthonydillon Yes, that would also work but in terms of preferring human-readable code, I'd prefer to stick with what I've got?

PR #431, opened 2020-04-16
Swap logo to Juju under Juju

### LINKED ISSUE #355: When running in Juju use Juju logos

hatched:
## Done

When deployed in Juju then change the logo to "Juju" from "JAAS"

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The logo should still be "JAAS"
- Change the configuration value `isJuju` in `config.js` to `true` and refresh
- The logo should now be "Juju"

## Details

fixes #355 

## Screenshots

Screen Shot 2020-04-16 at 1 52 28 PM


spencerbygraves:
@hatched I think this should have had a design review. Your screenshot is different to what the logo should look like. The Juju logo is the bottom one here:

image

Can we try again with this please?

Thanks.

hatched:
@spencerbygraves this simply swaps the JAAS text for Juju. The spacing is like that to make room for the collapsible sidebar. We can probably animate the gap down when the sidebar is open.

spencerbygraves:
@hatched we really can't have broken logos. In each case the correct spacing is the bottom logo here:

image

Is there a way to swap out the correct full logo in each case, and the orange circle when collapsed?

PR #432, opened 2020-04-16
Update renovate config to sync with other projects.

hatched:
## Done

Update renovate config to sync with other projects.

## QA

None necessary


hatched:
Thanks @squidsoup . I think it's ok to drop that devDeps rule as I'd rather have all major bumps treated separately as they could introduce breaking api changes.

PR #436, opened 2020-04-17
Disable controller UUID masking

hatched:
## Done

- If a user is admin and connected to JAAS then disable the controller UUID masking. 
- Adds JIMM facade for `disableControllerUUIDMasking` and `listControllers`.

## QA

You need to be an administrator on a controller so contact me to run you through the demo.


craco.config.js
```diff
@@ -55,15 +55,16 @@ module.exports = function({ env }) {
                       "ClientV2",
                       "ModelManagerV5",
                       "PingerV1",
+                      "JIMMV1",
```
hatched, line 58: This is the only real change to this file. The rest are just `prettier` changes.

src/app/jimm-facade.js
```diff
@@ -0,0 +1,56 @@
+const {
```
hatched, line 1: This file follows the same format as the jujulib auto-generated code which is why it looks a bit out of place. This will change with the next version of jujulib which will be coming early next cycle.

anthonydillon:
I think it's valuable to have someone run though a QA for this to gain knowledge.

barrymcgee:
@hatched Please ping me when you can walk me through this.

hatched:
I'm closing this as #437 includes all the same code and supersedes this branch.

PR #444, opened 2020-04-22
Create a link to go back to the old GUI 

hatched:
## Done

Added a link to the sidebar that opens a modal guiding users to the old GUI on JAAS. The link will only show up in JAAS unless we get confirmation that 2.8 will contain both UI's then we will expand this to handle that use case.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- You should see a link to switch back to the old GUI
- Click the link and a modal should pop up
- Click the link to the old GUI and it should work.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1167
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1172

## Screenshots

Screen Shot 2020-04-22 at 1 36 45 PM
Screen Shot 2020-04-22 at 3 20 52 PM

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.js
+++ b/src/components/PrimaryNav/PrimaryNav.js
@@ -9,6 +9,7 @@ import { externalNavActive } from "ui/actions";
 import { isExternalNavActive } from "ui/selectors";
 
 import UserMenu from "components/UserMenu/UserMenu";
+import Modal from "@canonical/react-components/dist/components/Modal";
 
 // Image imports
 import logoMark from "static/images/logo/logo-mark.svg";
@@ -47,6 +48,7 @@ const pages = [
 
 const PrimaryNav = () => {
   const [activeLinkValue, setActiveLinkValue] = useState("");
+  const [showSwitchModal, setShowSwitchModal] = useState(false);
   const extNavActive = useSelector(isExternalNavActive);
   const { blocked } = useSelector(getGroupedModelStatusCounts);
   const location = useLocation();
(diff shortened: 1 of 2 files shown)


src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -78,6 +80,43 @@ const PrimaryNav = () => {
...
+          title={"Switch back to the old Juju GUI"}
+        >
+          <p>
+            We're sorry to see you go, we would really appreciate it if you
+            could take a moment to tell us why you prefer the old GUI using the
+            methods below.
+          </p>
+          <p>
+            <a
+              className="p-list__link"
```
anthonydillon, line 95: This class wont do anything outside of its parent block `p-list`. You don't need it or the display block in the scss. Just put both links in separate `p`

src/components/PrimaryNav/PrimaryNav.js
```diff
@@ -212,11 +251,24 @@ const PrimaryNav = () => {
...
-            <span className="version">Version 0.0.2</span>
-            <span className="p-label--new">Alpha</span>
+            <span className="version">Version 0.1.0</span>
+            <span className="p-label--new">Beta</span>
           </li>
         </ul>
       </div>
+      {!isJuju ? (
+        <div className="p-primary-nav__bottom">
+          <li className="p-list__item">
```
anthonydillon, line 261: Missing `ul`. I think you just need a `p` here

src/components/PrimaryNav/_primary-nav.scss
```diff
@@ -56,6 +65,10 @@
     }
   }
 
+  .p-modal .p-list__link {
+    display: block;
```
anthonydillon, line 69: You should be able to remove this once you add `p` tags.

src/components/PrimaryNav/PrimaryNav.js
hatched: Done

src/components/PrimaryNav/PrimaryNav.js
hatched: I need to keep the `li` for the same styling, but I added a wrapper `ul`

src/components/PrimaryNav/_primary-nav.scss
hatched: yep removed.

barrymcgee:
Screenshot 2020-04-22 at 23 16 31

Layering modals is weird - the app modal should close before Usabilia opens on top of it?

hatched:
Unfortunately we can't know if the Usabilla link works as there doesn't appear to be a way to test if it has been loaded due to the use of `lightningjs` So we would only be test if the user clicked on a link. If we did this and Usabilla wasn't loaded then our modal would close and nothing would replace it. I figured this was the lesser of two evils.

barrymcgee:
Whereas now, if Usabilia fails to load, the modal will stay open but the link will just do nothing when the user clicks, which is a broken experience anyway? If we have a broken experience regardless, we should prioritise the happy path.

hatched:
The other side of it is once they submit the form on Usabilla if it works, then they will still likely want to head to the old GUI. This keeps it open for them to do so.

barrymcgee:
That's a weird journey but this is ultimately a design call @spencerbygraves

spencerbygraves:
Could we do this?

1. Click the "back to GUI" link
2. Load the feedback modal above the old GUI
3. If they click the Usabilla link, close our modal
4. If Usabilla doesn't load, they can still click the site feedback tab to try again

I think this might be @clagom's territory though 🙂

clagom:
+1 @spencerbygraves , I think that's reasonable

PR #445, opened 2020-04-22
Release banner and beta label

barrymcgee:
## Done

- Add a notification to layout with dismissal persisting to cookie

Note: will add blog and discourse links when available

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify that release notification appears
- Dismiss notification
- Refresh page and notification should not appear again

Screenshot 2020-04-22 at 22 49 35

### DIFF
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -1,5 +1,6 @@
 import React, { useEffect, useState, useRef } from "react";
 import { useSelector } from "react-redux";
+import Notification from "@canonical/react-components/dist/components/Notification/Notification";
 import PrimaryNav from "components/PrimaryNav/PrimaryNav";
 import classNames from "classnames";
 import useHover from "hooks/useHover";
@@ -49,7 +50,24 @@ const Layout = ({ children }) => {
         <PrimaryNav />
       </div>
       <main className="l-main" id="main-content">
-        {children}
+        <div data-test="main-children">{children}</div>
+        {!localStorage.getItem("release20_04") && (
+          <Notification
+            type="information"
+            close={() => {
+              localStorage.setItem("release20_04", true);
+            }}
+          >
+            Welcome the new JAAS Dashboard! This dashboard replaces the previous
+            Juju GUI here on jaas.ai and on your local Juju Controller from Juju
+            2.8.{" "}
+            <span className="u-hide--small">
+              Read more and join the discussion about this new product{" "}
+              <a href="#_">on Discourse</a>. We would love to hear your
+              feedback!
+            </span>
+          </Notification>
+        )}
       </main>
     </div>
   );
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -58,8 +58,8 @@ const Layout = ({ children }) => {
               localStorage.setItem("release20_04", true);
             }}
           >
-            Welcome the new JAAS Dashboard! This dashboard replaces the previous
-            Juju GUI here on jaas.ai and on your local Juju Controller from Juju
+            Welcome the new JAAS Dashboard! This dashboard is the replacement
+            for the Juju GUI in JAAS and individual Juju Controllers from Juju
             2.8.{" "}
             <span className="u-hide--small">
               Read more and join the discussion about this new product{" "}
--- a/src/components/Layout/Layout.test.js
+++ b/src/components/Layout/Layout.test.js
@@ -29,8 +29,8 @@ describe("Layout", () => {
         </Router>
       </Provider>
     );
-    expect(wrapper.find("#main-content").html()).toStrictEqual(
-      `<main class="l-main" id="main-content">content</main>`
+    expect(wrapper.find("[data-test='main-children']").html()).toStrictEqual(
+      `<div data-test="main-children">content</div>`
     );
   });
 });
--- a/src/components/Layout/_layout.scss
+++ b/src/components/Layout/_layout.scss
@@ -18,6 +18,20 @@
     padding-left: $sidebar-collapsed;
     width: 100vw;
   }
+
+  .p-notification--information {
+    bottom: 0;
+    margin: 0 1rem 1rem;
+    position: fixed;
+    width: calc(100vw - 18rem);
+    z-index: z("kappa");
+
+    @at-root {
+      .is-collapsed + .l-main .p-notification--information {
+        width: calc(100vw - 5rem);
+      }
+    }
+  }
 }
 
 .has-collapsible-sidebar .is-collapsed + .l-main {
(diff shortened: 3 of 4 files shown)


hatched:
This appears to be a duplicate of 444.

src/components/Layout/Layout.js
```diff
@@ -1,17 +1,21 @@
...
 import { isSidebarCollapsible } from "ui/selectors";
 
+import { useCookies } from "react-cookie";
+
 import "./_layout.scss";
 
 const Layout = ({ children }) => {
   const [sidebarRef, isSidebarHovered] = useHover();
   const [screenWidth, setScreenWidth] = useState(getViewportWidth());
+  const [cookies, setCookies] = useCookies(["release20_04"]);
```
hatched, line 18: Lets just use `localstorage` then we don't need to add another dependency. You can see how to use it here: https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/pages/Settings/Settings.js#L30

src/components/Layout/Layout.js
```diff
@@ -49,7 +53,23 @@ const Layout = ({ children }) => {
         <PrimaryNav />
       </div>
       <main className="l-main" id="main-content">
-        {children}
+        <div data-test="main-children">{children}</div>
+        {!cookies.release20_04 && (
+          <Notification
```
hatched, line 58: Just out of curiosity why did you have to wrap children in a div? Were you not able to have `<Notification>` along side `{children}`?

src/components/Layout/Layout.js
```diff
@@ -49,7 +53,23 @@ const Layout = ({ children }) => {
...
+            type="information"
+            close={() => {
+              setCookies("release20_04", true);
+            }}
+          >
+            This is the new JAAS Dashboard, replacing the previous Juju GUI here
```
hatched, line 64:
```suggestion
            Thanks for trying the new JAAS Dashboard! This dashboard is replacing the Juju GUI
```

src/components/Layout/Layout.js
```diff
@@ -49,7 +53,23 @@ const Layout = ({ children }) => {
...
+            on jaas.ai and on your local Juju Controller from Juju 2.8.{" "}
+            <span className="u-hide--small">
```
hatched, line 66: What is this for? It doesn't appear to have any effect.

src/components/Layout/Layout.js
```diff
@@ -49,7 +53,23 @@ const Layout = ({ children }) => {
...
+              Read more about this new product <a href="#_">on our blog</a> or{" "}
+              <a href="#_">discuss it on Discourse</a>. We would love to hear
```
hatched, line 68: Can you create an issue to update these so we don't forget?

src/components/Layout/Layout.js
barrymcgee: I'm using a cookie as Safari now nukes localStorage every 7 days - we don't want this reappearing every seven days.

src/components/Layout/Layout.js
barrymcgee: So I could target only the children with a test. The alternative was to include the notification text in the test which was too brittle.

src/components/Layout/Layout.js
hatched: I was under the impression that was only for PWA's and not general browsing. Can you find something to indicate otherwise?

src/components/Layout/Layout.js
barrymcgee: Reduce your screen width to less than 620px wide

src/components/Layout/Layout.js
barrymcgee: It's the opposite of that - it doesn't apply to PWAs added to the home screen but does apply to all sites accessed via Safari https://www.theregister.co.uk/2020/03/26/apple_relax_were_not_totally/

src/components/Layout/Layout.js
barrymcgee: Text as per @clagom's instruction https://codepen.io/clagom/pen/LYpxYMe?editable=true%3Dhttps%3A%2F%2Fcodepen.io%2Fembed%2Fprefill%2F

src/components/Layout/Layout.js
hatched: Ahh interesting, however the kicker is this part `...if the user does not interact with the associated website during this period.` So a regular user of the application would not have their storage cleared.

It also appears that cookies have had the same limitation since Feb 19'. So I feel like my request would still be the better way forward.

src/components/Layout/Layout.js
hatched: Well that instruction is incorrect 😆  At the very least we can't say "here on jaas.ai" as this message will also be shown when it's not served from jaas.ai :) (it's also a comma splice ;) )

src/components/Layout/Layout.js
hatched: Ahh I see, Do you not think those links are valuable if you're on a mobile device?

hatched:
It looks like the QA issue above is only visible when the sidebar is open.

src/components/Layout/Layout.js
barrymcgee: They're less valuable than the first line and the alternative obscures the entire content area.

src/components/Layout/Layout.js
barrymcgee: They're less valuable than the line above.

src/components/Layout/Layout.js
```diff
@@ -49,7 +50,24 @@ const Layout = ({ children }) => {
...
-        {children}
+        <div data-test="main-children">{children}</div>
+        {!localStorage.getItem("release20_04") && (
+          <Notification
+            type="information"
+            close={() => {
+              localStorage.setItem("release20_04", true);
+            }}
+          >
+            Welcome the new JAAS Dashboard! This dashboard replaces the previous
```
hatched, line 61: This wording isn't appropriate for all the environments we deploy this to.

hatched:
Looks good, just the wording needs some updates 👍

src/components/Layout/Layout.js
barrymcgee: This is useless feedback, please be more specific - what's not appropriate about it?

hatched: I had commented in my suggestion previously, but the issue is that this dashboard build is deployed in different environments so the messaging needs to be appropriate for those environments. So maybe something like...

```html
Welcome the new JAAS Dashboard! This dashboard is the replacement for the Juju GUI in JAAS and individual Juju Controllers from Juju 2.8.{" "}
<span className="u-hide--small">
  Read more and join the discussion about this new product{" "}
  <a href="#_">on Discourse</a>. We would love to hear your feedback!
</span>
```

PR #458, opened 2020-04-24
Use state to set localStorage for notification

barrymcgee:
## Done

Use state to set localStorage for notification

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Dismiss release banner
- It should immediately disappear
- Refresh
- It should not return

Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/457

### DIFF
--- a/src/hooks/useLocalStorage.js
+++ b/src/hooks/useLocalStorage.js
@@ -1,0 +1,33 @@
+import { useState } from "react";
+
+function useLocalStorage(key, initialValue) {
+  // State to store our value
+  // Pass initial state function to useState so logic is only executed once
+  const [storedValue, setStoredValue] = useState(() => {
+    try {
+      const item = window.localStorage.getItem(key);
+      return item ? JSON.parse(item) : initialValue;
+    } catch (error) {
+      console.error("Unable to parse local storage:", error);
+      return initialValue;
+    }
+  });
+
+  // Return a wrapped version of useState's setter function that persists the
+  // new value to localStorage.
+  const setValue = (value) => {
+    try {
+      // Allow value to be a function so we have same API as useState
+      const valueToStore =
+        value instanceof Function ? value(storedValue) : value;
+      setStoredValue(valueToStore);
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  return [storedValue, setValue];
+}
+
+export default useLocalStorage;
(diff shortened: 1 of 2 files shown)


src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+  // Pass initial state function to useState so logic is only executed once
+  const [storedValue, setStoredValue] = useState(() => {
+    try {
+      // Get from local storage by key
+      const item = window.localStorage.getItem(key);
+      // Parse stored json or if none return initialValue
+      return item ? JSON.parse(item) : initialValue;
+    } catch (error) {
+      // If error also return initialValue
+      console.log(error);
```
hatched, line 14: I'm not really a fan of the silent failure and this will be difficult to grep for.
```suggestion
      console.error('unable to parse local storage:', error);
```

Or have a shared error handler between these two try/catch blocks.

src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+      return initialValue;
+    }
+  });
+
+  // Return a wrapped version of useState's setter function that ...
+  // ... persists the new value to localStorage.
```
hatched, line 20: what's with the ...s?

src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+  const setValue = (value) => {
+    try {
+      // Allow value to be a function so we have same API as useState
+      const valueToStore =
+        value instanceof Function ? value(storedValue) : value;
+      // Save state
```
hatched, line 26: I'm not sure the overuse of comments explaining what each line does is necessary.

src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+      setStoredValue(valueToStore);
+      // Save to local storage
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      // A more advanced implementation would handle the error case
```
hatched, line 31: oh I'm guessing this was copied from somewhere? :D

src/hooks/useLocalStorage.js
```diff
@@ -0,0 +1,39 @@
...
+      console.log(error);
```
hatched, line 32: If you're not going to create a shared error handler then...
```suggestion
      console.error('unable to store value to localstorage:', error);
```

src/hooks/useLocalStorage.js
barrymcgee: https://usehooks.com/useLocalStorage/

PR #471, opened 2020-05-01
Send error to sentry if ErrorBoundary is triggered.

hatched:
## Done

Send error to sentry if ErrorBoundary is triggered.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- comment out the production check https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/index.js#L31
- Add some code to throw an error `throw "foo"` in a react component and see that an error is sent.

### DIFF
--- a/src/components/ErrorBoundary/ErrorBoundary.js
+++ b/src/components/ErrorBoundary/ErrorBoundary.js
@@ -1,4 +1,5 @@
 import React, { Component } from "react";
+import * as Sentry from "@sentry/browser";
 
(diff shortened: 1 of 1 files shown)


barrymcgee:
Sentry reports errors with the event it was passed?

Screenshot 2020-05-05 at 12 21 48

hatched:
I can't seem to get the application to compile to reproduce the error you're seeing, however if an error is thrown it appears to work as expected. See error ID 9720

PR #474, opened 2020-05-13
Generate release build tarball

hatched:
## Done

Generate a tarball that can be used in Juju and JAAS.

## QA

- Pull code
- Run `./run exec yarn run generate-release-tarball`
- Run `tar -ff juju-dashboard-0.1.1.tar.bz2

It should look like the build directory tar'd up.


hatched:
Good call, done

PR #476, opened 2020-05-13
Handle non-admins accessing the Controller view

### LINKED ISSUE #462: Console error across all envs
Chrome:
Screenshot 2020-04-27 at 19 29 46

Firefox:
Screenshot 2020-04-27 at 19 35 27

hatched:
## Done

If a user is not an administrator on the controller then do not try and request the controller data. 
If a non-admin tries to view the controller page then show a warning.

## QA

- Pull code
- Open http://0.0.0.0:8036/

JAAS
- View the controller page, you should see a warning like below as you're connected to public JAAS.

Self-Bootstrap
- `juju bootstrap localhost`.
- Access the existing GUI and accept the self-signed cert `juju gui`.
- Then modify `config.js` to point to your new controller `baseControllerURL: "10.123.321.1:17070",`
- Open the Dashboard and enter the credentials from the above `juju gui` command.
- View the controller page, it should show a controller view without data, but no warning. The data for this view on a self-bootstrap is coming next.

## Details

Fixes #462 

## Screenshots

Screen Shot 2020-05-13 at 4 39 40 PM

### DIFF
--- a/src/pages/Controllers/Controllers.test.js
+++ b/src/pages/Controllers/Controllers.test.js
@@ -58,4 +58,19 @@ describe("Controllers table", () => {
     );
     expect(wrapper.find("tbody tr").get(0)).toMatchSnapshot();
   });
+
+  it("renders a warning if you do not have access", () => {
+    dataDump.root.controllerConnection.info.user.controllerAccess = "";
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <MemoryRouter>
+        <Provider store={store}>
+          <Controllers />
+        </Provider>
+      </MemoryRouter>
+    );
+    expect(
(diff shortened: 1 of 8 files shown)


src/pages/Controllers/Controllers.test.js
```diff
@@ -58,4 +58,19 @@ describe("Controllers table", () => {
     );
     expect(wrapper.find("tbody tr").get(0)).toMatchSnapshot();
   });
+
+  it("renders a warning if you do not have access", () => {
```
anthonydillon, line 62: Can you add a test to check it does display the details if "superuser" or "admin"

hatched: That's the default case, but I can add a comment in the other test cases to make that clear.

PR #487, opened 2020-05-19
Tweak hr styling

### LINKED ISSUE #485: Reduce width of the horizontal lines in the side nav (2.8)
Can we reduce the width of the lines to match the content, as in the [designs on Zeplin](https://zpl.io/2GvKZL3).

If we can also reduce the opacity to 0.5 that would be good.

image

barrymcgee:
## Done

Tweak hr styling

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify tweaked hr rules in sidebar match the nav

## Details

Screenshot 2020-05-19 at 11 18 55

Fixes #485

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.js
+++ b/src/components/PrimaryNav/PrimaryNav.js
@@ -232,6 +232,7 @@ const PrimaryNav = () => {
           </li>
         ))}
       </ul>
+      <hr className="p-primary-nav__divider" />
       <div className="p-primary-nav__bottom">
         <ul className="p-list">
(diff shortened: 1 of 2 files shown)


spencerbygraves:
@anthonydillon I believe we agreed to not update to the new nav until it works properly in an apps layout, and has the collapse and expand on hover for example.

spencerbygraves:
@barrymcgee sorry, looks like it should be 0.1 opacity.

https://vanillaframework.io/docs/examples/patterns/side-navigation/dark

Thanks

PR #494, opened 2020-05-21
fix(model details): remove redundant table header

### LINKED ISSUE #486: Relation provider table isn't full width (2.8)
image

barrymcgee:
@hatched I removed this header as there is no content being generated for the column - correct?

## Done

(fix) Remove redundant table header
(fix) Correct first letter capitalisation util 

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View model details page for model with relations info
- View relations table and verify headers match columns

## Details

Fixes #486

### DIFF
--- a/src/pages/Models/Details/generators.js
+++ b/src/pages/Models/Details/generators.js
@@ -42,7 +42,6 @@ export const relationTableHeaders = [
   { content: "requirer" },
   { content: "interface" },
   { content: "type" },
-  { content: "message" },
 ];
 
 export function generateIconImg(name, namespace) {
--- a/src/scss/_utils.scss
+++ b/src/scss/_utils.scss
@@ -3,8 +3,12 @@
   text-transform: capitalize;
 
   // Capitalise only the first letter of a string
-  &--first-letter::first-letter {
-    text-transform: uppercase;
+  &--first-letter {
+    display: inline-block;
+
+    &::first-letter {
+      text-transform: uppercase;
+    }
   }
 }
 


hatched:
There can be....are you saying you've found a spec that removes this data from Juju?

barrymcgee:
@hatched I'm saying no logic exists to generate this column data so I'm asking if that reason is that it's unavailable to us?

hatched:
Ahh, no this data is potentially available, so if it's not provided lets just default to an empty string instead of removing the column.

src/pages/Models/Details/generators.js
```diff
@@ -42,7 +42,6 @@ export const relationTableHeaders = [
   { content: "requirer" },
   { content: "interface" },
   { content: "type" },
-  { content: "message" },
```
hatched, line 45: Lets just default this to an empty string if we do not have data to render, as the content may be available.

barrymcgee: Given we have limited screen real estate, particularly on smaller screens, I don't think we should be rendering columns of whitespace. When we have the data, then we render the column?

spencerbygraves:
@hatched I think @barrymcgee is right, can we use the space and show the column when we have the data.

@clagom would you agree?

clagom:
I think this can work nicely, something we can iterate on, but generally speaking would help to display the data, yes @spencerbygraves

hatched:
This is an architectural change to how the tables operate. We haven't currently hidden any columns if we do not have content for them. If that's something we want to implement then it's a much bigger task than this. We would probably want some way to fade in/out columns as the data arrives/leaves. This would also reflow the table, potentially while the user is using it, I'm not sure I've ever seen a tabular data representation with dynamic columns.

barrymcgee:
@hatched We appear to be talking at cross purposes - let's discuss in standup.

clagom:
I thought the question was 'let's do it' or 'not doing it' meaning something trivial. But if requires more than this then definitely it's not high priority

barrymcgee:
https://github.com/canonical-web-and-design/jaas-dashboard/issues/504

Issue #435, opened 2020-04-17
No indication things are loads


anthonydillon:
**Describe the bug**
When you first open the app it takes a while to populate the model data. In this time there is no user feedback and the three tables is strange without content. 

**Expected behavior**
I would expect this to vote to work like MAAS and only display a status table once there is content for it. If no models are available the application should say so.

clagom:
We have several design for the loading and empty states, but I remember @hatched wanted to see how the new watcher behaved (correct me if I’m wrong, Jeff). Not sure, but might be worth implementing a loading or empty message state anyway?

hatched:
Yes @clagom I still feel like we shouldn't implement those designs as the new watcher will be coming in the near future. 

This is interesting multi-step problem...

1. Application has loaded and we're connecting to the controller and fetching the initial payload.
2. The payload has been retrieved... 
a ...we have data for all tables.
b ...and we don't have data for all tables.
c ...we have no data for any table.
3. We're requesting updated data. 
4. Content is updating.

Because of this process I feel like we should start without any tables rendered and instead.. 

1. An indicator that we're connecting to the controller and fetching the payload.
2. This indicator goes away...
a ... and the data is rendered into the table.
b ... only tables with data are rendered.
c ... no tables are rendered and instead an onboarding guide on what to do next.
3. With the new watcher we'll never be requesting updating data, changes will be pushed to us.
4. We should highlight in some manner that a row/field was updated.

anthonydillon:
Yeah, this issue is about the initial render of the application and not the update life cycle. I agree 1 and 2 suggestions. 

Not sure about 4. I would say if it's reasonably "live" we just present the data and the user can be confident it's up to date. If something noteworthy occurs like a model modeling to error state we should provide notifications for that.

PR #497, opened 2020-05-21
(spike) Experimental table cell loading

barrymcgee:
## Done

(Spike) Experimental table cell loading using a skeleton loading component in place of each cell

## QA

**Do not QA** - this is monkey code for demo purposes only

Related: #435

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "react": "16.13.1",
     "react-dom": "16.13.1",
     "react-ga": "2.7.0",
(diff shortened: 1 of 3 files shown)


hatched:
Interesting idea, I haven't been a fan of this 'fake' data loading pattern as it's rare that the real data matches what's coming in, however maybe this can be done tastefully...

I would be -1 on this package though, it increases the application side post gzip by 7KB and the build by over 200KB. I figure we could accomplish something identical with 10 or so lines of js and css using our current architecture.

PR #501, opened 2020-05-21
fix(model details): correctly display the modelName in the expanded model topology view

### LINKED ISSUE #500: Topology can erroneously report model not loaded

barrymcgee:
## Done

(fix) Fall back to modelName if cannot split on slash -this prevents an erroneous "Model not loaded" message from appearing

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to model details pages, expand topologies and verify the title of the model appears correctly

## Details

Fixes #500

### DIFF
--- a/src/components/InfoPanel/InfoPanel.js
+++ b/src/components/InfoPanel/InfoPanel.js
@@ -66,7 +66,7 @@ const InfoPanel = () => {
       {showExpandedTopology ? (
         <Modal
           close={() => setShowExpandedTopology(false)}
-          title={modelName.split("/")[1] || "Error: model not loaded..."}
+          title={modelName.split("/")[1] || modelName}
           data-test="topology-modal"
         >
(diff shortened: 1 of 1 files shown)


src/components/InfoPanel/InfoPanel.js
```diff
@@ -66,7 +66,7 @@ const InfoPanel = () => {
       {showExpandedTopology ? (
         <Modal
           close={() => setShowExpandedTopology(false)}
-          title={modelName.split("/")[1] || "Error: model not loaded..."}
+          title={modelName.split("/")[1] || modelName}
```
hatched, line 69: You'll still want to keep the old fallback as a third option as that was added to resolve another issue.

barrymcgee: The original issue was preventing a blank value if the model name could not be split on slash... this change addresses that already.

Issue #491, opened 2020-05-20
Correctly handle timeouts when logging in


hatched:
When logging in via an identity provider it's possible that the user takes too long to successfully log in on the provider and then the request times out. After a successful log in the requests are made, which fail because it's no longer connected.

![Screen Shot 2020-05-20 at 9 38 40 AM](https://user-images.githubusercontent.com/532033/82466545-c9799780-9a7d-11ea-8ea9-418a18aeb7f3.png)

Issue #477, opened 2020-05-13
Persist u/p login for a short time


hatched:
If you log in with a user/pass you're immediately logged out on refresh. We should persist this for a short time like we do with macaroons.

cristinadresch:
@hatched are you planning to complete this one by end of this week?

hatched:
Unlikely, I'm having discussions with the juju team on various approaches to resolve this.

hatched:
When providing the full `user-tag` and `credentials` to the Juju API it does not return a macaroon, this is by design. To get it to return a macaroon you need to provide only the `user-tag`. It will then respond with a macaroon that needs to be discharged at the supplied location. In my quick tests I was not able to get this location to discharge the macaroon. This system is how the CLI performs a login when the user provides only the username and not the password so it is possible...

Juju uses https://github.com/go-macaroon-bakery/macaroon-bakery/tree/v2/httpbakery to discharge the macaroons. Investigation is needed into how this differs from the bakeryjs version that we use.

hatched:
Now that users can register additional controllers with the user/pass authentication system we should investigate using the macaroon token flow for u/p registrations. This would mean that we would only have to store the macaroons and not the user/pass directly.

cristinadresch:
@hatched what should be the next steps here?

huwshimi:
This is the relevant issue blocking us from implementing this: https://bugs.launchpad.net/juju/+bug/1864517.

PR #507, opened 2020-05-25
Render any login error messages.

### LINKED ISSUE #388: Provide login failure feedback
If login fails whether from an external identity provider or u/p there is no indication in the UI of a problem.

hatched:
## Done

If there is an error message when logging in via the user/pass dialog then render that error.

## QA

- Pull code
- Install the juju 2.8-rc2 snap
- `juju bootstrap` into a public cloud
- `juju dashboard` and connect to that dashboard to accept the self-signed cert.
  - take note of the ip address and u/p
- Open config.js and change: 
  - `baseControllerURL` to the ip address and port (if any) from above.
  - `isJuju` to `true`
  - `identityProviderAvailable` to `false`
- Open http://0.0.0.0:8036/
- Submit the form with an invalid username and password.
- Use the above u/p to log in successfully. 

## Details

Fixes #388 

## Screenshots

Screen Shot 2020-05-25 at 9 08 26 AM
Screen Shot 2020-05-25 at 9 08 37 AM

### DIFF
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -13,6 +13,7 @@ export const actionsList = {
   logOut: "LOG_OUT",
   storeBakery: "STORE_BAKERY",
   storeConfig: "STORE_CONFIG",
+  storeLoginError: "STORE_LOGIN_ERROR",
   storeUserPass: "STORE_USER_PASS",
   storeVersion: "STORE_VERSION",
   storeVisitURL: "STORE_VISIT_URL",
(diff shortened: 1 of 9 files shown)


barrymcgee:
The first error message "macaroon discharge failed" is completely meaningless to the user - can we change it to something human readable?

barrymcgee:
401 status is not reflected in the UI - we should indicate an error for anything other than success.

Screenshot 2020-05-26 at 08 29 41

hatched:
@barrymcgee thanks for the review/qa

That 401 you're seeing is because you're using the u/p login while trying to connect to an external identity provider. That issue is going to be handled in #491.

As far as the macaroon message, We can't address every possible error value returned from the various back ends so this was an example of one of those complicated cases. In this case I'm not sure what we could substitute it for. Although with the changes required for #477 that specific message will go away as the password field will not be shown initially.

Issue #526, opened 2020-05-28
cannot view a model consuming a CMR


mitechie:
**Describe the bug**
I have a CMR in a controller and the offering side is ok, but the consuming side won't load and I got a nice helpful link to file a bug here

**To Reproduce**

juju add-model keystone-backend
juju deploy (https://pastebin.canonical.com/p/4qB3D8DZkk/)

juju add-model keystone-frontend
juju deploy (https://pastebin.canonical.com/p/VWWgVG5cy7/)

juju dashboard
click models
...breakage

hatched:
Thanks for filing this issue @mitechie, great repro steps!

The issue is that the topology is trying to draw a relation to an application in the 'consuming' model `keystone-frontend` that doesn't exist.

<img width="1222" alt="Screen Shot 2020-05-28 at 7 47 17 PM" src="https://user-images.githubusercontent.com/532033/83212239-55de2700-a11c-11ea-8281-ca1b7fa7a648.png">

There are a number of tasks to fully support cross model relations but we should be able to at least prevent the application from breaking until the time when those other tasks can be completed.

PR #533, opened 2020-05-29
Render the controller name in the model table

anthonydillon:
## Done
Render the controller name in the model table

## QA
- `juju bootstrap` into a public cloud
- `juju dashboard` and connect to that dashboard to accept the self-signed cert.
  - take note of the ip address and u/p
- Follow the deployment steps in #526 
- Open config.js and change: 
  - `baseControllerURL` to the ip address and port (if any) from above.
  - `isJuju` to `true`
  - `identityProviderAvailable` to `false`
- `./run`
- Open http://0.0.0.0:8036/
- You should see your controller name in the list instead of the uuid

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/532

## Screenshots
image


src/components/ModelTableList/shared.js
```diff
@@ -136,21 +139,3 @@ export function getStatusValue(status, key) {
...
   return returnValue;
 }
-
-/**
-  Renders the controller UUID element. If the UUID is the JAAS controller then
-  it replaces the UUID with "JAAS".
-  @param {string} uuid The controller UUID.
-  @returns {Object} The element to render the controller UUID entry
-*/
-export function generateControllerUUID(uuid) {
```
anthonydillon, line 146: Removed this function as its not used anywhere anymore.

anthonydillon:
@hatched @barrymcgee this is now ready with tests.

src/app/selectors.js
```diff
@@ -367,9 +370,35 @@ const countModelStatusGroups = (groupedModelStatuses) => {
...
-const filterModelData = (filters, modelData) => {
+const filterModelData = (filters, modelData, controllerData) => {
+  const clonedModelData = cloneDeep(modelData);
+  // Add controllerName to the model data by mapping the controllerData
+  Object.entries(clonedModelData).forEach((model) => {
+    if (model[1].info) {
+      let controllerName = null;
+      if (controllerData) {
+        for (var i = 0; i < controllerData.length; i++) {
+          if (controllerData[i].uuid === model[1].info.controllerUuid) {
```
hatched, line 381: a bit pedantic but can you invert this....you're actually checking if the controller uuid is the same one in the model....so the model goes first 😆

src/app/selectors.js
```diff
@@ -367,9 +370,35 @@ const countModelStatusGroups = (groupedModelStatuses) => {
   @param {Object} modelData The model data from the redux store.
   @returns {Object} The filtered model data.
 */
-const filterModelData = (filters, modelData) => {
+const filterModelData = (filters, modelData, controllerData) => {
+  const clonedModelData = cloneDeep(modelData);
+  // Add controllerName to the model data by mapping the controllerData
+  Object.entries(clonedModelData).forEach((model) => {
+    if (model[1].info) {
+      let controllerName = null;
```
hatched, line 378: If you set the default value to the UUID then you can remove the `else` part of the conditional on 392.
```suggestion
      let controllerName = model[1].info.controllerUuid;
```

src/components/ModelTableList/CloudGroup.js
```diff
@@ -67,48 +63,57 @@ export default function CloudGroup({ activeUser, filters }) {
       modelGroup.forEach((model) => {
         const { highestStatus } = getModelStatusGroupData(model);
         cloudModels.rows.push({
+          "data-test-model-id": model?.uuid,
```
hatched, line 66: Maybe we should be more specific in our `data-test` attributes, so 'id' to 'uuid' here? If you agree the other files will need to be updated as well.

```suggestion
          "data-test-model-uuid": model?.uuid,
```

src/components/ModelTableList/CloudGroup.js
```diff
@@ -67,48 +63,57 @@ export default function CloudGroup({ activeUser, filters }) {
...
           columns: [
             {
+              "data-test-column": "name",
```
hatched, line 69: What do we think of using the prefix `data-test` for test-only attributes? Typically the attribute is simply `data-test` and then the value has the additional context to it.

I think I'm ok with this as it's still understandable and grep-able.

@barrymcgee

src/components/ModelTableList/CloudGroup.js
```diff
@@ -67,48 +63,57 @@ export default function CloudGroup({ activeUser, filters }) {
...
               ),
             },
-            // We're not currently able to get the controller name from the API
-            // so, display the controller UUID instead.
             {
-              content: generateControllerUUID(
-                getStatusValue(model.info, "controllerUuid")
+              "data-test-column": "controller",
+              content: (
+                <span data-test="controller-name">
```
hatched, line 109: Is it necessary to wrap this in a span and add another data attribute?

src/components/ModelTableList/ModelTableList.test.js
```diff
@@ -72,4 +73,57 @@ describe("ModelTableList", () => {
       expect(wrapper.find(table.component).prop("filters")).toBe(filters);
     });
   });
+
+  it("renders the controller name as JAAS", () => {
```
hatched, line 77: Great tests! 🎉

src/components/ModelTableList/CloudGroup.js
anthonydillon: I must prefer having bespoke attributes as it causes a separation of concerns between functionality and testing. It makes it clear from the code what the purpose of the attribute.

It's absolutely normal to have multiple word data attributes: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

src/app/selectors.js
```diff
@@ -367,9 +370,35 @@ const countModelStatusGroups = (groupedModelStatuses) => {
...
+      if (controllerData) {
+        for (var i = 0; i < controllerData.length; i++) {
```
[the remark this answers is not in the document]
anthonydillon: I know this is the new coolness but there is something nice about an old fashioned for loop with a `break;` :stuck_out_tongue:

src/components/ModelTableList/CloudGroup.js
anthonydillon: Good idea.

src/app/selectors.js
```diff
@@ -367,9 +370,35 @@ const countModelStatusGroups = (groupedModelStatuses) => {
   @param {Object} modelData The model data from the redux store.
   @returns {Object} The filtered model data.
 */
-const filterModelData = (filters, modelData) => {
+const filterModelData = (filters, modelData, controllerData) => {
+  const clonedModelData = cloneDeep(modelData);
+  // Add controllerName to the model data by mapping the controllerData
```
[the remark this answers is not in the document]
anthonydillon: Done

src/app/selectors.js
anthonydillon: Done

src/app/selectors.js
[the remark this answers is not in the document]
anthonydillon: Done

src/app/selectors.js
anthonydillon: Done

src/components/ModelTableList/CloudGroup.js
anthonydillon: Not at all. Leftover from before I know arbitrary attributes can do add.

src/components/ModelTableList/CloudGroup.js
hatched: Sounds all good to me!

hatched:
QA Note:

Looks like the column is empty on JAAS
Screen Shot 2020-06-02 at 3 25 37 PM

hatched:
But it works ok on a self-bootstrapped controller.

hatched:
Thanks for the updates, looks good both on JAAS and Juju

PR #540, opened 2020-06-01
Use agent status instead of instance status in machine table

### LINKED ISSUE #305: Stopped machines reporting as running
The latter "MESSAGE" is reporting correctly from Juju but the `status` should be "stopped" not "Running"

To reproduce, deploy an application and then once you have a machine run `juju remove-machine #` and wait. 

Screen Shot 2020-01-23 at 2 12 59 PM
Screen Shot 2020-01-23 at 2 14 06 PM

Screen Shot 2020-06-01 at 1 07 25 PM

hatched:
## Done

The machine table now uses the agent status instead of instance status to make it more Juju-centric and provide clearer error messaging. The upcoming machine view will have the instance status data to provide a full picture.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View the model details of a few models in error, alert, running to see the different values for this table.
- Add a unit to an application and watch the statuses update(depending on the provider and update cycles you may miss some if they go fast).

## Details

Fixes #305

## Screenshots

Screen Shot 2020-06-01 at 1 25 23 PM
Screen Shot 2020-06-01 at 1 27 18 PM

### DIFF
--- a/src/pages/Models/Details/generators.js
+++ b/src/pages/Models/Details/generators.js
@@ -259,15 +259,15 @@ export function generateMachineRows(modelStatusData, filterByApp) {
           ),
         },
         {
-          content: generateStatusElement(machine.instanceStatus.status),
+          content: generateStatusElement(machine.agentStatus.status),
           className: "u-capitalise",
         },
         { content: splitParts(machine.hardware)["availability-zone"] },
         { content: machine.instanceId },
         {
           content: (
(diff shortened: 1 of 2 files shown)


anthonydillon:
Once mine is merged you could write a test?

hatched:
@anthonydillon yeah totally, we don't have any tests for this stuff at all, so I think we'll have to go through the app one week and just write tests :)

PR #542, opened 2020-06-02
WIP: Reposition topology icons

barrymcgee:
@hatched @anthonydillon Pushing up this WIP draft PR so you can have visibility on this feature branch as it's long-running and will remain so for another while.

I'm making good progress but it's a little intricate.

Completed:

- [x] Prevent interaction on apps unless in modal
- [x] Highlight the active app when dragging apps
- [x] Move relation lines when dragging apps
- [x] Refactor positioning logic to place SVG relative to the container

Still to-do:

- [ ] Align SVG when not in modal and simplify code
- [ ] Restrict interactions based on permissions
- [ ] Restrict dragging out of bounds
- [ ] Save position and push back to Juju annotations on drag end

@hatched The constant re-rendering issue (https://github.com/canonical-web-and-design/jaas-dashboard/issues/430) is going to block me very soon... can you get any time soon to look at it?


src/components/Topology/Topology.js
```diff
@@ -1,7 +1,12 @@
-import React, { useRef, useEffect } from "react";
+import React, { useRef, useEffect, useState, memo } from "react";
+import { useSelector } from "react-redux";
 import * as d3 from "d3";
+// import { updateAnnotations } from "juju/index";
```
hatched, line 4: commented out code

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+        style={{ backgroundImage: `url(${fullScreenIcon})` }}
+        className="p-icon--expand p-icon--fullscreen"
+        data-test="icon--fullscreen"
+        onClick={() => setIsFullscreen(true)}
+      />
+    );
+  }
 };
+
+function freezeRendering(prevProps) {
```
hatched, line 470: Nice trick :)

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+        setIsFullscreen(false);
+      }
+    };
+    document.addEventListener("keydown", closeOnEscape);
+    return () => {
+      document.removeEventListener("keydown", closeOnEscape);
+    };
+  });
+
+  console.log("texts");
```
hatched, line 413: log

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+            />
+          )}
+          {isReadOnly && isFullscreen && (
+            <span className="read-only">Read only</span>
+          )}
+        </div>
+      </div>
+    );
+  } else {
+    return (
```
hatched, line 459: This doesn't look great, Could we have just an empty grey box rendered while it's waiting for data?

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+      }
+    };
+    document.addEventListener("keydown", closeOnEscape);
+    return () => {
+      document.removeEventListener("keydown", closeOnEscape);
+    };
+  });
+
+  console.log("texts");
+  if (Object.entries(modelData?.applications || {}).length) {
```
hatched, line 414: I think I'd prefer to just render an empty topology vs special casing when `modelData` or there are no applications. Consider the use case when you have a new model with no applications yet. It looks better to have an empty toplogy window than just the 'expand' icon in the top left.

Screen Shot 2020-06-15 at 11 16 41 AM

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+    return (
+      <div
+        className="topology"
+        data-fullscreen={isFullscreen}
+        data-readonly={isReadOnly}
+        ref={topologyRef}
+      >
+        <div className="topology__inner">
+          {/* Remove flash messaging until we can save positions back to juju */}
```
hatched, line 423: Lets remove this until we can implement it as this might not be the place we ultimately decide to put it.

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+            {flashMessage?.message}
+          </div> */}
+          {isFullscreen && (
+            <i
+              className="p-icon--close"
+              onClick={() => setIsFullscreen(false)}
+            />
+          )}
+
+          {modelData ? (
```
hatched, line 434: This doesn't work as expected because of the conditional on ln 414

src/components/Topology/Topology.js
```diff
@@ -205,24 +282,27 @@ export default ({ modelData, width, height }) => {
...
+          const translateX = (width - svgWidth * scale) / 2 || 0;
+          const translateY = (height - svgHeight * scale) / 2 || 0;
+          topo
+            .attr(
+              "transform",
+              `translate(${translateX},${translateY}) scale(${scale},${scale})`
+            )
+            .attr("width", `${width}`)
+            .call(
+              zoom.transform,
```
hatched, line 303: Can you create an issues to track the... 
 - 'zoom in on click' issue when the topology is in fullscreen mode.
 - the perpetually smaller zoom level on the small topology after interacting with the full screen mode.

These also only seem to happen when applications already have existing positional annotations. For applications that are being auto-placed these issues don't appear to exist.

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
     relationLines.exit().remove();
 
     return () => {
       topo.remove();
     };
-  }, [applications, deltaX, deltaY, height, width, maxX, maxY, relations]);
-  return <svg ref={ref} />;
+  }, [
+    applications,
+    conn,
```
hatched, line 377: You shouldn't rely on conn, it has a timestamp in it so it'll change every time a request is ever transferred over it.

src/components/Topology/Topology.js
```diff
@@ -240,25 +320,158 @@ export default ({ modelData, width, height }) => {
...
+    deltaX,
+    deltaY,
+    isFullscreen,
+    isReadOnly,
+    maxX,
+    maxY,
+    relations,
+  ]);
+
+  // XXX Remove flash messaging until we can save results back to juju
```
hatched, line 387: commented out code.

src/components/Topology/Topology.js
```diff
@@ -105,21 +116,41 @@ const getRelationPosition = (data) => {
...
 
-  const applications =
-    (modelData &&
-      Object.keys(modelData.applications).map((appName) => ({
-        ...modelData.annotations[appName],
-        ...modelData.applications[appName],
-        name: appName,
-      }))) ||
-    [];
+  // If active user and model owner are oen and the same, grant write access
```
hatched, line 136:
```suggestion
  // If active user and model owner are one and the same, grant write access
```

src/components/Topology/Topology.js
```diff
@@ -205,24 +282,27 @@ export default ({ modelData, width, height }) => {
...
-            width / (svgWidth + padding),
-            height / (svgHeight + padding)
-          );
-          const translateX = (width - svgWidth * scale) / 2;
-          const translateY = (height - svgHeight * scale) / 2;
-          topo.attr(
-            "transform",
-            `translate(${translateX},${translateY}) scale(${scale},${scale})`
-          );
+          const scale = Math.min(width / svgWidth, height / svgHeight) || 1;
```
hatched, line 292: I'm not entirely sure where the problem is happening but it doesn't appear to be able to handle negative values properly. Note how in the following screenshot the icons are spilling out of the Topology bounadry. This works properly in the production release.

Screen Shot 2020-06-15 at 11 33 11 AM

src/components/Topology/Topology.js
barrymcgee: Then what do we do for models without any applications?

src/components/Topology/Topology.js
barrymcgee: https://github.com/canonical-web-and-design/jaas-dashboard/issues/559

src/components/InfoPanel/InfoPanel.js
```diff
@@ -47,61 +23,12 @@ const InfoPanel = () => {
...
-                });
-              }}
-            >
-              Expand topology
-            </i>
-          )}
-        </div>
-      )}
+      <div className="info-panel__pictogram">
+        <Topology modelData={modelStatusData} data-test="topology" />
```
hatched, line 29: I'm not sure you're using the `data-test` in the tests here.

src/pages/Models/Details/ModelDetails.test.js
```diff
@@ -27,7 +27,6 @@ describe("ModelDetail Container", () => {
         </MemoryRouter>
       </Provider>
     );
-    expect(wrapper.find("Topology").length).toBe(1);
```
hatched, line 30: This should have still been an ok assertion to have here....did it fail?

barrymcgee:
Squashing and merging into feature branch for continued development.

PR #544, opened 2020-06-03
Set a `unitCount` value per application when filtering in model details

anthonydillon:
## Done
- Set a `unitCount` value per application when filtering to render the correct scaling value even if filtered out
- Added data attributes to table cells for testing

## QA
- Pull code
- Open http://0.0.0.0:8036/
- Go into a model details page and check the scale values on in the application table are correct
- Click on an application to filter it
- See that the scaling counts on all application remains correct

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/543

## Screenshots

| Not filtered  | Filter application with units | Filter application without units
| ------------- | ------------- | ------------- |
| 
Screenshot_2020-06-03 JAAS Dashboard
  | 
Screenshot_2020-06-03 JAAS Dashboard(1)
  | 
Screenshot_2020-06-03 JAAS Dashboard(2)
  |

### DIFF
--- a/src/pages/Models/Details/ModelDetails.js
+++ b/src/pages/Models/Details/ModelDetails.js
@@ -32,59 +32,69 @@ import "./_model-details.scss";
   @param {String} appName The name of the application to filter the data by.
 */
 const filterModelStatusData = (modelStatusData, appName) => {
-  if (!modelStatusData || appName === "") {
-    return modelStatusData;
-  }
-  const filteredData = cloneDeep(modelStatusData);
-  const application = filteredData.applications[appName];
-  // remove the units from the application objects that are not
-  // the filter-by app.
-  const subordinateTo = application.subordinateTo || [];
-  Object.keys(filteredData.applications).forEach((key) => {
-    if (key !== appName && !subordinateTo.includes(key)) {
-      filteredData.applications[key].units = {};
+  if (modelStatusData) {
+    const filteredData = cloneDeep(modelStatusData);
+    Object.keys(filteredData.applications).forEach((key) => {
+      filteredData.applications[key].unitsCount = Object.keys(
+        filteredData.applications[key].units
+      ).length;
+    });
+    if (appName === "") {
+      return filteredData;
     }
-  });
-  // Loop through all of the units of the remaining applications that are
-  // listed in the subordinateTo list. This is done because although
-  // a subordinate is supposed to be installed on each unit, that's not
-  // always the case.
-  subordinateTo.forEach((parentName) => {
-    const units = filteredData.applications[parentName].units;
-    Object.entries(units).forEach((entry) => {
-      const found = Object.entries(entry[1].subordinates).find(
-        (ele) => ele[0].split("/")[0] === appName
-      );
-      if (!found) {
-        delete units[entry[0]];
+
+    const application = filteredData.applications[appName];
+    // remove the units from the application objects that are not
+    // the filter-by app.
+    const subordinateTo = application.subordinateTo || [];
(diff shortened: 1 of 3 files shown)


src/pages/Models/Details/ModelDetails.js
```diff
@@ -32,59 +32,69 @@ import "./_model-details.scss";
...
-    (relation) => relation.key.indexOf(appName) > -1
-  );
+    // Remove all relations that don't involve the selected application.
+    filteredData.relations = modelStatusData.relations.filter(
+      (relation) => relation.key.indexOf(appName) > -1
+    );
 
-  return filteredData;
+    return filteredData;
+  } else {
```
hatched, line 95: This else isn't necessary because you're either returning at the end of the conditional above or returning `modelStatusData` there is no path in which you would return `undefined` at the end of the function. So you can just explicitly return `modelStatusData` at the end of the function.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -32,59 +32,69 @@ import "./_model-details.scss";
...
-  const application = filteredData.applications[appName];
-  // remove the units from the application objects that are not
-  // the filter-by app.
-  const subordinateTo = application.subordinateTo || [];
-  Object.keys(filteredData.applications).forEach((key) => {
-    if (key !== appName && !subordinateTo.includes(key)) {
-      filteredData.applications[key].units = {};
+  if (modelStatusData) {
+    const filteredData = cloneDeep(modelStatusData);
+    Object.keys(filteredData.applications).forEach((key) => {
```
hatched, line 37: I feel like this logic belongs in the selector as anywhere that requests this data may find this number useful. If you make the change there then you don't have to modify this file at all.

src/pages/Models/Details/generators.js
```diff
@@ -109,25 +109,36 @@ export function generateApplicationRows(
...
         },
-        { content: "CharmHub" },
         {
+          "data-test-column": "scale",
+          content: app.unitsCount,
+          className: "u-align--right",
+        },
+        {
+          "data-test-column": "store",
+          content: "CharmHub",
```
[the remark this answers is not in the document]
anthonydillon: Yeah left it for another day. Happy to create issues for this. Same for note :shrug:

src/pages/Models/Details/ModelDetails.js
anthonydillon: Good point, done.

src/pages/Models/Details/ModelDetails.js
anthonydillon: Won't it be better to wait until something else needs it before escalating to a selector?

anthonydillon:
@hatched I have addressed your points. Can you assign and change labels to reflect the state of the review please

src/pages/Models/Details/generators.js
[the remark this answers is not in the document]
hatched: No problem, I created it https://github.com/canonical-web-and-design/jaas-dashboard/issues/545

src/pages/Models/Details/ModelDetails.js
hatched: I agree that in this type of situation It's usually a good idea to wait for another use case but then you run into the problem when someone needs the unit count, will they know that this already exists and need to go update it?

In this specific situation however I know that we'll need the unit count with the new 'details' pages that are coming so we might as well head that off right now.

hatched: We've discussed this and decided to leave it here for now.

PR #569, opened 2020-06-18
Added 'Writing React components' and 'Writing tests' content to the README.

hatched:
This is meant to be a reasonable first start to these docs - I imagine that we'll expand and modify these over time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@ The Juju Dashboard uses [Vanilla CSS](https://vanillaframework.io/) and [Vanilla
 #### React component conventions
 
 - Components are stored in their own self-named folders using the TitleCase format.
-- Tests and css files are stored along side each component with the naming convention of ComponentName.css and ComponentName.test.js respectively.
-- Sub components are nested within subfolders their parent components if not shared.
+- Tests and Sass files are stored along side each component with the naming convention of `ComponentName.css` and `ComponentName.test.js` respectively.
+- Sub components are nested within subfolders of their parent components if they are not shared among other components.
 - Use functional components.
 - Use hooks.
 - Consider contributing generic components back to [Vanilla CSS](https://vanillaframework.io/) and [Vanilla React Components](https://github.com/canonical-web-and-design/react-components) where appropriate.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
+      <StatusGroup />
+    </Provider>
+  </MemoryRouter>
+);
+```
+
+If you need to modify the data used in the application, clone the data dump in each test so that you do not run the risk of overriding data in subsequent tests. You should only update the values directly.
+
+```
+const clonedData = cloneDeep(dataDump);
```
Caleb-Ellis, line 162: Just curious - does this add a significant amount of time to running tests?

README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
   - `ssh -fN -L *:17070:0:17070 ubuntu@10.223.241.32` Where the "10." ip is replaced with the ip from the output of the `juju dashboard` command.
 
+### Writing React components
+
+The Juju Dashboard uses [Vanilla CSS](https://vanillaframework.io/) and [Vanilla React Components](https://github.com/canonical-web-and-design/react-components). Be sure to check there first for any components before writing your own.
+
+#### React component conventions
+
+- Components are stored in their own self-named folders using the TitleCase format.
+- Tests and css files are stored along side each component with the naming convention of ComponentName.css and ComponentName.test.js respectively.
```
barrymcgee, line 88:
```suggestion
- Tests and Sass files are stored along side each component with the naming convention of `ComponentName.test.js` and `ComponentName.scss` respectively.
```

README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
+- Sub components are nested within subfolders their parent components if not shared.
```
barrymcgee, line 89:
```suggestion
- Sub components are nested within subfolders of their parent components if they are not shared among other components.
```

README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
+
+#### Mocking data
+
+The file [complete-redux-store-dump.js](src/testing/complete-redux-store-dump.js) contains a sanitized redux dump from a Dashboard instance with many models in different states. When writing a test use the real data provided by this file when generating a redux store that will provide the data to your component or library. Because every test uses the same data set this allows us to see when a library or component will fail if the data that's saved to the redux store, or a selector is modified.
+
+> **If you modify the data that is saved into the redux store then you must update this file**
+
+If you want to test states that are not included in the data dump, or if you want to see how a component will act in different data states you can modify it locally within the test. If you must modify some value see if there is value in adding another entry into the data set for others to use.
+
+#### Test Snippits
```
barrymcgee, line 121:
```suggestion
#### Test Snippets
```

README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
+- Do not mock components unless absolutely necessary. The data dump mentioned below should contain sufficient data to render the full component tree in the test.
+
+#### Mocking data
+
+The file [complete-redux-store-dump.js](src/testing/complete-redux-store-dump.js) contains a sanitized redux dump from a Dashboard instance with many models in different states. When writing a test use the real data provided by this file when generating a redux store that will provide the data to your component or library. Because every test uses the same data set this allows us to see when a library or component will fail if the data that's saved to the redux store, or a selector is modified.
+
+> **If you modify the data that is saved into the redux store then you must update this file**
+
+If you want to test states that are not included in the data dump, or if you want to see how a component will act in different data states you can modify it locally within the test. If you must modify some value see if there is value in adding another entry into the data set for others to use.
```
barrymcgee, line 119:
```suggestion
If you want to test states that are not included in the data dump, or if you want to see how a component will act in different data states you can modify it locally within the test. If you must modify some value, see if there is value in adding another entry into the dataset for others to use.
```

README.md
```diff
@@ -78,9 +78,112 @@ If you're bootstrapping a Juju controller within a `multipass` VM you will not h
...
+
+#### Test Snippits
+
+The typical imports and test case setup looks like the following. Note that we create a shared mockStore factory at the top of each test suite.
```
barrymcgee, line 123:
```suggestion
The typical import and test case setup looks like the following. Note that we create a shared `mockStore` factory at the top of each test suite.
```

README.md
hatched: We haven't had any performance issues with our tests so tbh I didn't test it but I took this as an opportunity to run some micro-benchmarks and the averages are:

19ms to clone the dataDump which is 11,500 lines of heavily nested objects.
95ms to mount the `ModelTableList` component to the DOM in `jsdom`
18ms to run `wrapper.find` an element in the DOM.

So it looks like it doesn't really have any appreciable performance hit in the tests but we also discourage this type of modification where possible so it's only used 4 times throughout the whole test suite.

PR #576, opened 2020-06-25
Remove the 'configuration' header in the model list tables and move the column next to the model name.

hatched:
## Done

Remove the 'configuration' header in the model list tables and move the column next to the model name.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- See the new layout in the model list view in all 3 groupings (status, owner, cloud)

Screen Shot 2020-06-25 at 1 50 16 PM
Screen Shot 2020-06-25 at 1 51 13 PM
Screen Shot 2020-06-25 at 1 52 49 PM

### DIFF
--- a/src/components/ModelTableList/CloudGroup.js
+++ b/src/components/ModelTableList/CloudGroup.js
@@ -35,9 +35,9 @@ function generateCloudTableHeaders(cloud, count) {
       content: generateStatusElement(cloud, count, false),
       sortKey: cloud.toLowerCase(),
     },
+    { content: "", sortKey: "summary" }, // The unit/machines/apps counts
     { content: "Owner", sortKey: "owner" },
     { content: "Status", sortKey: "status" },
-    { content: "Configuration", sortKey: "summary" },
     { content: "Region", sortKey: "region" },
     { content: "Credential", sortKey: "credential" },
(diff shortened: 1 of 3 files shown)


spencerbygraves:
@hatched shouldn't this include the [latest right alignment for app/units/machines](https://zpl.io/a7AZxxR)?

Also, we don't have the correct alignment in the first column. Ideally the status column would also be like this, as shown in the [designs for the modular table](https://zpl.io/a3oQAPA).

image

hatched:
@spencerbygraves nah that can be another issue, this one was just to move the column and remove the header label.

spencerbygraves:
> this one was just to move the column

Can that column align at least?

image

clagom:
> > this one was just to move the column
> 
> Can that column align at least?
> 
> image

I got the same impression, but basically these are 3 different screenshots of the tables grouped differently @spencerbygraves

spencerbygraves:
Yes, but it jumps around. When I switch between the 3 group by tabs, the second column content is in a different position each time.

image

hatched:
So it turns out that that alignment issue is because the cloud table has an extra column and because the columns are dyanmic width it uses up some of that white space.

PR #579, opened 2020-06-29
Fix lint config for installed dependencies.

hatched:
## Done

We no longer use the `airbnb` rules as we use prettier.
There were also a few other plugins that were installed but never ran. This has enabled those and fixed the errors.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The app should work as expected.

### DIFF
--- a/src/app/root.js
+++ b/src/app/root.js
@@ -1,9 +1,9 @@
-import produce from "immer";
+import immerProduce from "immer";
 
 import { actionsList } from "./actions";
 
 function rootReducer(state = {}, action) {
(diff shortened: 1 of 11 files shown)


src/app/root.js
```diff
@@ -1,9 +1,9 @@
-import produce from "immer";
+import immerProduce from "immer";
```
hatched, line 1: This change was because `immer` also has a named export `produce` which throws an issue with the `import` eslint config package.

barrymcgee:
Nice.

PR #597, opened 2020-07-17
Convert the error messages on the model list page into links.

### LINKED ISSUE #595: Make error links link to respective model details page
Screen Shot 2020-07-16 at 11 55 15 AM

hatched:
## Done

Convert the error messages on the model list page into links to the respective model details page.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Are the error links in the model details page links to the model details page?

## Details

Fixes #595

### DIFF
--- a/src/components/ModelTableList/StatusGroup.js
+++ b/src/components/ModelTableList/StatusGroup.js
@@ -48,7 +48,7 @@ const generateWarningMessage = (model, activeUser) => {
   const title = messages.join("; ");
   const link = generateModelDetailsLink(
     model.model.name,
-    model.info && model.info.ownerTag,
+    model?.info?.ownerTag,
     activeUser,
     title
   );
--- a/src/components/ModelTableList/StatusGroup.js
+++ b/src/components/ModelTableList/StatusGroup.js
@@ -40,14 +40,21 @@ function generateStatusTableHeaders(label, count) {
 /**
   Generates the warning message for the model name cell.
   @param {Object} model The full model data.
+  @param {String} activeUser The active user name.
   @return {Object} The react component for the warning message.
 */
-const generateWarningMessage = (model) => {
+const generateWarningMessage = (model, activeUser) => {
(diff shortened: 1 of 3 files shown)


src/components/ModelTableList/StatusGroup.js
```diff
@@ -40,14 +40,21 @@ function generateStatusTableHeaders(label, count) {
...
+  @param {String} activeUser The active user name.
   @return {Object} The react component for the warning message.
 */
-const generateWarningMessage = (model) => {
+const generateWarningMessage = (model, activeUser) => {
   const { messages } = getModelStatusGroupData(model);
   const title = messages.join("; ");
+  const link = generateModelDetailsLink(
+    model.model.name,
+    model.info && model.info.ownerTag,
```
barrymcgee, line 51: Optional chaining?

src/components/ModelTableList/_model-table-list.scss
```diff
@@ -4,7 +4,8 @@
   margin-top: 0.5rem;
 }
 
-.model-table-list_error-message {
+.model-table-list_error-message a,
+.model-table-list_error-message a:visited {
```
barrymcgee, line 8: You should use the parent (&) selector here to reprevent duplication;

```css
.model-table-list_error-message {
  &, &:hover { .. }
}
```

barrymcgee:
From an accessibility point of view, rather than create two adjacent links to the same location, you wrap them in a single link.. other wise screenreaders and keyboard users have to double tab.

src/components/ModelTableList/_model-table-list.scss
hatched: done

hatched:
The double tab concern is true, however the end goal for this is to link to the actual units/machines in error so they will ultimately end up being different links. That work will need to be completed once we finish the new unit/machine views however so this is somewhat of a stepping stone to that.

PR #623, opened 2020-07-23
Register controller UX tweaks

### LINKED ISSUE #601: Only require users to provide a controller IP
Can we add the `wss://` and `/api` parts automatically when registering a new controller? It's more natural for a user to just use the IP from the `juju dashboard` command with is just the IP.

For referrence:
```
You'll typically want to use the public IP address for the controller.
ex) wss://123.456.789.0:17070/api
```

### LINKED ISSUE #611: Gracefully handle invalid controller data in localStorage
In the event that the data in localStorage becomes invalid and is unable to be parsed we need to provide a mechanism to recover. This could mean providing a way to clear it, providing instructions on how to modify it, or providing a UI to the user to recover. We don't want to simply clear it at the point it becomes invalid as we risk blowing away any good data that may be stored there.

This will become less of an issue with proper form validation from #608

### LINKED ISSUE #613: Register controller user/pass tick box should be dynamic
@hatched It appears from this message that if both user and pass input fields have a value, this checkbox should disappear?

Screenshot 2020-07-23 at 08 52 51

### LINKED ISSUE #617: "Visit the controller" is a dead link
Screenshot 2020-07-23 at 09 08 31

### LINKED ISSUE #618: Required fields should be marked as such
The SSL cert checkbox is required to be ticked before the main action here "Add controller" can be activated but it is not visually different to the checkbox above it which is optional.

Screenshot 2020-07-23 at 09 20 00

hatched:
## Done

- Required fields are now marked as such.
- The 'Identity Provider' checkbox how hides if you provide a U/P.
- The user no longer needs to provide the protocol and path for the controller; just the IP.
- The 'Visit Controller' link now only becomes a link when you provide an IP.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The bugs listed below should be resolved

## Details

Fixes #601 
Fixes #611 
Fixes #613 
Fixes #617 
Fixes #618 

## Screenshots

Screen Shot 2020-07-23 at 10 12 58 AM

### DIFF
--- a/src/app/model-poller.js
+++ b/src/app/model-poller.js
@@ -40,15 +40,21 @@ export default async function connectAndListModels(
       controllerList = controllerList.concat(additionalControllers);
     }
     controllerList.forEach(async (controllerData) => {
-      const { conn, error, juju, intervalId } = await loginWithBakery(
-        ...controllerData
-      );
-
-      if (error) {
-        reduxStore.dispatch(storeLoginError(error));
-        return;
+      let conn, error, juju, intervalId;
+      try {
+        ({ conn, error, juju, intervalId } = await loginWithBakery(
+          ...controllerData
+        ));
+        if (error) {
+          reduxStore.dispatch(storeLoginError(error));
+          return;
+        }
+      } catch (e) {
+        return console.log("unable to log into controller", e, controllerData);
       }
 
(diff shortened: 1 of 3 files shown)


hatched:
Thanks @barrymcgee !

Issue #610, opened 2020-07-22
Ask users if they would like to clear saved controllers on log-out


hatched:
Because the controller log in credentials are stored locally in the browser we should ask if the user would like to persist that data when logging out.

hatched:
We could persist the controller data and only clear the password or both the user/pass to make it less inconvenient when logging back in. 

Related #477

anthonydillon:
I would prefer to leave it, additional controllers, in local storage as it would be annoying to set everything up and logout and lose it all. We could add remove functionality to the additional controller list for management of them.

PR #636, opened 2020-07-31
Correctly terminate the controller connections and allow login again.

### LINKED ISSUE #289: Unable to log in after logging out without refresh
Doesn't update the page

### LINKED ISSUE #635: Logging out doesn't gracefully log out of all controllers.
Logging out clears all the user data on the machine and closes the connection effectively logging the user out
but it doesn't explicitly log out of all Juju controllers before doing so causing abruptly closed connections.

hatched:
## Done

- After the multi-controller update logouts were clearing the user data and then closing the connection abruptly. This allows graceful logouts across all registered controllers.
- When using an external identity provider, if you logged out you were not able to log back in without a refresh. This solves that issue by re-connecting to the controller and getting another endpoint to poll on for the macaroon data. 
  - This isn't a reliable long term solution to this problem as eventually the wait url times out and you have to refresh anyways. 
  - A complete solution would involve a multi-step login procedure where the user selects the controllers (with external identity providers) they would like to log into, we then connect to those at that time and present the user with the necessary links to visit to complete their authentication. 

#### Setup
- Bootstrap a controller to use to register an additional controller.
#### JAAS
- Visit the demo link and log in. 
- Register the above controller and it should all be working as expected.
- Log out. It should log you out of both controllers and the user data should be cleared in localStorage.
- Log back in. It should connect to the JAAS controller but not the registered one as it's been cleared (related #610)
- Note the issue outlined above about the timeout if you're not able to log back in.
#### Juju
- Pull down the branch and run it locally.
- Point it at your bootstrapped controller above using the instructions found [here](https://github.com/canonical-web-and-design/jaas-dashboard#developing-while-connected-to-a-juju-controller)
- You should be able to log in using the user/pass.
- Log out
- Log back in, this should work without issue.

## Details

Fixes #289
Fixes #635


src/app/selectors.js
```diff
@@ -67,24 +67,14 @@ export const getLoginError = (state) => state?.root?.loginError;
   @param {Object} state The application state.
   @returns {Object|Null} The juju api instance or null if none found.
```
barrymcgee, line 68: Should this comment be updated to reflect that this now returns more than one instance?

cristinadresch:
@barrymcgee to speak to ant about this

barrymcgee:
- [ ] Clicking the demo link, navigate to /controllers and hard refresh page results in 404
- [ ] Trying to add a controller on demo link, the cert link does not load
- [ ] Try to add a controller on demo fails without any user notification https://share.getcloudapp.com/Jru6wog1

hatched:
> - [ ] Clicking the demo link, navigate to /controllers and hard refresh page results in 404

This is a problem with the docker image that's used for the demo and not this branch.

> - [ ] Trying to add a controller on demo link, the cert link does not load
> - [ ] Try to add a controller on demo fails without any user notification https://share.getcloudapp.com/Jru6wog1

You're trying to add a local controller which the app doesn't have access too so I wouldn't expect this to work. There is a separate issue to provide feedback on failure.

To QA this you'll need to bootstrap a controller in a public cloud or you'll need to set up the appropriate forwarding for your local network for it to access the controller in your VM.

hatched:
Thanks all!

PR #652, opened 2020-08-18
Make the slide panel slide

barrymcgee:
## Done

Make the slide panel slide

## QA

- Pull code
- Open http://0.0.0.0:8036/controllers
- Click 'Add a controller'
- The side panel should slide in and then slide out when closed

### DIFF
--- a/src/components/SlidePanel/_slide-panel.scss
+++ b/src/components/SlidePanel/_slide-panel.scss
@@ -1,6 +1,7 @@
 @import "vanilla-framework/scss/vanilla";
 
 .slide-panel {
+  @include vf-animation(transform, brisk, out);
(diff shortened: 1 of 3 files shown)


src/components/SlidePanel/_slide-panel.scss
```diff
@@ -11,6 +11,12 @@
   position: fixed;
   right: 0;
   top: $top-header-height;
+  transform: translateX(100%);
+  transition: transform 0.5s;
```
hatched, line 15: This feels a bit slow, can we speed it up a bit?

anthonydillon: Please use Vanilla animation speeds please.

PR #658, opened 2020-08-20
Add apps panel to models details page

barrymcgee:
## Done

- Added apps panel to model details page

## QA

- Pull feature branch
- Go to models page, click a model
- On model details page, click an app
- Verify side panel opens from the right as expected with correct model info

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1368

### DIFF
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -233,7 +233,7 @@ export const debounce = (fn, wait) => {
 /**
   Returns a bool on whether a user can admin controllers
   @param {Obj} conn The user connection.
-  @returns {Bool} True or false
+  @returns {Boolean} If they are an admin or not.
 */
 export const userIsControllerAdmin = (conn) => {
   const controllerAccess = conn?.info?.user?.controllerAccess;
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -230,6 +230,11 @@ export const debounce = (fn, wait) => {
   };
 };
 
+/**
+  Returns a bool on whether a user can admin controllers
+  @param {Obj} conn The user connection.
+  @returns {Bool} True or false
+*/
 export const userIsControllerAdmin = (conn) => {
   const controllerAccess = conn?.info?.user?.controllerAccess;
   return ["superuser", "admin"].includes(controllerAccess);
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -239,3 +239,13 @@ export const userIsControllerAdmin = (conn) => {
   const controllerAccess = conn?.info?.user?.controllerAccess;
   return ["superuser", "admin"].includes(controllerAccess);
 };
+
+/**
+  Returns the charm name for provided namespace.
+  @param {String} namespace The fully qualified charm name.
+  @returns {String} The charm name.
+*/
+export const extractCharmName = (namespace) => {
+  if (!namespace) return null;
+  return namespace.replace("cs:~", "").split("/")[1].split("-")[0];
+};
--- a/src/app/utils.js
+++ b/src/app/utils.js
@@ -239,13 +239,3 @@ export const userIsControllerAdmin = (conn) => {
   const controllerAccess = conn?.info?.user?.controllerAccess;
   return ["superuser", "admin"].includes(controllerAccess);
 };
-
-/**
-  Returns the charm name for provided namespace.
-  @param {String} namespace The fully qualified charm name.
-  @returns {String} The charm name.
-*/
-export const extractCharmName = (namespace) => {
-  if (!namespace) return null;
-  return namespace.replace("cs:~", "").split("/")[1].split("-")[0];
-};
--- a/src/components/Filter/_filter.scss
+++ b/src/components/Filter/_filter.scss
@@ -41,7 +41,7 @@
   }
 
   &[aria-disabled="true"] {
-    opacity: 0;
+    opacity: 0.5;
     pointer-events: none;
   }
 }
--- a/src/components/Filter/_filter.scss
+++ b/src/components/Filter/_filter.scss
@@ -1,6 +1,8 @@
 @import "vanilla-framework/scss/vanilla";
 
 .p-filter {
+  @include vf-animation(opacity, brisk, out);
+
   display: grid;
   font-size: 0.875rem;
   gap: 0.5rem;
@@ -36,5 +38,10 @@
 
   & + & {
     padding-left: 1rem;
+  }
+
+  &[aria-disabled="true"] {
+    opacity: 0;
+    pointer-events: none;
   }
 }
--- a/src/components/SlidePanel/SlidePanel.js
+++ b/src/components/SlidePanel/SlidePanel.js
@@ -10,8 +10,19 @@ function SlidePanel({ children, onClose, isActive }) {
         onClose();
       }
     });
+    const closeOnClickOutside = document.addEventListener("click", (e) => {
+      console.log(e.target.closest('[role="row"]'));
+      if (
+        isActive &&
+        !e.target.closest(".slide-panel") &&
+        !e.target.closest('[role="row"]')
+      ) {
+        onClose();
+      }
+    });
     return () => {
       document.removeEventListener("keydown", closeOnEscape);
+      document.removeEventListener("click", closeOnClickOutside);
     };
   }, [isActive, onClose]);
 
--- a/src/components/SlidePanel/SlidePanel.js
+++ b/src/components/SlidePanel/SlidePanel.js
@@ -1,8 +1,20 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import "./_slide-panel.scss";
 
 function SlidePanel({ children, onClose, isActive }) {
+  // If Escape key is pressed when slide panel is open, close it
+  useEffect(() => {
+    const closeOnEscape = document.addEventListener("keydown", (e) => {
+      if (isActive && e.code === "Escape") {
+        onClose();
+      }
+    });
+    return () => {
+      document.removeEventListener("keydown", closeOnEscape);
+    };
+  }, [isActive, onClose]);
+
   return (
     <div className="slide-panel" aria-hidden={!isActive}>
       <button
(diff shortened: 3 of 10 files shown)


src/pages/Models/Details/ModelDetails.js
```diff
@@ -229,6 +245,33 @@ const ModelDetails = () => {
             )}
           </div>
         </div>
+        <SlidePanel
+          isActive={slidePanelActive}
+          onClose={() => setSlidePanelData({})}
+        >
+          {appSlidePanel}
```
barrymcgee, line 252: Will rename this to something more specific

src/pages/Models/Details/ModelDetails.js
```diff
@@ -104,6 +106,8 @@ const ModelDetails = () => {
 
   const [viewFilterToggle, setViewFilterToggle] = useState({ all: true });
 
+  const [slidePanelData, setSlidePanelData] = useState({});
```
hatched, line 109: I'm not sure this approach as implemented will work because of the subtle differences between the entity type displays.

Instead of this taking `{ app }` maybe it should instead just take the view type then we have a different component rendered into the panel depending on the type. It would then be this components responsibility to fetch the data it needs from the redux store.

(typescript I know, but it helps to illustrate the data type)
```typescript
enum Entity {
  app = "app",
  unit = "unit",
  machine = "machine",
  relation = "relation",
};

const panelType Entity = "app";

setSlidePanelData({ panelType });
```

src/pages/Models/Details/ModelDetails.js
```diff
@@ -124,48 +129,61 @@ const ModelDetails = () => {
...
-    const currentTarget = e.currentTarget;
-    if (currentTarget.classList.contains("is-selected")) {
-      setFilterByApp("");
-      return;
-    }
-    setFilterByApp(currentTarget.dataset.app);
+  const handleAppRowClick = (e, app) => {
+    console.log(app);
+    setSlidePanelData({ app });
+    setFilterByApp(e.currentTarget.dataset.app);
```
hatched, line 142: If the panel component is responsible for fetching its own data, as mentioned above, then none of this logic is necessary to filter the data here. It'll make this component simpler and we can get rid of more of the 'generator' logic below and have it more contained by the component.

src/pages/Models/Details/ModelDetails.js
barrymcgee: Noted but, as discussed, will refactor to this approach after this initial app panel lands.

hatched:
QA Notes:
- 🤔 Clicking the application name opens a new tab to the charm on `jujucharms.com`. It also opens the side panel, I'm guessing this is unintentional? I actually expected clicking the application name to open the panel and not the link. Maybe we remove that link and add it into the panel instead? It's a bit confusing that a row click opens the panel but the app name click opens the charm details page. 
- 🐞 When trying to click on an application row in the `hadoopspark` test model the application crashes with:
Screen Shot 2020-08-26 at 11 24 39 AM
- ✏️ The status text seems to be quite tight compared to the design, I think the line heights are off...
Screen Shot 2020-08-26 at 11 27 02 AM
- ✏️ The table designs don't match the design (and the columns don't align), I'm assuming this is a follow-up with the new table implementation?
- 🤔 Hitting the browsers back button doesn't close the tab as we don't create a history entry for it, should it?
- 👨‍🎨 Clicking to the left of the panel doesn't close the panel, I think it should as I keep trying to do that interaction :).

src/app/utils.js
```diff
@@ -230,7 +230,22 @@ export const debounce = (fn, wait) => {
...
   return ["superuser", "admin"].includes(controllerAccess);
 };
+
+/**
+  Returns the charm name for provided namespace.
+  @param {String} namespace The fully qualified charm name.
+  @returns {String} The charm name.
+*/
+export const extractCharmName = (namespace) => {
+  if (!namespace) return null;
```
hatched, line 249: Because the 'happy' path returns a `string` type, the failure path in this case should probably as well.

src/app/utils.js
```diff
@@ -230,7 +230,22 @@ export const debounce = (fn, wait) => {
   };
 };
 
+/**
+  Returns a bool on whether a user can admin controllers
+  @param {Obj} conn The user connection.
+  @returns {Bool} True or false
```
hatched, line 236: I think the type is `Boolean`
```suggestion
  @returns {Boolean} If they are an admin or not.
```

src/app/utils.js
```diff
@@ -230,7 +230,22 @@ export const debounce = (fn, wait) => {
...
 };
+
+/**
+  Returns the charm name for provided namespace.
+  @param {String} namespace The fully qualified charm name.
+  @returns {String} The charm name.
+*/
+export const extractCharmName = (namespace) => {
+  if (!namespace) return null;
+  return namespace.replace("cs:~", "").split("/")[1].split("-")[0];
```
hatched, line 250: This is the cause of the QA bug. It's pretty presumptive of the structure of the charm name. It'll need to be more resilient.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -179,16 +194,16 @@ const ModelDetails = () => {
               filters={viewFilters}
               setFilterToggle={setViewFilterToggle}
               filterToggle={viewFilterToggle}
+              disabled={slidePanelActive}
```
hatched, line 197: This is one way of doing it, but why not just not render it at all? 🤷

src/components/Filter/_filter.scss
```diff
@@ -37,4 +39,9 @@
   & + & {
     padding-left: 1rem;
   }
+
+  &[aria-disabled="true"] {
+    opacity: 0;
```
hatched, line 44: disabled is typically greyed out....or a 50%ish capacity not 0.

src/components/SlidePanel/SlidePanel.js
```diff
@@ -1,8 +1,20 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import "./_slide-panel.scss";
 
 function SlidePanel({ children, onClose, isActive }) {
+  // If Escape key is pressed when slide panel is open, close it
+  useEffect(() => {
+    const closeOnEscape = document.addEventListener("keydown", (e) => {
```
hatched, line 8: It looks like you missed this use case when adding the test suite.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -124,48 +125,62 @@ const ModelDetails = () => {
...
+    [baseAppURL, modelStatusData]
   );
+
   const unitTableRows = useMemo(
-    () => generateUnitRows(filteredModelStatusData, filterByApp, baseAppURL),
-    [baseAppURL, filterByApp, filteredModelStatusData]
+    () => generateUnitRows(modelStatusData, baseAppURL),
+    [baseAppURL, modelStatusData]
+  );
+  const unitSlidePanelRows = useMemo(
```
hatched, line 154: It might be nice to add a small comment to these lines that are going to be removed (or moved) when this is refactored as in the comment above.

src/pages/Models/Details/_model-details.scss
```diff
@@ -16,6 +16,11 @@
     @media (min-width: $breakpoint-m-large) {
       grid-template-columns: 350px 1fr;
     }
+
+    &[aria-disabled="true"] {
+      opacity: 0.5;
```
hatched, line 21: I think there is a 'disabled' state in Vanilla.

src/pages/Models/Details/generators.js
```diff
@@ -64,7 +65,13 @@ export function generateIconImg(name, namespace, baseAppURL) {
   );
 }
 
-export function generateEntityLink(namespace, name, subordinate, baseAppURL) {
+export function generateEntityIdentifier(
+  namespace,
+  name,
+  subordinate,
+  baseAppURL,
+  disableLink = false
```
hatched, line 73: This function is starting to get a bit overloaded...

src/pages/Models/Details/ModelDetails.js
barrymcgee: Because then it doesn't slide, it just appears.

src/pages/Models/Details/_model-details.scss
barrymcgee: Not as a standalone util.

barrymcgee:
> Hitting the browsers back button doesn't close the tab as we don't create a history entry for it, should it?

Nice to have, non blocker

> The table designs don't match the design (and the columns don't align), I'm assuming this is a follow-up with the new table implementation?

Correct.

> The status text seems to be quite tight compared to the design, I think the line heights are off...

Will defer to design review

ziheliu214:
Looks amazing Barry, Thanks!

long-chung:
Echoing what Jeff said in his comments.
Margin between rows in the apps header area (should be 4px, 0.25rem)
Reduce margin between unit. machine and relations.(should be 16px, 1rem)
Table row headings misalignment 
Seems to be a delay when you click the left side of the panel to close the window, also feels janky compared to if you close the panel hitting the X button.
Why is the charm name truncated? there's alot of space in that area

barrymcgee:
> Why is the charm name truncated? there's alot of space in that area

Notes go to the right, [see design](https://zpl.io/V108qxB)

> Table row headings misalignment

This uses the same table alignment as the underlying model details view. We'll soon be replacing all these tables with the new rewrite from Vanilla.

barrymcgee:
> Seems to be a delay when you click the left side of the panel to close the window, also feels janky compared to if you close the panel hitting the X button.

I removed errant, memory intensive, console.log which was causing this delay

long-chung:
- [ ] margin needs to be applied to the app name and the status in the header area
Otherwise good to go
Screenshot 2020-09-01 at 15 32 43

src/components/SlidePanel/SlidePanel.js
```diff
@@ -1,8 +1,30 @@
-import React from "react";
+import React, { useEffect } from "react";
 
 import "./_slide-panel.scss";
 
 function SlidePanel({ children, onClose, isActive }) {
+  // If Escape key is pressed when slide panel is open, close it
+  useEffect(() => {
```
hatched, line 7: This will still attach the events if the panel is open or not, maybe instead you could move the `isActive` conditional around the body of the `useEffect` function?

PR #669, opened 2020-09-01
Fix broken "N models" link at the top of Controllers page

benhoyt:
The `/models` link is a broken link and returns a 404. I believe this should be `/dashboard/models` (which works).

Context for this: I just started at Canonical on Monday (on the Juju team), and found this when playing with Juju locally and clicking around the GUI.

### DIFF
--- a/src/pages/Controllers/Controllers.js
+++ b/src/pages/Controllers/Controllers.js
@@ -386,7 +386,7 @@ export default function Controllers() {
       <Header>
         <div className="controllers--count">
           {controllerCount} controllers,{" "}
-          <a href="/models">{modelCount} models</a>
+          <a href="/dashboard/models">{modelCount} models</a>
         </div>
       </Header>
       <div className="l-content controllers">


src/pages/Controllers/Controllers.js
```diff
@@ -386,7 +386,7 @@ export default function Controllers() {
       <Header>
         <div className="controllers--count">
           {controllerCount} controllers,{" "}
-          <a href="/models">{modelCount} models</a>
+          <a href="/dashboard/models">{modelCount} models</a>
```
hatched, line 389: This is actually only the correct root path when deployed via Juju directly and not in any other environment. Thanks for finding this bug but I THINK the fix is actually to use the `Link` component from `react-router-dom` so that it resolves correctly using the defined base path.

I'm actually working in this file right now so I can include this fix as a driveby.....or if you'd like you can update this PR :)

(untested)
```diff
diff --git a/src/pages/Controllers/Controllers.js b/src/pages/Controllers/Controllers.js
index 6a06853..3cd2dcb 100644
--- a/src/pages/Controllers/Controllers.js
+++ b/src/pages/Controllers/Controllers.js
@@ -1,5 +1,6 @@
 import React, { useState } from "react";
 import { useDispatch, useSelector, useStore } from "react-redux";
+import { Link } from "react-router-dom";
 import cloneDeep from "clone-deep";
 import classNames from "classnames";

@@ -386,7 +387,7 @@ export default function Controllers() {
       <Header>
         <div className="controllers--count">
           {controllerCount} controllers,{" "}
-          <a href="/models">{modelCount} models</a>
+          <Link to="/models">{modelCount} models</Link>
         </div>
       </Header>
       <div className="l-content controllers">
```

benhoyt:
Thanks. Go ahead and fix in your other change!

hatched:
Sure thing, thanks again for the PR!

PR #673, opened 2020-09-02
Show all controllers in the controller lists

### LINKED ISSUE #606: Show all registered controlers
It's possible that a controller entry has been added that was invalid or cannot connect for whatever reason. These aren't currently shown. We should show them to have a location where we can display connection errors and also provide a location where the user can remove them.

Related: #605

hatched:
## Done

Previously controllers would only be shown in the controller list once they had connected. This branch renders all of the registered controllers whether they have connected or not.

This has not yet been designed so as an interim UI I've shown the websocket URL instead of the name and coloured it red to indicate an issue. The URL has a tooltip to indicate "disconnected". This will all be updated with a new UI in the near future.

Notes:
This register controller logic is needlessly complex and we will need to schedule some time to revisit this and clean it up.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Register a controller with invalid data, see it go red.
- Bootstrap a controller on your cloud of choice and register it, see it work as it should.

## Details

Fixes #606 

## Screenshots

Screen Shot 2020-09-02 at 4 45 58 PM


src/pages/Controllers/Controllers.js
```diff
@@ -90,15 +91,26 @@ function Details() {
     </span>
   );
 
+  function generatePathValue(c) {
+    const column = { content: "" };
+    column.content = c.path === "admin/jaas" ? "JAAS" : c.path;
+    if (!column.content) {
+      column.content = c.wsControllerURL;
+      column.className = "disconnected-controller";
```
anthonydillon, line 99: I see how you got here but setting a status of the connection here would be easier to reason here and leave the presentation of the controller states to the render function or presentational function so the concerns are separated.

src/pages/Controllers/Controllers.js
```diff
@@ -90,15 +91,26 @@ function Details() {
     </span>
   );
 
+  function generatePathValue(c) {
+    const column = { content: "" };
+    column.content = c.path === "admin/jaas" ? "JAAS" : c.path;
+    if (!column.content) {
```
anthonydillon, line 97: I think it would be much more clear to use a set of 'if' statements here to set column.content. you are mixing ternary operator with traditional if statement to achieve this. The use of different techniques takes a min for the reader to understand it's all part of the same logic.

src/pages/Controllers/_controllers.scss
```diff
@@ -45,6 +45,10 @@
     }
   }
 
+  .disconnected-controller {
```
anthonydillon, line 48: Please use a state class syntax here. For example, `is-disconnected`. Makes the purpose of the class more clear to other developers. 

https://canonical-web-and-design.github.io/practices/coding/stylesheets.html#state-classes

anthonydillon (changes requested):
Some comments in the code. If there is retractor required there is no time like the present while it's in your mind.

src/pages/Controllers/Controllers.js
```diff
@@ -90,15 +91,26 @@ function Details() {
     </span>
   );
 
+  function generatePathValue(c) {
```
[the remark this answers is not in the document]
hatched: heh, my bad, I've been doing a lot of typed work lately in which this would have been `function generatePathValue(c: ControllerData) {` :) 

I've updated this.

src/pages/Controllers/Controllers.js
hatched: I've updated this to use a single set of conditionals.

src/pages/Controllers/Controllers.js
hatched: This function is the function that determines what gets presented as it's called by the table generation code. I could add a `connected` value further up in the stack but I don't want to keep two sources of truth around the connected status. One being the actual connection object and this being the other. In the future these two will be merged so it will be less of an issue. But anyways, for now I think this is the best place for this logic.

src/pages/Controllers/_controllers.scss
hatched: good call, done.

anthonydillon (approved):
Nice, reads so much clearer now. :+1:

PR #678, opened 2020-09-11
Convert filter buttons into a ButtonGroup to toggle tables in the Model Details view...and more...

hatched:
## Done

- Converted `ModelGroupToggle` component to a generic `ButtonGroup` component.
- Fixed bug in `FilterTags` that caused multiple history entries to be added if multiple components tried to update the query strings.
- Fixed bug in `LogIn` that caused the whole application to be destroyed and re-rendered on initial load.
- Created `useQueryString` hook that can be used directly by the `ButonGroup` component, or independently, to update a segment of the queryString for the component in question.
- Updated the `ModelDetails` component so that it only shows a single table depending on the selected view from the aforementioned `ButtonGroup`.
- 

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The application should log you in as expected.
- When viewing your models list you should be able to click the grouping buttons to change the grouping and the back button should work as expected.
- After clicking a model to view its details, clicking the view buttons should change the table that's shown and the back button should work as expected.
- Viewing the model details page at different widths, the view button group should move depending on the 3 breakpoints.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1365

## Screenshots

Screen Shot 2020-09-10 at 10 02 12 PM

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -3,21 +3,21 @@ import classNames from "classnames";
 
 import "./_button-group.scss";
 
-const ButtonGroup = ({ buttons, groupedBy, setGroupedBy }) => {
+const ButtonGroup = ({ buttons, label, activeButton, setActiveButton }) => {
   return (
     <div className="p-button-group">
       <div className="p-button-group__inner">
-        <span className="p-button-group__label">Group by:</span>
+        <span className="p-button-group__label">{label}</span>
         <div className="p-button-group__buttons">
           {buttons.map((label) => (
             <button
-              aria-label={`group by ${label}`}
+              aria-label={`view by ${label}`}
               key={label}
               className={classNames("p-button-group__button", {
-                "is-selected": groupedBy === label,
+                "is-selected": activeButton === label,
               })}
               value={label}
-              onClick={(e) => setGroupedBy(e.currentTarget.value)}
+              onClick={(e) => setActiveButton(e.currentTarget.value)}
             >
               {label}
             </button>
--- a/src/components/ButtonGroup/ButtonGroup.js
+++ b/src/components/ButtonGroup/ButtonGroup.js
@@ -1,21 +1,19 @@
 import React from "react";
 import classNames from "classnames";
 
-import "./_model-group-toggle.scss";
+import "./_button-group.scss";
 
-const buttons = ["status", "cloud", "owner"];
-
-const ModelGroupToggle = ({ groupedBy, setGroupedBy }) => {
+const ButtonGroup = ({ buttons, groupedBy, setGroupedBy }) => {
   return (
-    <div className="p-model-group-toggle">
-      <div className="p-model-group-toggle__inner">
-        <span className="p-model-group-toggle__label">Group by:</span>
-        <div className="p-model-group-toggle__buttons">
+    <div className="p-button-group">
+      <div className="p-button-group__inner">
+        <span className="p-button-group__label">Group by:</span>
+        <div className="p-button-group__buttons">
           {buttons.map((label) => (
             <button
               aria-label={`group by ${label}`}
               key={label}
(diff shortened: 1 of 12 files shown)


src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import classNames from "classnames";
+
+import "./_button-group.scss";
+
+const ButtonGroup = ({ buttons, label, activeButton, setActiveButton }) => {
```
hatched, line 6: Once this lands I'll be proposing this to the Vanilla WG.

src/components/FilterTags/FilterTags.js
```diff
@@ -109,14 +109,14 @@ const FilterTags = () => {
 
   // Update query params when adding and removing filters
   useEffect(() => {
-    const queryParams = queryString.parse(location.search);
+    const queryParams = queryString.parse(window.location.search);
```
hatched, line 112: This component needs to do a better job of it's queryString handling but I figured I should leave it until we switch to the new Filter UI.

src/components/LogIn/LogIn.js
```diff
@@ -45,11 +45,10 @@ export default function LogIn({ children }) {
             {generateErrorMessage(loginError)}
           </div>
         </div>
-        <main>{children}</main>
```
hatched, line 48: This implementation was causing the whole app to be torn down and re-rendered after logging in.

src/pages/Models/Details/_model-details.scss
```diff
@@ -26,10 +26,16 @@
...
+    display: grid;
+    grid-template-columns: 1fr;
 
     @media (min-width: $breakpoint-medium) {
-      display: flex;
+      gap: 1rem;
+      grid-template-columns: 240px 1fr;
+    }
+
+    @media (min-width: $breakpoint-m-large) {
```
hatched, line 37: I'm not sure we want to be creating custom breakpoints, we should investigate instead overriding the vanilla `breakpoint-x-large` instead.

src/components/ButtonGroup/ButtonGroup.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div className="p-button-group">
+      <div className="p-button-group__inner">
+        <span className="p-button-group__label">{label}</span>
+        <div className="p-button-group__buttons">
+          {buttons.map((label) => (
+            <button
+              aria-label={`view by ${label}`}
```
barrymcgee, line 14:
```suggestion
              aria-label={`View by ${label}`}
```

barrymcgee:
Select several filter tags and refresh page:

Screenshot 2020-09-11 at 10 33 36

src/pages/Models/Details/_model-details.scss
barrymcgee: We don't want to override vanilla breakpoints that may break any components relying on that existing breakpoint value.

src/components/FilterTags/FilterTags.js
barrymcgee: Shouldn't this use your new useQuery hook?

hatched:
The demo service doesn't support loading the dashboard from anything other than `/`.

src/components/FilterTags/FilterTags.js
hatched: Yes but this PR was already feature creeping considerably.

src/components/ButtonGroup/ButtonGroup.js
barrymcgee: Screen readers are not the only assistive technology to use ARIA. However, some screen readers use capitalisation as cues for intonation.

hatched: Cool, I'll make a note to go update the rest of them in a follow-up.

barrymcgee (approved):

PR #681, opened 2020-09-14
Add entity counts to each view in the model details page

hatched:
## Done

- Added counts to the top of each entity view in the model details page.
- Updated pluralizer to support special case words.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a few models and look at the counts on each entity view.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1412
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1413
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1414

## Screenshots

Screen Shot 2020-09-14 at 4 55 20 PM

### DIFF
--- a/src/components/Counts/Counts.js
+++ b/src/components/Counts/Counts.js
@@ -5,7 +5,7 @@ import "./_counts.scss";
 
 const Counts = ({ primaryEntity, secondaryEntities }) => {
   const statuses = secondaryEntities
-    .map((entity) => {
+    ?.map((entity) => {
       return `${entity.count} ${pluralize(entity.count, entity.label)}`;
     })
     .join(", ");
--- a/src/components/Counts/Counts.js
+++ b/src/components/Counts/Counts.js
@@ -10,10 +10,10 @@ const Counts = ({ primaryEntity, secondaryEntities }) => {
     })
     .join(", ");
   return (
-    <div className="entity-counts">
+    <small className="entity-counts">
       {primaryEntity.count}{" "}
       {pluralize(primaryEntity.count, primaryEntity.label)}: {statuses}
-    </div>
+    </small>
   );
 };
 
--- a/src/components/Counts/Counts.js
+++ b/src/components/Counts/Counts.js
@@ -1,0 +1,20 @@
+import React from "react";
+import { pluralize } from "app/utils";
+
+import "./_counts.scss";
+
+const Counts = ({ primaryEntity, secondaryEntities }) => {
+  const statuses = secondaryEntities
+    .map((entity) => {
+      return `${entity.count} ${pluralize(entity.count, entity.label)}`;
+    })
+    .join(", ");
+  return (
+    <div className="entity-counts">
+      {primaryEntity.count}{" "}
+      {pluralize(primaryEntity.count, primaryEntity.label)}: {statuses}
+    </div>
+  );
+};
+
+export default Counts;
--- a/src/components/Counts/_counts.scss
+++ b/src/components/Counts/_counts.scss
@@ -1,0 +1,5 @@
+.entity-counts {
+  font-size: 0.875rem;
+  line-height: 0.875rem;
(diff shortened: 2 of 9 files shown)


anthonydillon (changes requested):
Some general issues:
- Design-wise I find this uninspired. It feels like cluttered noise which the user will skip over regularly.
- Could you add some tests for the Counts component.
- There is some simular functions here but does result in clean usage in the `renderCounts` method. I think its ok but not very DRY

hatched:
Oops sorry this one was supposed to be marked as a draft. But I'll address your comments 👍

barrymcgee:
Agree with @anthonydillon that design jars - particularly as it pushes the tables down the page. There is also no distinction between count labels and the count values.

src/components/Counts/Counts.js
```diff
@@ -0,0 +1,20 @@
...
+import "./_counts.scss";
+
+const Counts = ({ primaryEntity, secondaryEntities }) => {
+  const statuses = secondaryEntities
+    ?.map((entity) => {
+      return `${entity.count} ${pluralize(entity.count, entity.label)}`;
+    })
+    .join(", ");
+  return (
+    <div className="entity-counts">
```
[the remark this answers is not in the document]
hatched: done!

src/components/Counts/_counts.scss
```diff
@@ -0,0 +1,5 @@
+.entity-counts {
```
hatched, line 1: These styles are necessary to position the `small` in the correct location as per the design.

src/pages/Models/Details/ModelDetails.test.js
```diff
@@ -36,7 +36,7 @@ describe("ModelDetail Container", () => {
     const wrapper = mount(
       <Provider store={store}>
         <MemoryRouter
-          initialEntries={["/models/space-man@external/frontend-ci"]}
+          initialEntries={["/models/spaceman@external/hadoopspark"]}
```
hatched, line 39: The previous model didn't even exist...This test only passed because it was simply checking for the existence of the tables not that they were populated.

anthonydillon (approved):
Thanks :+1:

anthonydillon:
I'm not keen on the counts jumping from the header to the table panel but it matches the design so :+1:

spencerbygraves:
@clagom @ziheliu214 do you want to address the comments above?

Not sure I should give design +1 at this point.

clagom:
The width of the panel covers the list on the left completely, is it going to be addressed later?

clagom:
> @clagom @ziheliu214 do you want to address the comments above?
> 
> Not sure I should give design +1 at this point.

I am not sure what the comment refers to

clagom:
Screenshot 2020-09-16 at 14 06 49
I thought we went with this design to allow longer charm name (cs:~hatch/preci...) so the truncation seems unnecesary since there's a lot of space.
That said, we also discussed to implement the other design, since we don't need to display the whole "cs:~hatch/preci..." since it's a link?

clagom:
The background behind the panel seems to get semitransparent:
Screenshot 2020-09-16 at 14 08 48

This panel is not a modal, so there shouldn't be any background when the panel is open

ziheliu214:
@hatched @spencerbygraves @clagom 
1) the count on the 'status' tab is only showing apps which isn't right, making users ignore the units machines and relations below
2) generally can we try adding colour dots to the count or making the text coloured? This was mentioned by users in user research a lot.

Thanks!

clagom:
I just realised I commented in a wrong PR, sorry about that. Please let me know if I should open separate issue for each comment I made @hatched

hatched:
>     1. the count on the 'status' tab is only showing apps which isn't right, making users ignore the units machines and relations below

Each page has its own count summary so the status only has apps (as per the design) units has units, etc.

PR #682, opened 2020-09-16
Refactor apps panel

### LINKED ISSUE #677: Controller view on master is broken
Run master and login. Click on Controllers in the navigation and you get the following error:
```
TypeError: controllerData is undefined
```

barrymcgee:
## Done

- Refactor apps panel to decouple from ModelDetails
- Added URL persistence through query strings
- Added loading state to slide panels
- Driveby: Short circuit `getControllerDataByUUID` if `!controllerData` 
- Driveby: add conditional checks to couple of util queries 

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click app in model details - verify correct app info appears
- Refresh the page, verify panel persists with loading state

## Details

Fixes #677, Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1415


src/app/selectors.js
```diff
@@ -565,6 +565,7 @@ export const getWSControllerURL = createSelector(
 */
 export const getControllerDataByUUID = (controllerUUID) => {
   return createSelector(getControllerData, (controllerData) => {
+    if (!controllerData) return null;
```
hatched, line 568: When I attempted this fix myself locally I ended up with intermittent failures but on your branch it seems to work as expected 🤷‍♂️ 

Lets keep our eyes out for a race here.

src/app/utils.js
```diff
@@ -252,3 +253,70 @@ export const userIsControllerAdmin = (conn) => {
   const controllerAccess = conn?.info?.user?.controllerAccess;
   return ["superuser", "admin"].includes(controllerAccess);
 };
+
+/**
+  Returns the modelStatusData filtered by the supplied values.
+  @param {Object} modelStatusData The model status data to filter
+  @param {String} appName The name of the application to filter the data by.
+*/
+export const filterModelStatusData = (modelStatusData, appName) => {
```
hatched, line 262: Is this how we want to do this? I feel like a better approach might be to have selectors like the following so that we can use them elsewhere.
- `getUnitsByApplication()`
- `getMachinesByApplication()` 
- etc...

src/hooks/useQueryString.js
```diff
@@ -43,10 +43,13 @@ function useQueryString(queryKey, resetValue) {
     }
     queryStrings[queryKey] = queryValue;
     const updatedQs = queryString.stringify(queryStrings);
-    history.push({
-      pathname,
-      search: queryValue === resetValue ? null : updatedQs,
-    });
+    // Don't allow empty query string values
+    if (queryValue) {
```
hatched, line 47: I'm curious to find out why this change was necessary? Technically a query parameter without a value is still valid so I don't think this is the correct fix for whatever issue it's resolving.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -269,51 +200,44 @@ const ModelDetails = () => {
...
-    () => generateRelationRows(filteredModelStatusData, baseAppURL),
-    [filteredModelStatusData, baseAppURL]
-  );
-
-  const appSlidePanelHeader = useMemo(
-    () => generateAppSlidePanelHeader(slidePanelData.currentApp, baseAppURL),
-    [slidePanelData.currentApp, baseAppURL]
-  );
 
   const slidePanelActive = Object.entries(slidePanelData).length > 0;
```
hatched, line 236: This is being used for the aria disabled toggle but I don't see how it can result in anything but being always 'enabled' as no value is ever defined for `slidePanelData`.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -226,8 +161,7 @@ const ModelDetails = () => {
   const { 0: modelName } = useParams();
   const dispatch = useDispatch();
   const [filterByApp, setFilterByApp] = useState("");
-
-  const [slidePanelData, setSlidePanelData] = useState({});
+  const [slidePanelData] = useState({});
```
hatched, line 164: This is unused and can be removed. (see below)

src/pages/Models/Details/ModelDetails.js
```diff
@@ -269,51 +200,44 @@ const ModelDetails = () => {
...
-  const appSlidePanelHeader = useMemo(
-    () => generateAppSlidePanelHeader(slidePanelData.currentApp, baseAppURL),
-    [slidePanelData.currentApp, baseAppURL]
-  );
 
   const slidePanelActive = Object.entries(slidePanelData).length > 0;
 
   const [activeView, setActiveView] = useQueryString("view", "status");
+  const [panelType, setPanelType] = useQueryString("panel");
+  const [entity, setEntity] = useQueryString("entity");
```
hatched, line 240: Ahh here is the problem that's causing the QA failure, it's creating two history entries, one for each of these two values. We'll have to use a single query string or update `useQueryString` to be able to support multiple params at once. 

Lets chat about this one.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -269,51 +200,44 @@ const ModelDetails = () => {
...
     const currentApp = cloneDeep(app);
     currentApp.name = e.currentTarget.dataset.app;
-    setSlidePanelData({ currentApp });
     setFilterByApp(currentApp.name);
+    setPanelType("apps");
+    setEntity(currentApp.name);
+  };
+
+  const closePanel = () => {
+    history.push(`/models/${modelName}`);
```
hatched, line 209: If you watch the entries when closing the panel it loses it's mind creating new entries and rendering.

I think that you will only need to set the `panel` and `entity` to empty. But this will involve the discussion from the other qa issue below.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -375,38 +299,12 @@ const ModelDetails = () => {
...
-                emptyStateMsg={"There are no relations in this model"}
-              />
-            </div>
-          </>
-        </SlidePanel>
+        <AppsPanel
+          entity={entity}
+          isActive={panelType}
+          onClose={() => closePanel()}
+          filterByApp={filterByApp}
```
hatched, line 306: This is unused.

hatched (changes requested):
QA Note:
This works really well, I like the loading indicator when you refresh...
When you're viewing an apps panel however the back button is 'broken'. To reproduce:
- View an application panel.
- Hit the back button, it will show a 'loading' spinner. It appears this is because the `entity` value is now missing from the url.
- Hit back again, it'll go to the status view as expected.
Also if you close the panel it creates an excessive number of entries and has difficulty rendering all those states.
See comments in the code for these two issues.

src/app/utils.js
barrymcgee: Perhaps.. that's for another PR though.

hatched: Yup, no eyeballs on it :D

src/hooks/useQueryString.js
barrymcgee: It was because when closing the panel, the url would be `/model?panel=&app=`

This may however be a symptom of the other issue you've described.

clagom:
I opened the issue about the background already, do you have any other points you want to address @spencerbygraves @ziheliu214 ?

spencerbygraves:
Looking good, a couple of things from me.

- Does it have the same speed as the left nav panel? Difficult to tell because of the distance of travel
- On my screen I don't see the app names just the icons. However, this is probably related to the [other issue](https://github.com/canonical-web-and-design/jaas-dashboard/issues/683) that is being [addressed in Vanilla](https://github.com/canonical-web-and-design/vanilla-framework/issues/3282)
- Can we align the top icon with those in the tables below please
image

Thanks.

barrymcgee:
> Does it have the same speed as the left nav panel? Difficult to tell because of the distance of travel

Yes, they are both the Vanilla speed of 'brisk'

> On my screen I don't see the app names just the icons. However, this is probably related to the other issue that is being addressed in Vanilla

Yes, that's correct.. it's currently set to 75% of screen width.

src/components/ButtonGroup/ButtonGroup.test.js
```diff
@@ -1,17 +1,20 @@
...
 
 describe("ButtonGroup", () => {
   it("shows active button", () => {
     const wrapper = mount(
       <MemoryRouter>
-        <ButtonGroup
-          buttons={["status", "cloud", "owner"]}
-          activeButton="cloud"
-        />
+        <QueryParamProvider ReactRouterRoute={Route}>
```
hatched, line 12: The `ButtonGroup` component doesn't do anything with the query params or routes so I think this and `MemoryRouter` can be removed in these tests.

src/pages/Models/Details/ModelDetails.js
```diff
@@ -266,54 +201,37 @@ const ModelDetails = () => {
...
+  };
+
+  const closePanel = () => {
+    setQuery({ panel: undefined, entity: undefined });
   };
 
   const applicationTableRows = useMemo(
     () =>
       generateApplicationRows(modelStatusData, handleAppRowClick, baseAppURL),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
hatched, line 215: Can you add a comment why this is required? Usually when this is required something else is wrong :)

hatched (approved):
Thanks @barrymcgee Just a couple small cleanups +1 (with changes)

src/components/ButtonGroup/ButtonGroup.test.js
barrymcgee: Removed.

src/pages/Models/Details/ModelDetails.js
barrymcgee: Refactored useMemo so this exception is no longer required.

spencerbygraves (approved):

PR #686, opened 2020-09-18
Add an offline banner

barrymcgee:
## Done

Whipped up an offline banner this evening as an exercise in creating a custom hook - thoughts?

## QA

- View the demo
- Toggle your internet connection via dev tools or the [Work Offline Chrome extension](https://chrome.google.com/webstore/detail/work-offline/gcjdagaoojgeilcobhibkgbodajcojda?hl=en) 
- See the banner in action

Screen Recording 2020-09-18 at 10 19 47 pm

### DIFF
--- a/src/components/Banner/Banner.js
+++ b/src/components/Banner/Banner.js
@@ -2,6 +2,10 @@ import React from "react";
 
 import "./_banner.scss";
 
-export default function Banner({ children }) {
-  return <div className="banner">{children}</div>;
+export default function Banner({ isActive, children }) {
+  return (
+    <div className="banner" data-active={isActive}>
+      {children}
+    </div>
+  );
 }
--- a/src/components/Banner/Banner.js
+++ b/src/components/Banner/Banner.js
@@ -2,9 +2,9 @@ import React from "react";
 
 import "./_banner.scss";
 
-export default function Banner({ isActive, children }) {
+export default function Banner({ isActive, children, type }) {
   return (
-    <div className="banner" data-active={isActive}>
+    <div className="banner" data-active={isActive} data-type={type}>
       {children}
     </div>
   );
--- a/src/components/Banner/Banner.js
+++ b/src/components/Banner/Banner.js
@@ -1,0 +1,7 @@
+import React from "react";
+
+import "./_banner.scss";
+
+export default function Banner({ children }) {
+  return <div className="banner">{children}</div>;
+}
--- a/src/components/Banner/Banner.test.js
+++ b/src/components/Banner/Banner.test.js
@@ -1,0 +1,30 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import Banner from "./Banner";
+
+describe("Banner", () => {
+  it("should display banner text", () => {
+    const wrapper = shallow(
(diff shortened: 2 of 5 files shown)


spencerbygraves:
@barrymcgee there is some work happening on notifications at the moment, but might it be better to use the current Vanilla ones so they can be updated easily?

spencerbygraves:
Good work though!

cassiocassio:
CC @Will Grant <will.grant@canonical.com>

will:
Hey Martin, you tried to cc Will Grant, but actually you cc’d me, since that is how GitHub at mentions work.

hatched:
@wgx

barrymcgee:
@spencerbygraves I did think of using Vanilla notification pattern but they felt quite intrusive here as they typically don't sit flush with their parent container and require external margins. This kind of strip banner is more subtle and more typically used in this context.. see mattermost etc.

wgx:
> @spencerbygraves I did think of using Vanilla notification pattern but they felt quite intrusive here as they typically don't sit flush with their parent container and require external margins. This kind of strip banner is more subtle and more typically used in this context.. see mattermost etc.

@barrymcgee As Spencer mentioned, I'm working on some Notifications patterns at the moment. I'd just like to understand what makes this banner 'special', and warrants an additional piece of new UI. If we can deliver the experience with existing UI it'll make upgrading Notifications easier in future.

hatched:
@wgx the primary issue with the notification is that in this case (due to technical limitations) we need to notify the user that they must refresh to connect and the application is inactive until that time. This is something that we plan on working towards resolving in the next cycle but in the short term we need something with prominence.

barrymcgee:
@wgx More holistically, I see this as a 'strip' or 'banner' notification which applies site wide as opposed to the current Vanilla notification which feels to me to be more contextual to the view in which is sits?

wgx:
@barrymcgee Makes sense. The plan for the new Notificaitons pattern is to make it much more 'high level' and less bound to each context. Have a browse of the early sketches here: https://miro.com/app/board/o9J_kmE7rLM=/?moveToWidget=3074457349876052680&cot=12 for an idea of what we have in mind.

barrymcgee:
This noticed the Zeplin web app implementation

Screenshot 2020-09-22 at 11 45 19

clagom:
Definitely a use-case to consider for the notification pattern.

spencerbygraves:
I have no problem with moving towards something like this, but we already have a notification on the dashboard using the current styling, which also seems to be focused site wide.

If UX don't mind that, and you don't mind updating something bespoke to a Vanilla component in the future, then perhaps we can use this.

barrymcgee:
Reflecting decision outcome chatted about in Mattermost - @spencerbygraves and I have agreed while notification work is ongoing in Vanilla, this solution can land in the interim as it brings immediate value to users.

Vanilla working group proposal: https://github.com/canonical-web-and-design/vanilla-framework/issues/3298

spencerbygraves (approved):
Thanks @barrymcgee, great work.

hatched (approved):

PR #709, opened 2020-10-12
Enable panel tables on model details page

barrymcgee:
## Done

- Enable panel tables on model details page
- Change active status dot to green
- Add a loading spinner to model details page
- Conditionally render model details tables based on data available

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click on a model
- Click the first app
- Observe app panel open
- You can now navigate the tables to slice and dice across apps/units and machines
- Refresh the page
- Observe the loading spinner

## Details

Related: https://github.com/canonical-web-and-design/juju-squad/issues/1388

### DIFF
--- a/src/components/panels/AppsPanel/AppsPanel.js
+++ b/src/components/panels/AppsPanel/AppsPanel.js
@@ -1,7 +1,6 @@
 import React, { useMemo } from "react";
 import { useSelector } from "react-redux";
 import { getConfig } from "app/selectors";
-import SlidePanel from "components/SlidePanel/SlidePanel";
 import MainTable from "@canonical/react-components/dist/components/MainTable";
 
 import useModelStatus from "hooks/useModelStatus";
@@ -24,12 +23,7 @@ import {
 
 import "./_apps-panel.scss";
 
-export default function AppsPanel({
-  isActive,
-  onClose,
-  entity,
-  panelRowClick,
-}) {
+export default function AppsPanel({ entity, panelRowClick }) {
   // Get model status info
   const modelStatusData = useModelStatus();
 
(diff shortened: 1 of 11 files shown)


src/components/panels/UnitsPanel/UnitsPanel.js
```diff
@@ -1,24 +1,63 @@
...
+  onClose,
+  entity: unitId,
+  panelRowClick,
+}) {
   const modelStatusData = useModelStatus();
   const appName = unitId?.split("/")[0];
   const unit = modelStatusData?.applications[appName]?.units[unitId];
   const app = modelStatusData?.applications[appName];
 
+  const filteredModelStatusDataByMachine = useCallback(
```
hatched, line 30: Doing this all the time is very inefficient, we should move this into an optimized selector.

hatched (changes requested):
This works great, navigating around them is buttery smooth 👍

src/components/panels/UnitsPanel/UnitsPanel.js
hatched: This is for a follow-up sometime, not for now :)

hatched:
QA Issue:

when viewing the broken model `gomboli@external/mymodel` and then clicking on any of the machines `flannel` is listed. It's a subordinate but I don't think it should be on every machine.

Screen Shot 2020-10-13 at 2 24 58 PM

src/components/panels/AppsPanel/AppsPanel.js
```diff
@@ -24,7 +24,12 @@ import {
 
 import "./_apps-panel.scss";
 
-export default function AppsPanel({ isActive, onClose, entity }) {
+export default function AppsPanel({
+  isActive,
+  onClose,
+  entity,
+  panelRowClick,
```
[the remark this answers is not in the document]
barrymcgee: These panels already are, by design, tightly coupled to ModelDetails. They are the coupled implementation of the generic uncoupled `SlidePanel` component.

I investigated moving the update function into a hook but it already depends on the useQueryParams hooks which is also being used in ModelDetails to control the activeView filter. Refactoring all that at this stage is not worth the cost in my opinion.

barrymcgee:
Flannel was slipping through as it didn't have any units to interrogate. Now resolved.

src/components/panels/MachinesPanel/MachinesPanel.js
```diff
@@ -86,6 +90,73 @@ export default function MachinesPanel({
...
+        Object.keys(filteredModelStatusData.applications).forEach(
+          (application) => {
+            const units =
+              filteredModelStatusData.applications[application]?.units;
+
+            if (Object.entries(units).length) {
+              Object.values(units).forEach((unit) => {
+                if (
+                  unit.machine !== machineId ||
+                  !Object.entries(units).length
```
hatched, line 106: Can you add a comment in the code about this as it's not clear why this is being done

hatched (approved):

clagom:
Nice! Not a blocker, but I wonder if we can refresh the content of the panel instead of closing/opening it again when clicking on a unit or machine from an app panel. We refresh the content when clicking on other apps from the list instead of closing/opening the panel again, so we could apply the same. @barrymcgee @hatched

barrymcgee:
@clagom Good call.. that change actually simplifies the code too. I've now made that change.. can you re-review?

src/components/panels/MachinesPanel/MachinesPanel.js
barrymcgee: Done

PR #714, opened 2020-10-15
Unmount slide panels after animation completion

barrymcgee:
## Done

Unmount slide panels after animation completion

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify panel component is unmounted after animation completes


anthonydillon (commented):
Should we add a test for this?

PR #715, opened 2020-10-16
WIP: Fade in and out different panel types on model details

barrymcgee:
## Done

@clagom @spencerbygraves @long-chung 

After consolidating all model details panel types within one single SlidePanel.. I experimented with a little fade in/out animation for when you switch panel types..

## QA

- Click through to demo
- Go to model details
- Click an app.. any click to another entity type within the panel will trigger the 1s fade out/in animation.. clicking an entity of the same type will not.. (looking at an app and you click another app.. no animation... looking at an app but you click a unit.. animation)..

Yay or nay?

(Gif below is illustrative but janky.. click through to demo to view in browser.)

Screen Recording 2020-10-16 at 10 47 27 am

### DIFF
--- a/src/animations/FadeInOut/FadeInOut.scss
+++ b/src/animations/FadeInOut/FadeInOut.scss
@@ -1,0 +1,27 @@
+@keyframes fadeIn {
+  0% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 1;
+  }
+}
+
+@keyframes fadeOut {
+  0% {
+    opacity: 1;
+  }
+
+  100% {
+    opacity: 0;
+  }
+}
+
+.fade-in {
+  animation: fadeIn 0.5s;
+}
+
+.fade-out {
+  animation: fadeOut 0.5s;
+}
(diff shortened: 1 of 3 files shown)


src/animations/FadeInOut/FadeInOut.scss
```diff
@@ -0,0 +1,27 @@
+@keyframes fadeIn {
+  0% {
+    opacity: 0;
+  }
+
+  100% {
+    opacity: 100%;
```
anthonydillon, line 7: Opacity takes values from `0` to `1` therefore, this should be `1`.

long-chung: I'm not seeing the fade effect on either chrome or firefox. Just getting an almost show/hide effect i.e white screen then content

spencerbygraves:
@barrymcgee the gif looks good but the demo is showing an on/off effect, the same as @long-chung says.

barrymcgee:
@spencerbygraves @long-chung Remember the animation only fires if you change panel type.. so click one of the table rows inside the apps panel.

spencerbygraves:
@barrymcgee ok, it's working for me now. I like the animation, is there a reason you think it shouldn't also be used when you change apps?

anthonydillon:
Looks good to me. Feels very fast!

long-chung:
Nice!

PR #730, opened 2020-10-26
Add feature flag for WebCLI

barrymcgee:
## Done

Add feature flag for WebCLI

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check WebCLI does not appear
- Open `config.js`
- Change `showWebCLI` to `true`
- WebCLI should now appear on model details page

### DIFF
--- a/public/config.js
+++ b/public/config.js
@@ -12,4 +12,6 @@ var jujuDashboardConfig = {
   // only be used for superficial updates like logos. Use feature detection
   // for other environment features.
   isJuju: false,
+  // Show WebCli component if enabled for dev purposes, even if not natively supported.
+  showWebCLI: false,
 };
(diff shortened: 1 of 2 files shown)


hatched (approved):

hatched:
So this obviously doesn't actually work with Juju under 2.9 but I can see the use for it in the short term.

public/config.js
```diff
@@ -12,4 +12,6 @@ var jujuDashboardConfig = {
   // only be used for superficial updates like logos. Use feature detection
   // for other environment features.
   isJuju: false,
+  // Show WebCli component if enabled
```
anthonydillon, line 15: Is this accurate? Is it more like force display the WebCLI even if natively unsupported.

barrymcgee: The WebCLI has a natively unsupported state. I'll update the comment to explicitly state it's for dev purposes.

PR #731, opened 2020-10-27
Use event listener hook

barrymcgee:
## Done

Move complexity around event listeners in useEffects into a custom hook where instantiation and cleanup are abstracted away.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Ensure all tests pass
- Check code


hatched (changes requested):
`Output.js` and `useOffline.js` also use `addEventListener` and can take advantage of this.

barrymcgee:
As discussed in Mattermost with @hatched .. 

- useOffline doesn't use it because I think custom hooks shouldn't depend on each other
- Output doesn't use it because of complexity (and possible bug) in cleanup

hatched (approved):

PR #735, opened 2020-10-27
Update site wide offline banner

### LINKED ISSUE #694: Small design updates for the offline banner
@barrymcgee until we have the finished notifications work, here are a few thoughts on improving what we have.

- [x] I've updated the padding to have .5rem top and bottom
- [x] Centred the text, currently it looks slightly off
- [x] I've added a dismiss icon on the right. @wgx mentioned in his demo that our users may find this useful
- [x] If we're true to the [Vanilla spec for inverted links](https://vanillaframework.io/docs/patterns/links#inverted), it should be 400 weight and underline on hover
- [x] I amended the text slightly on both

**Offline**

image

**Online**

image

Thanks

barrymcgee:
## Done

Update site wide offline banner

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify changes requested in #694

## Details

 Fixes #694


spencerbygraves (approved):
Thanks @barrymcgee, looks good.

src/components/Banner/Banner.js
```diff
@@ -1,11 +1,32 @@
-import React from "react";
+import React, { useState } from "react";
+import classnames from "classnames";
 
 import "./_banner.scss";
 
-export default function Banner({ isActive, children, type }) {
+export default function Banner({ isActive = true, children, type }) {
```
hatched, line 6: Shouldn't the default state of this be `false`?

hatched (approved):

hatched:
QA Issue

If you hit the `X` while offline to close the banner then you can never get it back again. I would have expected that it would show up again when we went back online.

src/components/Banner/Banner.js
barrymcgee: I'd expect anyone using it to pass it that prop to toggle display but if not, it'll display by default.

barrymcgee:
@hatched I assumed that if someone has dismissed it.. they would want it to disappear for the lifetime of their session? Clarify @spencerbygraves ?

spencerbygraves:
I think it makes sense for it to reappear, so the user knows to refresh.

src/components/Banner/Banner.js
```diff
@@ -1,11 +1,37 @@
...
+import React, { useState, useEffect } from "react";
+import classnames from "classnames";
 
 import "./_banner.scss";
 
-export default function Banner({ isActive, children, type }) {
+export default function Banner({ isActive = false, children, type }) {
+  const [bannerClosed, setBannerClosed] = useState(false);
+
+  // Open banner every time banner type changes
```
hatched, line 9: Can you add a test for this plz.

hatched (approved):
Thanks, approved with one small test request.

PR #738, opened 2020-10-30
Refresh the model status data on submitting a command

### LINKED ISSUE #1459: Juju-dashboard on juju 3.1
Hi, I am trying to install juju-dashboard on juju version 3.1. it doesn't work completely!!! Nginx is still blocked!!!
Even the juju-dashboard command gives wrong information:
http://localhost:31666
While I connect with the ip address on port 8080.
Thank's.
Capture d’écran du 2023-04-14 09-22-29
Capture d’écran du 2023-04-14 09-21-12
Capture d’écran du 2023-04-14 09-31-51

hatched:
## Done

- Refresh the model status data on submitting a command in the Juju Web CLI. This will be removed once we go to the new watcher.
- Split the websocket connection out into its own file to stop working against the React paradigm. 
- Correctly parse the controller/model data to connect to the correct controller in a multi-controller environment like JAAS.

## QA

- Bootstrap a new controller with Juju 2.9 (the Juju snap has the 2.9 RC available)
- Build and upload the dashboard to that controller 
  - `yarn`
  - `yarn generate-release-tarball`
  - `juju upgrade-dashboard <path to tarball>` 
- Then log in, the wCLI on the model details page should continue to work as expected.
- Run `juju deploy wordpress` in a model and it should show up in the details page above moments after submitting the command.

## Details

Fixes #1459

### DIFF
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -562,12 +562,44 @@ export const getWSControllerURL = createSelector(
   Returns the controller data in the format of an Object.entries output.
   [wsControllerURL, [data]]
   @param {String} controllerUUID The full controller UUID.
+  @returns {Array} The controller data in the format of an Object.entries output.
 */
 export const getControllerDataByUUID = (controllerUUID) => {
   return createSelector(getControllerData, (controllerData) => {
     if (!controllerData) return null;
-    return Object.entries(controllerData).find(
-      (controller) => controllerUUID === controller[1][0].uuid
-    );
+    const found = Object.entries(controllerData).find((controller) => {
+      // Loop through the sub controllers for each primary controller.
+      // This is typically only seen in JAAS. Outside of JAAS there is only ever
+      // a single sub controller.
+      return controller[1].find(
+        (subController) => controllerUUID === subController.uuid
+      );
+    });
+    return found;
+  });
+};
+
+/**
+  @param {String} controllerUUID The full controller UUID.
+  @returns {Object} The controllerData.
+*/
+export const getModelControllerDataByUUID = (controllerUUID) => {
+  return createSelector(getControllerData, (controllerData) => {
+    if (!controllerData) return null;
+    let modelController = null;
+    Object.entries(controllerData).some((controller) => {
+      // Loop through the sub controllers for each primary controller.
+      // This is typically only seen in JAAS. Outside of JAAS there is only ever
+      // a single sub controller.
+      const modelControllerData = controller[1].find(
+        (subController) => controllerUUID === subController.uuid
+      );
+      if (modelControllerData) {
+        modelController = modelControllerData;
+        return true;
+      }
+      return false;
+    });
+    return modelController;
   });
 };
--- a/src/components/WebCLI/WebCLI.js
+++ b/src/components/WebCLI/WebCLI.js
@@ -1,25 +1,20 @@
-import React, { useCallback, useEffect, useRef, useState } from "react";
+import React, { useEffect, useMemo, useRef, useState } from "react";
 
 import useAnalytics from "../../hooks/useAnalytics";
 
 import WebCLIOutput from "./Output.js";
 
+import Connection from "./connection";
+
 import "./_webcli.scss";
 
 const DEFAULT_PLACEHOLDER = "enter command";
-
-const generateAddress = (controllerWSHost, modelUUID, protocol = "wss") => {
-  if (!controllerWSHost || !modelUUID) {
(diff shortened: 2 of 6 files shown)


src/app/selectors.js
```diff
@@ -562,12 +562,44 @@ export const getWSControllerURL = createSelector(
...
   return createSelector(getControllerData, (controllerData) => {
     if (!controllerData) return null;
-    return Object.entries(controllerData).find(
-      (controller) => controllerUUID === controller[1][0].uuid
-    );
+    const found = Object.entries(controllerData).find((controller) => {
+      // Loop through the sub controllers for each primary controller.
+      // This is typically only seen in JAAS. Outside of JAAS there is only ever
+      // a single sub controller.
+      return controller[1].find(
```
barrymcgee, line 574: If the purpose of this to is to return the first controller, even if there is only one - why isn't this array zero based?

src/components/WebCLI/WebCLI.js
```diff
@@ -68,56 +66,17 @@ const WebCLI = ({ controllerWSHost, credentials, modelUUID, protocol }) => {
       action: "WebCLI command sent",
     });
     inputRef.current.value = ""; // Clear the input after sending the message.
+    setTimeout(() => {
+      // Delay the refresh long enough so that the Juju controller has time to
+      // respond before we request the updated status.
+      refreshModel();
+    }, 500);
```
barrymcgee, line 73: What happens if it takes longer than 500ms to respond?

src/components/WebCLI/WebCLI.js
```diff
@@ -138,7 +97,7 @@ const WebCLI = ({ controllerWSHost, credentials, modelUUID, protocol }) => {
             type="text"
             name="command"
             ref={inputRef}
-            placeholder={placeholder}
+            placeholder="enter command"
```
barrymcgee, line 100: Should enter not be capitalised?

src/components/WebCLI/__snapshots__/WebCLI.test.js.snap
```diff
@@ -86,7 +89,7 @@ exports[`WebCLI renders correctly 1`] = `
           autoCorrect="off"
           className="webcli__input-input"
           name="command"
-          placeholder="no web cli backend available"
+          placeholder="enter command"
```
barrymcgee, line 92: Capitalise?

src/components/WebCLI/connection.js
```diff
@@ -0,0 +1,69 @@
...
+
+  disconnect() {
+    this._ws.close();
+  }
+
+  _handleMessage(e) {
+    try {
+      const data = JSON.parse(e.data);
+      if (data["redirect-to"]) {
+        // This is a JAAS ccontroller and we need to instead
```
barrymcgee, line 32:
```suggestion
        // This is a JAAS controller and we need to instead
```

src/components/WebCLI/connection.js
```diff
@@ -0,0 +1,69 @@
...
+        return;
+      }
+      if (!data.output) {
+        // This is the first message, an empty object and a newline.
+        return;
+      }
+      this._pushToMessageBuffer(`\n${data?.output[0]}`);
+    } catch (e) {
+      console.log(e);
+      // XXX handle the invalid data response
```
barrymcgee, line 50: Why is this handler not part of this PR?

barrymcgee (changes requested):
Some notes in code.

src/app/selectors.js
hatched: It is the result from an `Object.entries` in which the `0` index is the key and the `1` index is the value. In this case, the list of controllers.

src/components/WebCLI/WebCLI.js
hatched: Then it doesn't get reflected until the next poll, which happens every 30s. This is only a temporary fix until we implement the watcher.

src/components/WebCLI/WebCLI.js
hatched: Not as per the design.  I tried it as "Enter command" and it didn't look right in the prompt.

src/components/WebCLI/connection.js
hatched: I haven't yet found an invalid data response to test with :)

hatched: But we also haven't figured out how to do the error handling on this connection.

PR #748, opened 2020-11-17
Fix 404 pages

### LINKED ISSUE #745: Our 404 page is unreachable
We should be able to tell the user if they have hit a genuine 404 - right now, we seem to use hang with whitespace.

barrymcgee:
## Done

- Fix 404 pages

## QA

- Pull code
- Open http://0.0.0.0:8036/fdlkfjdkjf
- Verify 404 page is shown
- Click all other routes and ensure they load as expected

## Details

Fixes #745


anthonydillon:
Worth getting a design review as it's a new view.

src/components/Routes/Routes.js
```diff
@@ -35,7 +40,20 @@ export function Routes() {
...
+        </Login>
+      </Route>
     );
   });
+
+  return (
+    <Switch>
+      {routes}
+      <Route>
+        <NotFound path="*" />
```
hatched, line 55: the `path` prop here is not being used, did you mean to put this on the `Route`?

src/pages/NotFound/NotFound.test.js
```diff
@@ -1,10 +1,24 @@
...
 describe("NotFound page", () => {
-  it("renders without crashing and matches snapshot", () => {
-    const wrapper = shallow(<NotFound />);
-    expect(wrapper).toMatchSnapshot();
+  it("should display when unknown route is accessed", () => {
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter initialEntries={["/foobar11"]}>
+          <NotFound />
```
hatched, line 18: This test isn't actually testing anything, it's just rendering the `NotFound` component. You'll want to use the `TestRoute` component. You can see it in action in the `InfoPanel` or `ModelDetail` tests.

src/components/Routes/Routes.js
```diff
@@ -35,7 +40,20 @@ export function Routes() {
       );
     }
     return (
-      <Route key={path[0]} path={path[0]} exact component={path[1].component} />
+      <Route key={path[0]} path={path[0]} exact>
```
hatched, line 43: This is now wrapping each route in a `Login` component, not all routes. I'm curious why that change was made? I'm not totally sure why this structure change solves the problem, can you elaborate?

hatched (changes requested):

src/components/Routes/Routes.js
barrymcgee: I meant to remove it, now removed.

src/components/Routes/Routes.js
barrymcgee: A switch statement does not work for nested routes, so I needed to move the <Login> inside each route.

hatched: Ahh makes sense, thanks

long-chung:
Good to go from design. 
*Will look at a redesign of the 404 page at a later date as per conversation we had in standup on 17.11.2020

src/pages/NotFound/NotFound.test.js
```diff
@@ -1,10 +1,24 @@
...
+  it("should display when unknown route is accessed", () => {
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter initialEntries={["/foobar11"]}>
+          <Routes />
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("NotFound").length).toBe(1);
```
hatched, line 22: Could you also add a check to make sure nothing else was rendered by `Routes` that wasn't supposed to be?

hatched (approved):

PR #754, opened 2020-11-18
Switch to v2 of the jujulib

### LINKED ISSUE #653: Convert to JujuLib 2.0
When the new version of jujulib stabilizes we'll need to integrate it into the dashboard. This should be fairly straightforward as the UI layer doesn't communicate directly with the API so we can shim any large issues and update them incrementally. There are still a number of unknowns here so I'm going to size it quite large.

Doing this change enables access to the new watcher api's and shell access.

hatched:
## Done

Switch to v2 of jujulib

## QA

A full top to bottom QA is required across all substrates.
See [QA Instructions](https://github.com/canonical-web-and-design/jaas-dashboard#qaing-the-dashboard)

## Details

fixes #653


package.json
```diff
@@ -16,18 +16,18 @@
...
     "generate-version-file": "./scripts/generate-version-file",
     "lint": "yarn run eslint-check && yarn run prettier-check && yarn run stylelint-check",
     "prettier-check": "prettier --check 'src/**/*'",
     "serve": "yarn run start",
-    "start": "craco start",
+    "start": "react-scripts start",
     "stylelint-check": "stylelint 'src/**/*.scss'",
     "stylelint-fix": "stylelint --fix 'src/**/*.scss'",
-    "test": "craco test"
+    "test": "react-scripts test --transformIgnorePatterns 'node_modules/(?!@canonical)/jujulib'"
```
hatched, line 30: While the app runs and builds properly, the tests fail if we do not babelize jujulib to remove the `import` statements. This has the unfortunate side effect of making circleci run out of memory so I'll have to dig into this in the morning.

package.json
```diff
@@ -51,10 +51,9 @@
   },
   "dependencies": {
     "@canonical/jaaslib": "0.6.0",
-    "@canonical/jujulib": "1.3.2",
+    "@canonical/jujulib": "^2.0.0-beta.1",
```
hatched, line 54: Once Juju 2.9 gets closer to official release I'll create the actual `2.0.0` release of jujulib with the most recent schema and update this.

package.json
```diff
@@ -74,7 +73,7 @@
     "react-ga": "3.2.1",
     "react-redux": "7.2.2",
     "react-router-dom": "5.2.0",
-    "react-scripts": "3.4.4",
+    "react-scripts": "4.0.0",
```
hatched, line 76: This was necessary to do here because of a `nanoid` issue in cra < 4.0.

src/app/actions.js
```diff
@@ -166,7 +166,11 @@ export function logOut(store) {
       // again.
       dispatch(connectAndStartPolling(store, bakery));
     }
-  };
+  }
+  // Define a name that won't be munged by the minifier to check
+  // against in the check-auth middleware.
+  logOut.NAME = "logOut";
```
hatched, line 172: This is a workaround for CRA's munging of function names in its minifier. This was required so that we could get rid of `craco` while also keeping authentication checks on dispatched thunks.

src/app/check-auth.js
```diff
@@ -53,21 +51,22 @@ const checkLoggedIn = (state, wsControllerURL) => {
       wsControllerURL: The full controller websocket url that the controller
         is stored under in redux in order to determine it's logged in status.
 */
+// eslint-disable-next-line import/no-anonymous-default-export
```
hatched, line 54: Adding a name here made the middleware go funky so I skipped the lint. Something to look into later.

src/app/jimm-facade.js
```diff
@@ -1,7 +1,4 @@
-const {
-  autoBind,
-  createAsyncHandler,
-} = require("@canonical/jujulib/api/transform.js");
+const { autoBind } = require("@canonical/jujulib/dist/api/utils.js");
```
hatched, line 1: The jimm facade is the only one still manually generated. This is likely changing in the future but for now....here it is.

src/app/utils.js
```diff
@@ -51,13 +51,14 @@ export const getModelStatusGroupData = (model) => {
       messages.push(app.status.info);
       return;
     }
-    Object.keys(app.units).forEach((unitId) => {
-      const unit = app.units[unitId];
+    const units = app.units || {}; // subordinates do not have units.
```
hatched, line 54: You'll see this throughout. The old jujulib returned default values for returned datasets. This was incorrect and has been remedied in the new jujulib but this means we have to handle it in the application where necessary.

src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -35,9 +35,7 @@ describe("Info Panel", () => {
     const store = mockStore(dataDump);
     const wrapper = mount(
       <Provider store={store}>
-        <MemoryRouter
-          initialEntries={["/models/activedev@external/group-test"]}
-        >
+        <MemoryRouter initialEntries={["/models/group-test"]}>
```
hatched, line 38: The mock dataset has been updated so you'll see a number of these changes.

src/index.js
```diff
@@ -77,7 +77,6 @@ function bootstrap() {
     {
       ui: {
         userMenuActive: false,
-        externalNavActive: false,
```
hatched, line 80: This was no longer being used after the application layout update.

src/juju/index.js
```diff
@@ -1,13 +1,12 @@
...
-// mangle.reserved list in craco.config.js.
-import annotations from "@canonical/jujulib/api/facades/annotations-v2";
-import client from "@canonical/jujulib/api/facades/client-v2";
-import cloud from "@canonical/jujulib/api/facades/cloud-v3";
-import controller from "@canonical/jujulib/api/facades/controller-v5";
-import modelManager from "@canonical/jujulib/api/facades/model-manager-v5";
-import pinger from "@canonical/jujulib/api/facades/pinger-v1";
+import { connect, connectAndLogin } from "@canonical/jujulib";
+
+import annotations from "@canonical/jujulib/dist/api/facades/annotations-v2";
```
hatched, line 4: Some of these aren't the latest versions but we can update these in follow-ups if we determine that it's necessary.

src/juju/index.js
```diff
@@ -44,14 +43,15 @@ function generateConnectionOptions(usePinger = false, bakery, onClose) {
     closeCallback: onClose,
     debug: false,
     facades,
+    wsclass: WebSocket,
```
hatched, line 46: The old jujulib defaulted to the window.WebSocket whereas the new one requires one passed in to allow for better testing infrastructure.

src/pages/Controllers/__snapshots__/Controllers.test.js.snap
```diff
@@ -5,35 +5,35 @@ exports[`Controllers table counts models, machines, apps, and units 1`] = `
   role="row"
 >
   <TableCell>
-    admins/1-eu-west-1-aws-jaas
+    JAAS
   </TableCell>
   <TableCell>
-    aws/eu-west-1
+    unknown/unknown
```
hatched, line 11: This is a known bug.

anthonydillon:
## JAAS QA issues
- When filtering the model details tables from status, units, machines, etc the topology rerenders incorrectly
Peek 2020-11-21 10-11

## Juju QA issues
- None

src/app/check-auth.js
barrymcgee: Is this tracked by an issue?

barrymcgee (approved):

hatched:
Thanks all for the reviews and qa's. Remaining issues have been fixed and then issues for the follow-ups created 👍

PR #755, opened 2020-11-19
Replace Search and filter component

barrymcgee:
## Done

Replace Search and filter component using component from @canonical/react-components

## QA

- Pull code
- Click 'Search and filter' component in top right and click through all possible filters available
- Verify the models table filters accordingly
- Type a random word and press enter 
- Ensure only models with that word in their top level attributes appear
- Refresh your page and ensure the filters persist via query strings
- Test over multiple screen resolutions
- Test using a keyboard

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1467

## Screenshots

Screenshot 2020-11-19 at 09 53 43


package.json
```diff
@@ -81,7 +82,7 @@
     "redux-thunk": "2.3.0",
     "reselect": "4.0.0",
     "use-query-params": "1.1.9",
-    "vanilla-framework": "2.19.3"
+    "vanilla-framework": "2.19.2"
```
anthonydillon, line 85: Why roll Vanilla back?

src/app/selectors.js
```diff
@@ -94,7 +94,7 @@ const getFilteredModelData = (filters) =>
   createSelector(
     [getModelData, getControllerData],
     (modelData, controllerData) =>
-      // inject modeldata with controller data
+      // inject modelData with controller data
```
anthonydillon, line 97: I don't think this comment is required I believe it's clear from the code.

src/app/selectors.js
```diff
@@ -372,37 +372,53 @@ const filterModelData = (filters, modelData, controllers) => {
...
+    const owner = data?.info && extractOwnerName(data.info.ownerTag);
+    // Combine all of the above to create string for fuzzy custom search
+    const combinedModelAttributes = `${modelName} ${cloud} ${credential} ${region} ${owner}`;
+
+    const remove = Object.entries(filterSegments).some(
+      ([segment, valuesArr]) => {
+        const values = valuesArr[0];
+        switch (segment) {
+          case "cloud":
+            return !values.includes(cloud);
```
anthonydillon, line 400: Using `indexOf` is [dramatically faster](https://stackoverflow.com/questions/5296268/fastest-way-to-check-a-string-contain-another-substring-in-javascript) particularly in Chrome.

src/hooks/useModelAttributes.js
```diff
@@ -0,0 +1,39 @@
...
+  const credentials = [];
+
+  // Loop the model data and pull out the available filters
+  Object.values(modelData).forEach((model) => {
+    if (!model.info) {
+      return;
+    }
+    // Extract cloud filters
+    const cloudFilter = extractCloudName(model.info.cloudTag);
+    if (!clouds.includes(cloudFilter)) clouds.push(cloudFilter);
```
anthonydillon, line 20: Same here with the `indexOf`

src/app/selectors.js
```diff
@@ -372,37 +372,53 @@ const filterModelData = (filters, modelData, controllers) => {
...
+  Object.entries(filters).forEach((filter) => {
+    if (filter[1].length === 0) return;
+    if (!filterSegments[filter[0]]) {
+      filterSegments[filter[0]] = [];
     }
-    filterSegments[values[0]].push(values[1]);
+    filterSegments[filter[0]].push(filter[1]);
   });
 
   Object.entries(clonedModelData).forEach(([uuid, data]) => {
```
anthonydillon, line 385: This seems like duplicated functionality with useModelAttributes?

src/pages/Models/_models.scss
```diff
@@ -23,6 +24,24 @@
...
+      @media (min-width: $breakpoint-medium) {
+        max-width: 270px;
+        position: absolute;
+        right: 0.75rem;
+        top: 0.3rem;
+        width: 100%;
+      }
+
+      .p-chip {
+        // Overrdie required until upstream fix
```
anthonydillon, line 38:
```suggestion
        // Override required until upstream fix
```

anthonydillon (changes requested):
Few comments in the code.

anthonydillon:
## QA
- When refreshing the page I get 404. For example https://jaas-dashboard-755.demos.haus/models?cloud=google

hatched:
@anthonydillon the demo service doesn't support loading the app on anything but the base path, that's why you're getting a 404 there.

src/app/selectors.js
hatched: Those are old benchmarks (and micro benchmarks at that). It is a very tiny bit slower now but not enough to make a difference as this isn't a hot path.

src/hooks/useModelAttributes.js
hatched: see above.

package.json
```diff
@@ -65,6 +65,7 @@
     "enzyme-to-json": "3.6.1",
     "immer": "7.0.14",
     "mock-socket": "^9.0.3",
+    "nanoid": "3.1.16",
```
hatched, line 68: Is this required? Should react-components not be pulling it in as its own dependency?

package.json
barrymcgee: To match with the version currently on react-components.

package.json
barrymcgee: It should but it currently does not: https://github.com/canonical-web-and-design/react-components/issues/327

When they fix that, we can drop this dep.

anthonydillon:
If you have more then one custom it only picks up one http://localhost:8036/models?custom=algebra&custom=mymodel you probably need to append the one custom query.

anthonydillon:
Should you see all filters if already filtered away? 
image

anthonydillon (approved):
QA issue filed https://github.com/canonical-web-and-design/jaas-dashboard/issues/756

package.json
barrymcgee: I've bumped this back again.

src/app/selectors.js
barrymcgee: `useModelAttributes` takes all available models and pull out the filters that can be applied to them based on the attributes available.. owner, region, etc

This selector takes the filters which have been added to the Search and Filter component and segments them so they can be used to filter out the model data.

long-chung:
Height of search box looks much taller than vanilla default. Not a blocker, but would like to revisit this when the UI updates for the new designs gets implemented.

ziheliu214:
UX and logic (within the same category is 'OR' and cross-category is 'AND') is perfect! @barrymcgee Thanks

PR #776, opened 2020-11-30
Convert Banner to Typescript

barrymcgee:
## Done

Convert Banner to Typescript

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check all of the things pass


src/components/Banner/Banner.tsx
```diff
@@ -1,21 +1,31 @@
...
+type Props = {
+  isActive: boolean;
+  children: any;
+  variant: "positive" | "caution" | "negative";
+} & HTMLProps<HTMLElement>;
+
+export default function Banner({
+  isActive,
+  children,
+  variant,
```
hatched, line 15: `type` was changed to `variant` here and in the tests but not in `layout.js`

src/components/Banner/Banner.tsx
```diff
@@ -1,21 +1,31 @@
...
+}: Props): JSX.Element {
```
squidsoup, line 16: We've always used `JSX.Element`, but you may want to consider using `React.Element` which I believe is considered preferable.

src/components/Banner/Banner.tsx
```diff
@@ -1,21 +1,31 @@
...
   const [bannerClosed, setBannerClosed] = useState(false);
```
hatched, line 17: can you add the type to `bannerClosed` please.

hatched (changes requested):
Good first start but you forgot to update the usage of this component ;)

src/components/Banner/Banner.tsx
barrymcgee: D'oh

src/components/Banner/Banner.tsx
barrymcgee: This is inferred from the initial state value?

Screenshot 2020-12-01 at 11 07 14

barrymcgee: Screenshot 2020-12-01 at 11 24 02

hatched: Interesting, must be a special case that it knows what `useState` does.

hatched: Especially because you can pass a function to `useState` so I wonder if it'll require the type of the return value of the function in that case.

hatched (approved):

PR #778, opened 2020-12-01
Convert Button to Typescript

barrymcgee:
## Done

Convert Button to Typescript

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check tests pass and all is as expected


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { HTMLProps } from "react";
+
+import "./_button.scss";
+
+type Props = {
+  onClick: () => void;
+  children: string;
+} & HTMLProps<HTMLElement>;
```
hatched, line 8: Because we aren't adding any html props to these components we don't need this `& HTMLProps<HTMLElement>` definition here, or in the `Banner` component, can you remove it from both?

```suggestion
};
```

src/components/Button/Button.test.tsx
```diff
@@ -5,7 +5,9 @@ import Button from "./Button";
 
 describe("Button", () => {
   it("should display button text", () => {
-    const wrapper = shallow(<Button>Button text</Button>);
+    const wrapper = shallow(
+      <Button onClick={() => jest.fn()}>Button text</Button>
```
hatched, line 9: Can you add a new test that when you click on the button it executes the onclick handler.

hatched (changes requested):

hatched (approved):

PR #779, opened 2020-12-01
Wrap pages in subtle FadeIn animation

barrymcgee:
## Done

Wrap pages in subtle FadeIn animation

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click through each page

### DIFF
--- a/src/animations/FadeIn.tsx
+++ b/src/animations/FadeIn.tsx
@@ -6,7 +6,7 @@ type Props = {
   isActive: boolean;
 };
 
-export default function FadeUpIn({
+export default function FadeIn({
   children,
   isActive = true,
 }: Props): ReactElement {
@@ -14,10 +14,7 @@ export default function FadeUpIn({
     <>
       {isActive && (
         <AnimatePresence>
-          <motion.div
-            initial={{ opacity: 0, y: 10 }}
-            animate={{ opacity: 1, y: 0 }}
-          >
+          <motion.div initial={{ opacity: 0 }} animate={{ opacity: 1 }}>
             {children}
           </motion.div>
         </AnimatePresence>
--- a/src/pages/Models/Models.js
+++ b/src/pages/Models/Models.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { useSelector } from "react-redux";
+import Spinner from "@canonical/react-components/dist/components/Spinner";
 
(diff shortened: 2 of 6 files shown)


spencerbygraves (approved):
Thanks @barrymcgee, nice touch 👍

src/animations/FadeIn.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
```
hatched, line 1: I guess this is no longer required so you can just import `ReactElement` as in the `Button` branch.

src/animations/FadeIn.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { motion, AnimatePresence } from "framer-motion";
+
+type Props = {
+  children: any;
```
hatched, line 5: we should try to avoid `any` whenever possible....This technically should be a `ReactElement` yes?

src/animations/FadeUpIn.tsx
```diff
@@ -6,7 +6,7 @@ type Props = {
   isActive: boolean;
```
hatched, line 6: See my comment in `FadeIn` for the `children` type of this file as well.

src/animations/FadeIn.tsx
```diff
@@ -0,0 +1,24 @@
...
+  isActive: boolean;
+};
+
+export default function FadeIn({
+  children,
+  isActive,
```
hatched, line 11: Maybe this can default to true?

src/pages/Models/Models.js
```diff
@@ -131,20 +136,28 @@ export default function Models() {
...
+                    { label: "Running", count: running },
+                  ],
+                }}
+              />
+              <ModelTableList groupedBy={groupModelsBy} filters={filters} />
+            </div>
+          </div>
+        </FadeIn>
+      ) : (
+        <div className="model-details__loading">
```
hatched, line 157: The spinner could also fade in too?

hatched (changes requested):

src/pages/Models/Models.js
barrymcgee: Tried this.. visual fade in for loading is very difficult to see, not worth added complexity in code.

hatched (approved):

PR #805, opened 2020-12-09
Add Remote Apps and Offers to the model details page

barrymcgee:
## Done 

- Added empty states for app and units on model details page
- Added offers and remote apps table to 'Apps' tab in model details
- Added offers and remote apps panels (disabled)

## QA

- You will need to fire up models with both offers and offerings
- You can roll back the last commit to enable panels
- Also fire up model with no applications to test empty states
- All of the above will need to be done in your own bootstrapped controller as JIMM does not yet support CMR

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1509,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1521,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1508,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1513


src/components/InfoPanel/InfoPanel.js
```diff
@@ -36,6 +36,7 @@ const InfoPanel = () => {
   const { 0: modelName } = useParams();
   const [showExpandedTopology, setShowExpandedTopology] = useState(false);
   const modelStatusData = useModelStatus();
+  const applicationsCount = Object.entries(modelStatusData.applications).length;
```
hatched, line 39: it's possible that `modelStatusData.applications` is not defined. We should probably have the reducer set sane defaults for the top level keys like this.
```suggestion
  const applicationsCount = Object.entries(modelStatusData.applications || {}).length;
```

src/pages/Models/Details/_model-details.scss
```diff
@@ -209,9 +209,11 @@
     }
   }
 
+  // .model-details__offers, // DISABLED PANEL
+  // .model-details__remote-apps, // DISABLED PANEL
```
hatched, line 213: Do these really need to be commented out if the panels are disabled?

src/components/panels/RemoteAppsPanel/RemoteAppsPanel.tsx
```diff
@@ -0,0 +1,44 @@
...
+  entity: string;
+};
+
+export default function RemoteAppsPanel({
+  entity: appId,
+}: Props): ReactElement {
+  // Get model status info
+  const modelStatusData = useModelStatus();
+
+  const remoteApp = modelStatusData["remote-applications"][appId];
```
hatched, line 18: I believe that `remote-applications` can also be undefined so this should be a bit more resilient around if this appId can not be found.

src/components/panels/RemoteAppsPanel/RemoteAppsPanel.tsx
```diff
@@ -0,0 +1,44 @@
...
+  interface tableData {
```
hatched, line 19: Typically you want to collect all of the type definitions in one spot so lets move this to the top with the `Props` type.

src/components/panels/RemoteAppsPanel/RemoteAppsPanel.tsx
```diff
@@ -0,0 +1,44 @@
...
+    { th: "Store", td: "-" },
+    { th: "Application", td: remoteApp["offer-url"] },
+    { th: "Status", td: remoteApp.status.status },
+    { th: "Endpoint", td: "-" },
+    { th: "Interface", td: "-" },
+  ];
+
+  return (
+    <div className="panel">
+      <span className="p-muted-heading">flannel:etcd</span>
```
hatched, line 37: Not a huge issue as this panel is hidden but this will need to be updated.

src/components/panels/shared.tsx
```diff
@@ -0,0 +1,15 @@
+type data = {
```
hatched, line 1: You've been switching between the use of `type` and `interface` - while they are very similar we should probably just stick with one....my pick would be `interface` unless there is a reason to use `type`.

src/components/panels/OffersPanel/OffersPanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const modelStatusData = useModelStatus();
+  const offers = modelStatusData?.offers[application];
+
+  const [accordionPanelId, setAccordionPanelId] = useState("");
+
+  const handleAccordionToggle = (id: string) => {
+    setAccordionPanelId(id === accordionPanelId ? "" : id);
+  };
+
+  interface tableData {
```
hatched, line 24: Lets collect the type declarations together at the top.

src/components/panels/OffersPanel/OffersPanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const tab3: tableData[] = [
+    { th: "User", td: "-" },
+    { th: "display-name", td: "-" },
+    { th: "access", td: "-" },
+  ];
+
+  return (
+    <div className="panel">
+      <span className="p-muted-heading">mysql:db</span>
```
hatched, line 59: There are a number of hardcoded values here, we'll have to be sure to go back and update them all before this goes live.

src/components/panels/shared.tsx
hatched: After doing some more investigations, `type` might be the one to stick with unless we need the features from `interface` :)

src/pages/Models/Details/_model-details.scss
barrymcgee: Yes, otherwise the hover state still exists but clicking does nothing.

hatched (approved):

PR #815, opened 2021-01-09
Create configuration panel UI

hatched:
## Done

Create the UI for fetching and displaying the configuration values.

Note: There are a few features yet to be implemented, see the associated Epic.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Load the `charmed-kubernetes` model and click on `containerd`. This is a good config because it has both a string and boolean type.
- [ ] When viewing the configuration pane it should show a spinner while it's loading.
- [ ] It should render both textarea and boolean config UI's
- [ ] When changing a configuration value a `use default` UI element should appear and a `reset all fields` button should appear at the top
- [ ] When focusing on an input or clicking around a config option it should show the description on the right and highlight the input.
- [ ] Clicking the `use default` button should change the field to the default 
- [ ] Clicking the `reset all` should reset all the fields to the default.
- [ ] After modifying an input and saving, it should actually save (you can check this by closing and re-opening the panel)
- [ ] Play around with modifying and restoring values, to see if it all works as expected.
- [ ] Hitting cancel should close the panel.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1567
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1571
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1566
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1569
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1568
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1573
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1583
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1572
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1570

## Screenshots

Screen Shot 2021-01-12 at 3 38 29 PM


anthonydillon:
@hatched after a quick play I would suggest this as the best code and UX for this pattern https://codepen.io/anthonydillon/pen/oNzPNJq.

The use of a placeholder to indicate the default value is used is not normal usage of a placeholder and therefore will lead to confusing that the input is empty but the default is the suggested value. Therefore I used the "Use default" element to indicate the state of the input. 

Placing the "Use default" action on the right side can be problematic as that is where a lot of plugins put their actions too. Such as lastpass. Therefore I moved it to the right along side the label.

I think the wording "Restore default" is clearer.

clagom:
Can the field show "e.g." and then the default value? So that is clear that is an example for an empty field and not an active input.
**channel**
[ e.g. 3.4/stable ]
And probably being a placeholder it should disappear on focus.

@ziheliu214

clagom:
When the panel is open, if you click on another app the panel closes down. Shouldn't it refresh and open the new selected app @hatched?

hatched:
> Can the field show "e.g." and then the default value? So that is clear that is an example for an empty field and not an active input.

The issue is that these values aren't an example, that's what it's actually set to. I will still use the greyed out colour but make it actual text and not a placeholder. 

> When the panel is open, if you click on another app the panel closes down. Shouldn't it refresh and open the new selected app @hatched?

There are many bugs and incomplete features, that's why it's a WIP :)

long-chung:
- [ ] **BUG** Use default doesn't work on radios
- [ ] Missing placeholder images (Assets can be DL from Zeplin)
- [ ] Paragraph and spacing issues with the description. Is this raw data coming from the config?
- [ ] **BUG** When you resize the browser width the radio values float beneath the radio button. If you resize it down to mobile and then adjust the width it fixes itself.
Screenshot 2021-01-15 at 09 24 23

- [ ] Reset button overlaps app name and close CTA at a certain BP
Screenshot 2021-01-15 at 09 24 00

- [ ] At < 1024 can we use mobile layout

barrymcgee:
I cannot click 'false' in this radio

Screenshot 2021-01-15 at 09 50 39

barrymcgee:
Icon and app name should align

Screenshot 2021-01-15 at 09 51 27

barrymcgee:
Link should be clickable

Screenshot 2021-01-15 at 09 53 01

barrymcgee:
- [ ] It's possible for loading state to resolve but no form appear in left section of panel
- [ ] The entire panel should show a loading state until form options appear other wise supplementary text on right is confusing

src/components/panels/ConfigPanel/BooleanConfig.tsx
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+      if (falseRef?.current) {
+        falseRef.current.checked = true;
+      }
+    }
+    setShowUseDefault(false);
+  }
+
+  return (
+    // XXX How to tell aria to ignore the click but not the element?
```
barrymcgee, line 66: This is a smell, the click handler should be on the input, not the input container.

src/components/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+              className={classnames(
+                "p-button--positive config-panel__save-button",
+                {
+                  "is-active": savingConfig,
+                }
+              )}
+              onClick={handleSubmit}
+            >
+              {!savingConfig ? (
+                "Save & apply"
```
barrymcgee, line 188:
```suggestion
                "Save &amp; apply"
```

src/components/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -0,0 +1,251 @@
...
+              ) : (
+                <>
+                  <i className="p-icon--spinner u-animation--spin is-light"></i>
+                  <span>Saving...</span>
```
barrymcgee, line 192:
```suggestion
                  <span>Saving&hellip;</span>
```

src/components/panels/ConfigPanel/TextAreaConfig.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  let inputValue = config.default;
+  if (isSet(localNewValue) && isSet(config.newValue)) {
+    inputValue = localNewValue;
+  } else if (config.default !== config.value) {
+    inputValue = config.value;
+  }
+
+  useEffect(() => {
+    if (selectedConfig?.name === config.name) {
```
barrymcgee, line 29: Couldn't this be simplified to;

```
setInputFocused(selectedConfig?.name === config.name);
```

src/components/panels/LocalAppsPanel/LocalAppsPanel.js
```diff
@@ -19,86 +19,103 @@ import {
...
+import "./_local-apps-panel.scss";
+
+// Generate panel header for given entity
+const generateAppPanelHeader = (app, title, showConfig) => {
+  return (
+    <div className="panel-header">
+      {app && (
+        <div className="row">
+          <div className="col-4">
+            <div>{title}</div>
```
barrymcgee, line 34: This should probably be a heading element.

src/components/panels/OffersPanel/OffersPanel.tsx
```diff
@@ -17,7 +17,7 @@ type tableData = {
 export default function OffersPanel({
   entity: application,
 }: Props): ReactElement {
-  const modelStatusData = useModelStatus();
+  const modelStatusData: any = useModelStatus();
```
barrymcgee, line 20: Aren't we avoiding `any` types?

barrymcgee (changes requested):

barrymcgee:
On small screens, the supplementary panel becomes useless as it sits out of sight below the form.

barrymcgee:
- [ ] When I X the config panel, I'd expect to return to the app panel, not have both panels dismissed.

src/app/utils.js
```diff
@@ -320,3 +320,5 @@ export const filterModelStatusDataByApp = (modelStatusData, appName) => {
 
   return modelStatusData;
 };
+
+export const isSet = (val) => val || val === "" || val === 0;
```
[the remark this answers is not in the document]
hatched: this was intentionally done like this to be clear that an empty string and a 0 value were acceptable values.

src/components/SlidePanel/SlidePanel.js
```diff
@@ -1,10 +1,23 @@
...
 
 import useEventListener from "hooks/useEventListener";
 
 import "./_slide-panel.scss";
 
+function generatePanelContent(isLoading, children, onClose) {
+  if (isLoading) {
+    return <Spinner />;
+  }
+  if (children && !Array.isArray(children)) {
```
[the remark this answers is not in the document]
hatched: Sure thing.

It's possible to pass a single child to a slide panel or a list of them but I only wanted to pass an `onClose` handler down to the children in the event it was just a single child. It's a limitation to try and keep the usage of the component cleaner when using this approach but we can change it later by removing this check and cloning all children.

```jsx
<SlidePanel>
  <Config/>
</SlidePanel>

<SlidePanel>
  <p>text</p>
  <p>text</p>
  <p>text</p>
</SlidePanel>
```

src/components/panels/ConfigPanel/BooleanConfig.tsx
hatched: We need the click handler on the container as well so that the user can click on the container to highlight the field and show the description on the right. This increases the click/tap area but also works really nice for the radio buttons. I was thinking of using a label instead but I didn't feel it right to try and jam a bunch of extraneous block level elements into the label. 

I'm open to alternative approaches....or an aria fix 👍

src/components/panels/ConfigPanel/TextAreaConfig.tsx
hatched: Yep, updated

src/components/panels/LocalAppsPanel/LocalAppsPanel.js
hatched: The `title` prop contains a number of components as it reuses the heading renderer from the other panel headings.

src/components/panels/OffersPanel/OffersPanel.tsx
hatched: yep, I only added this so that it would compile (it required it with my version of TS for some reason) but I've now updated this to use a custom `TSFixMe` type.

hatched:
Hi guys, I've addressed all the comments where appropriate and targeted this PR to a feature branch as discussed so that it can land and be iterated on outside of master.

Please take another look and if there are any large outstanding issues that aren't already covered in another issue in the epic please create one and +1 this so it can land and it can be iterated on with much smaller PR's ;)

long-chung:
Some major and minor bugs listed below. Bugs that have are major should be fixed in this release, all others can go in to a smaller PR
- [ ] (Major) Save & apply typo
- [ ] (Major) When you click save the buttons disappear and can no longer be accessed
- [ ] Save & apply CTA not aligned to the rest of the form fields
- [ ] BUG Reset CTA is still overlapping close button on smaller widths
- [ ] BUG Reset CTA overlaps form field on smaller widths

long-chung:
CTAs should always be visible. Default state of "save & apply" cta should be faded out 50% If there are no edits. Cancel cta should always remain visible

long-chung:
@hatched i'm assuming an app without any config parameters are landing on a different pr?

hatched:
> @hatched i'm assuming an app without any config parameters are landing on a different pr?

Yep! In a follow-up

PR #831, opened 2021-01-20
Highlight changed fields

hatched:
## Done

When a field value has changed from the value Juju currently has it will be highlighted grey until it has been saved.
There were a lot of additional changes to clean up issues with state in this PR to enable the primary task.

## QA

- Run the demo
- Change a field value and then click to another field, it should highlight grey.
- Save that value and it'll go back to white.
- Play around setting and resetting values to ensure the state changes maintain functionality.

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1584

## Screenshots

Screen Shot 2021-01-20 at 1 17 04 PM

### DIFF
--- a/src/components/panels/ConfigPanel/BooleanConfig.tsx
+++ b/src/components/panels/ConfigPanel/BooleanConfig.tsx
@@ -12,6 +12,7 @@ export default function BooleanConfig({
   setNewValue,
 }: ConfigProps): ReactElement {
   const [inputFocused, setInputFocused] = useState(false);
+  const [inputChanged, setInputChanged] = useState(false);
   const [showUseDefault, setShowUseDefault] = useState(
     config.value !== config.default
   );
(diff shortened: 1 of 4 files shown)


src/components/panels/ConfigPanel/ConfigPanel.test.tsx
```diff
@@ -2,23 +2,26 @@ import { mount } from "enzyme";
...
+  getApplicationConfig: jest.fn(),
 }));
 
-const mockStore = configureStore([]);
-
 describe("ConfigPanel", () => {
+  beforeEach(() => jest.resetModules());
+
   it("displays a message if the app has no config", async () => {
+    (getApplicationConfig as jest.Mock).mockImplementation(() => {
```
hatched, line 22: This is so that it's still type safe but doesn't throw an error from TypeScript

long-chung:
For some reason we have different blue hovers in design and live. 
Jeff could you change the blue highlight to match whats live and i'll do the same for designs.
Please change it to ##E7F9FF.
Going to +1 this as its not a blocker

anthonydillon:
I know it's annoying but could you look at the small screen views. I think a few lines of styling would improve it a great deal.

spencerbygraves:
>For some reason we have different blue hovers in design and live.

@long-chung are you referring to the different blue colour in the table? It looks like we need to update that to the newer blue.

image

long-chung:
Created a new issue to update our hover blue to the new blue #DDEDFE
[Ticket 1597](https://app.zenhub.com/workspaces/juju-squad-593fe9c0b747921fa7ecd561/board?repos=94216425)

barrymcgee (approved):

PR #836, opened 2021-01-22
Revert function typing in header

barrymcgee:
## Done

Revert function typing in header

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,10 +1,12 @@
+import { ReactElement } from "react";
+
 import "./_header.scss";
 
 type Props = {
   children: JSX.Element;
 };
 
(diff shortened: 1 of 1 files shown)


src/components/Header/Header.tsx
```diff
@@ -1,12 +1,10 @@
-import { FunctionComponent } from "react";
-
 import "./_header.scss";
 
 type Props = {
   children: JSX.Element;
 };
 
-const Header: FunctionComponent<Props> = ({ children }) => {
+const Header = ({ children }: Props) => {
```
hatched, line 7: Can you add a return type here too plz.

hatched (changes requested):

hatched (approved):

PR #837, opened 2021-01-22
Convert InfoPanel to Typescript

barrymcgee:
## Done

Convert InfoPanel to Typescript

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Convert InfoPanel to Typescript

### DIFF
--- a/src/components/InfoPanel/InfoPanel.tsx
+++ b/src/components/InfoPanel/InfoPanel.tsx
@@ -31,15 +31,7 @@ const infoPanelDimensions = () => {
 };
 
 const InfoPanel = () => {
-  const params = useParams();
-  let modelName;
-  if (Object.entries(params).length) {
-    Object.values(params).forEach((param, i) => {
-      i === 0 ? (modelName = param) : (modelName = "");
-    });
-  } else {
-    modelName = "";
-  }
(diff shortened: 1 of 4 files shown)


src/components/InfoPanel/InfoPanel.tsx
```diff
@@ -31,7 +31,16 @@ const infoPanelDimensions = () => {
 };
 
 const InfoPanel = () => {
-  const { 0: modelName } = useParams();
+  const params = useParams();
```
hatched, line 34: Oooo that's a lot, does something like this satisfy the compiler?

```typescript
let modelName = "";
{ 0: modelName } = useParams();
```

src/components/InfoPanel/InfoPanel.tsx
```diff
@@ -31,7 +31,8 @@ const infoPanelDimensions = () => {
 };
 
 const InfoPanel = () => {
-  const { 0: modelName } = useParams();
+  const { 0: modelName = "" } = useParams<{ 0: string }>();
```
barrymcgee, line 34: 👍🏻

hatched (approved):
Thanks for this, I think going forward we'll want to add some more robust typing instead of relying on inference but for now this will be fine.

PR #843, opened 2021-01-25
Add UI for viewing and editing charm configuration

hatched:
## Done

Add a UI for viewing and editing charm configuration.

## QA

Note: each PR from the feature branch has already been reviewed and qa'd but this is for all the marbles!

- View the demo link.
- Navigate to a model and view the configuration for an application.
- See if you can break it.
- Try it out on mobile/mobile views

## Screenshots

Screen Shot 2021-01-25 at 12 54 35 PM

PR #844, opened 2021-01-25
Add animations to the configuration UI

hatched:
## Done

Add animations and fades to give the config some life.

Blocked on #843 landing, once it's landed then this can be reviewed and qa'd

## QA

- Use the demo link
- Play around with the config and see the animations work, the list of the addressed items are in the issue.

## Details

https://github.com/canonical-web-and-design/juju-squad/issues/1601

### DIFF
--- a/src/components/panels/ConfigPanel/BooleanConfig.tsx
+++ b/src/components/panels/ConfigPanel/BooleanConfig.tsx
@@ -88,9 +88,12 @@ export default function BooleanConfig({
         {config.name}
       </h5>
       <button
-        className={classnames("u-float-right p-button--base", {
-          "u-hide": !showUseDefault,
-        })}
+        className={classnames(
+          "u-float-right p-button--base config-panel__hide-button",
+          {
+            "config-panel__show-button": showUseDefault,
+          }
+        )}
         onClick={resetToDefault}
       >
         use default
--- a/src/components/panels/ConfigPanel/BooleanConfig.tsx
+++ b/src/components/panels/ConfigPanel/BooleanConfig.tsx
@@ -19,6 +19,8 @@ export default function BooleanConfig({
   );
   const trueRef = useRef<HTMLInputElement>(null);
   const falseRef = useRef<HTMLInputElement>(null);
+  const descriptionRef = useRef<HTMLDivElement>(null);
+  const [maxDescriptionHeight, setMaxDescriptionHeight] = useState("0px");
 
   let inputValue = config.default;
   if (isSet(config.newValue)) {
@@ -26,6 +28,27 @@ export default function BooleanConfig({
   } else if (config.default !== config.value) {
     inputValue = config.value;
   }
+
+  useEffect(() => {
+    if (descriptionRef.current?.firstChild) {
+      setMaxDescriptionHeight(
+        `${
+          (descriptionRef.current.firstChild as HTMLDivElement).clientHeight
+        }px`
+      );
+    }
+  }, []);
+
+  useEffect(() => {
+    if (!descriptionRef.current) {
+      return;
+    }
+    if (showDescription) {
+      descriptionRef.current.style.maxHeight = maxDescriptionHeight;
+    } else {
+      descriptionRef.current.style.maxHeight = "0px";
+    }
+  }, [showDescription, maxDescriptionHeight]);
 
   useEffect(() => {
     if (selectedConfig?.name === config.name) {
@@ -100,10 +123,13 @@ export default function BooleanConfig({
       </button>
       <div
         className={classnames("config-input--description", {
-          "u-hide": !showDescription,
+          "config-input--description__show": showDescription,
         })}
+        ref={descriptionRef}
       >
-        {config.description}
+        <div className="config-input--description-container">
+          {config.description}
+        </div>
       </div>
       <div className="row">
         <label className=".p-radio--inline col-2">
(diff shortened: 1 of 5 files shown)


spencerbygraves:
Thanks for looking at this @hatched, a couple of things I noticed...

- For consistency, are we using the same animation values as elsewhere in the dashboard? For example, the fade on the table row hover blue looks to be slightly different to the config blue
- When I focus a field, only the first description has a fade in. When I focus the others there is no fade in/out

Perhaps unrelated to this PR...

- When I select an app, sometimes the reset button and use default message is already there, others not until I focus a field

image

- The reset all button doesn't work after I've edited some values

anthonydillon:
@hatched can you adjust the motion duration. It seems a little slow leading to a laggy feel. Maybe try halving the default. We have some standard durations here: https://vanillaframework.io/docs/settings/animation-settings

hatched:
These animations are all using the defined Vanilla speeds, they are all set to `brisk`. It's easy to update to another but this was the speed that felt good to me :)

> When I select an app, sometimes the reset button and use default message is already there, others not until I focus a field

This is it working correctly. If the configuration value has been changed from the default already then these will already be there when you view the config.

src/components/panels/ConfigPanel/BooleanConfig.tsx
```diff
@@ -85,19 +108,25 @@ export default function BooleanConfig({
...
+          }
+        )}
         onClick={resetToDefault}
       >
         use default
       </button>
       <div
         className={classnames("config-input--description", {
-          "u-hide": !showDescription,
+          "config-input--description__show": showDescription,
```
anthonydillon, line 123: Is this a good case for a [CSS state class](https://canonical-web-and-design.github.io/practices/coding/stylesheets.html#state-classes)

src/components/panels/ConfigPanel/TextAreaConfig.tsx
```diff
@@ -26,6 +28,27 @@ export default function TextAreaConfig({
...
+      );
+    }
+  }, []);
+
+  useEffect(() => {
+    if (!descriptionRef.current) {
+      return;
+    }
+    if (showDescription) {
+      descriptionRef.current.style.maxHeight = maxDescriptionHeight;
```
anthonydillon, line 46: This logic is duplicated

src/components/panels/ConfigPanel/TextAreaConfig.tsx
```diff
@@ -51,6 +74,10 @@ export default function TextAreaConfig({
     setNewValue(config.name, config.default);
   }
 
+  function handleShowDescription() {
+    setShowDescription(!showDescription);
```
anthonydillon, line 78: These set of words are confusing 😂

src/components/panels/ConfigPanel/_config-panel.scss
```diff
@@ -5,8 +5,14 @@
...
   $magic-min-height: calc(100vh - #{$magic-height});
 
-  @media (max-width: $breakpoint-large) {
-    grid-template-columns: 1fr !important;
+  .config-content {
+    @media (max-width: $breakpoint-large) {
+      grid-template-columns: 1fr !important;
+    }
+
+    &.row {
```
anthonydillon, line 13: You removed the `row` class right?

src/components/panels/ConfigPanel/_config-panel.scss
```diff
@@ -129,6 +142,10 @@
     @media screen and (max-width: $breakpoint-x-small) {
       padding-right: 0;
     }
+
+    button {
```
anthonydillon, line 146: Can you target a class here in case someone updates the mark up.

anthonydillon (approved):
LGTM, few comments in the code but no blockers

src/components/panels/ConfigPanel/TextAreaConfig.tsx
hatched: Yeah, actually a lot of it is duplicated between the two config types. I was going to extract them out but I'm going to wait until after the Formik conversion is completed to avoid duplicating work.

src/components/panels/ConfigPanel/TextAreaConfig.tsx
hatched: I know, I feel like I need to add a comment that says `description` just for fun 😆

src/components/panels/ConfigPanel/_config-panel.scss
hatched: It is easy to miss but I removed it from the wrapper and into the `FadeIn` component so this is still required....I'm refactoring how this is all laid out in a follow-up to better work with the confirmation dialog so this will change a bit anyways very soon.

src/components/panels/ConfigPanel/_config-panel.scss
hatched: Done!

src/components/panels/ConfigPanel/BooleanConfig.tsx
hatched: Turns out this was actually supposed to be removed - in the conflict resolution I appear to have lost an entire commit. That stuff is back now and these lines no longer exist 👍 good catch!

hatched:
Thanks for the reviews!

PR #845, opened 2021-01-27
Render the config panel based on the URL.

hatched:
## Done

- Update the url handling so that the config panel is rendered based on the URL and not part of the App panel.
- Fix sticky description bug from animation branch.

## QA

- Pull code
- Run dashboard
- Navigate to the config UI for an application
- Refresh the page, it should end up loading the config panel again.
- Make your browser vertically small, focus a config input and scroll the page, the config description should stick to the header.

## Details

Fixes https://github.com/canonical-web-and-design/jaas-dashboard/pull/845


src/components/SlidePanel/SlidePanel.js
```diff
@@ -38,6 +38,7 @@ function SlidePanel({
     if (
       isActive &&
       !e.target.closest(".slide-panel") &&
+      !e.target.closest('[role="row"]') &&
```
hatched, line 41: Drive-by to fix a bug where if you clicked on a row in the panel it would close the panel. We'll want a more robust system than this otherwise this list will continue to grow.

src/components/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -220,12 +224,13 @@ export default function ConfigPanel({
           </div>
           <div className="config-panel__description col-6">
             {selectedConfig ? (
-              <FadeIn isActive={true}>
-                <div className="config-panel__description-wrapper">
-                  <h4>Configuration Description</h4>
-                  <h5>{selectedConfig.name}</h5>
-                  <pre>{selectedConfig.description}</pre>
-                </div>
+              <FadeIn
```
hatched, line 227: Drive-by to fix an issue where the description was no longer sticky when scrolling.

src/components/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -16,11 +16,13 @@ import boxImage from "static/images/no-config-params.svg";
 import BooleanConfig from "./BooleanConfig";
 import TextAreaConfig from "./TextAreaConfig";
 
+import { generateIconImg } from "../../../pages/Models/Details/generators";
```
barrymcgee, line 19: This should be an absolute import.

```suggestion
import { generateIconImg } from "pages/Models/Details/generators";
```

barrymcgee (changes requested):
Small change requested

barrymcgee (approved):

PR #848, opened 2021-01-28
Check for model before accessing uuid in poller

barrymcgee:
## Done

Check for model before accessing uuid in poller

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code

## Details

Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/789

### DIFF
--- a/src/app/model-poller.js
+++ b/src/app/model-poller.js
@@ -124,7 +124,7 @@ export async function connectAndPollController(
         wsControllerURL: controllerData[0],
       });
       const modelUUIDList = models["user-models"].map(
-        (item) => item.model.uuid
+        (item) => item.model?.uuid
       );
       await fetchAllModelStatuses(
         controllerData[0],


src/app/model-poller.js
```diff
@@ -124,7 +124,7 @@ export async function connectAndPollController(
         wsControllerURL: controllerData[0],
       });
       const modelUUIDList = models["user-models"].map(
-        (item) => item.model.uuid
+        (item) => item.model?.uuid
```
hatched, line 127: Let's chat about this, this technically shouldn't be possible to fail as to have a model it must have an id.

hatched: The issue is that it can't access the uuid of undefined so in this case the model key is undefined. If that's the case then the map will return undefined instead of a uuid so it'll fail down the line.

NOW....the question is why....

https://github.com/juju/js-libjuju/blob/master/api/facades/model-manager-v8.ts#L311

model is not optional in the Juju api but `models['user-models']` must have a record else the map callback wouldn't run.

SO....I think we need to propose a different fix.....something like....
(untested)
```typescript
// For some reason the model data is sometimes missing from 
// the model entry in the user models list so remove those entries 
// before they enter the official model list.
const filteredModels = models['user-models'].filter(item => item.model !== undefined);
if (filteredModels.length !== models['user-models']) {
  // Send error to sentry with the contents of `models`
}
const modelUUIDList = filteredModels.map((item) => item.model.uuid)
```

So that we can get more information as to why this is failing

hatched (changes requested):

PR #855, opened 2021-02-03
Add model counts as chips

barrymcgee:
## Done

- Add model counts as chips
- Switched StatusStrip to TS
- Moved utils.js into it's own folder and started to add tests

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check models and model details views across breakpoints and verify chips look mint

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1632


hatched:
QA Note:
- [ ] Lets not show labels that have 0 quantity as that list could be very large elsewhere...
Screen Shot 2021-02-03 at 4 01 28 PM

- [ ] Hovering and clicking the chips changes their background but they aren't actionable. I know these will eventually be connected to the S&F but maybe for now remove the hover & active state?

src/app/utils/utils.test.js
```diff
@@ -0,0 +1,19 @@
+import { pluralize } from "./utils";
+
+describe("Util", () => {
+  it("pluralize should correctly pluralize a single item", () => {
+    const singleItems = 1;
+    const label = pluralize(singleItems, "item");
+    expect(label).toBe("item");
+  });
+  it("pluralize should correctly pluralize a multiple item", () => {
```
hatched, line 9:
```suggestion
  it("pluralize should correctly pluralize multiple items", () => {
```

src/app/utils/utils.test.js
```diff
@@ -0,0 +1,19 @@
+import { pluralize } from "./utils";
+
+describe("Util", () => {
+  it("pluralize should correctly pluralize a single item", () => {
```
hatched, line 4: Can you group like tests with another describe....so in this case `describe("pluralize", () =>`

src/app/utils/utils.test.js
```diff
@@ -0,0 +1,19 @@
...
+    const singleItems = 1;
+    const label = pluralize(singleItems, "item");
+    expect(label).toBe("item");
+  });
+  it("pluralize should correctly pluralize a multiple item", () => {
+    const multipleItems = 2;
+    const label = pluralize(multipleItems, "item");
+    expect(label).toBe("items");
+  });
+  it("pluralize special cases correctly", () => {
```
hatched, line 14: This test is quite fragile as the list of special cases is internal to the function and cannot be easily tested. Maybe consider moving the special list out of the `pluralize` function and exporting that specials list. You can then loop through that list in the test to assert that it works as expected.

src/components/StatusStrip/StatusStrip.test.tsx
```diff
@@ -0,0 +1,19 @@
+import { mount } from "enzyme";
+
+import StatusStrip from "./StatusStrip";
+
+describe("Status Strip", () => {
```
hatched, line 5: Can you add a test that has entities with a count of 0 as I don't think they should be shown.

src/components/StatusStrip/StatusStrip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  count: number;
+};
+
+type Props = {
+  statusList: {
+    model: Status[];
+  };
+};
+
+export default function StatusStrip({ statusList }: Props) {
```
hatched, line 16: Maybe now a better name for this would be `StatusChipList` now 🤷‍♂️

src/components/StatusStrip/StatusStrip.tsx
```diff
@@ -0,0 +1,68 @@
+import { useState, useEffect, ReactNode } from "react";
+
+import "./_status-strip.scss";
+
+type Status = {
```
hatched, line 5: Lets try to avoid having two types with the same name in a file.

src/components/StatusStrip/StatusStrip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  label: string;
+  count: number;
+};
+
+type Props = {
+  statusList: {
+    model: Status[];
```
hatched, line 12: I don't think that this data structure makes much sense anymore as it no longer is required to be in context of a modal.. maybe something simple like: `type Props = Status[]`

src/components/StatusStrip/StatusStrip.tsx
```diff
@@ -0,0 +1,68 @@
...
+      count = count + incrementBy;
+    }
+
+    Object.entries(statusList).forEach(([groupName, statuses]) => {
+      const statement =
+        statuses &&
+        Object.values(statuses).map((status) => {
+          incrementCount(status.count);
+          return (
+            <div className="p-chip" key={status.label}>
```
hatched, line 46: Should this use the `Chip` component from `react-components` now? It would need some work to support this content but it would help with some of the interaction logic.

src/components/StatusStrip/StatusStrip.tsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+};
+
+export default function StatusStrip({ statusList }: Props) {
+  type Status = {
+    groupName: string;
+    statement: ReactNode;
+    count: number;
+  };
+  const [status, setStatus] = useState<Status>({
```
hatched, line 22: I'm not sure you need all this logic any longer and can just loop through and render the data from the props if my new type suggestion above is ok.

hatched (changes requested):
After all those comments :) Maybe a better idea here would be to scrap the `StatusStrip` component and start fresh with a `ChipGroup` component instead. Lets have a quick chat about this.

hatched:
After a chat we've decided to fix up the smaller issues here and resolve the qa issues then land this. The Chip Group component will be addressed as a follow-up to replace this.

long-chung:
- Looks great. Those tags even help to lift the current dashboard. 
- Lets remove any hover states if they're not doing anything for now

barrymcgee:
@hatched All refactors completed, need review again please.

hatched (approved):

hatched:
The top of the chips is cut off on the model details pages.

model list:
Screen Shot 2021-02-08 at 2 53 20 PM
model details:
Screen Shot 2021-02-08 at 2 53 27 PM

PR #869, opened 2021-02-08
Try and fetch the users user/pass to support environments without an identity provider

hatched:
## Done

Try and fetch the users user/pass to support environments without an identity provider.

#### JAAS
- In the demo build, use the `charmed-kubernetes` model, view an application and modify a config value. It should work as expected.

#### Juju
- Pull code
- Bootstrap a new controller
- Add a model and deploy an application into it.
- Change the config.js to point to your new controller and `yarn start`
- Open http://0.0.0.0:8036/
- View your new model and application, edit the config, it should work as expected.


barrymcgee:
Had trouble QA'in this - let's jump on a call about it.

barrymcgee (approved):

PR #878, opened 2021-02-15
Switch 'Register Controller' panel to use Vanilla panel

barrymcgee:
## Done

Switch 'Register Controller' panel to use Vanilla panels and Framer motion animation

Note: Due to the meandering path of this feature branch, this PR should squash and merge all commits into one before landing on master.

## QA

- Pull code
- Open http://0.0.0.0:8036/controllers
- Click 'Register new controller'
- See "Register new controller" panel work as expected

Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/866,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1629


src/components/RegisterController/RegisterController.js
```diff
@@ -0,0 +1,263 @@
...
+    const dashboardLink = `https://${controllerIP}/dashboard`;
+    return (
+      <a href={dashboardLink} target="_blank" rel="noopener noreferrer">
+        the controller
+      </a>
+    );
+  }
+
+  return (
+    <SlideInOut>
```
hatched, line 66: If the `AnimatePresence` is going to be in the panels then this should probably be in there too as all Panels will want the same animation.

src/pages/Controllers/Controllers.js
```diff
@@ -147,9 +145,7 @@ function Details() {
         <div className="controllers--register">
           <button
             className="p-button--positive"
-            onClick={() => {
-              setShowRegisterAController(!showRegisterAController);
-            }}
+            onClick={() => history.push("/controllers/register")}
```
hatched, line 148: I feel like we should use a querystring for the panels...That way we can open up any panel over any route.

Thoughts?

barrymcgee: Then we can't use React Router for this purpose.

src/components/RegisterController/RegisterController.js
barrymcgee: AnimatePresence must be a direct parent of the children being swapped in/out. Still refactoring this to make the animations work properly.

src/pages/Controllers/Controllers.js
anthonydillon: I don't think we need to complicate the UI with query params to forms as we have moved all details to there own page. I thought we had saved ourselves the pain of handling panels with query strings?

anthonydillon: Ok I see, we are using the URL as a global state store.

anthonydillon:
Are you planning to make the panel and route persist with refreshes?

barrymcgee:
> Are you planning to make the panel and route persist with refreshes?

Yep!

src/components/RegisterController/RegisterController.js
```diff
@@ -0,0 +1,269 @@
+import { useState } from "react";
```
barrymcgee, line 1: This file was lifted and shifted from Controllers.js

src/panels/panels.tsx
```diff
@@ -0,0 +1,20 @@
...
+import "./_panels.scss";
+
+export default function Panels() {
+  const panelQs = useQueryParam("panel", StringParam)[0];
+
+  const generatePanel = () => {
+    switch (panelQs) {
+      case "register-controller":
+        return <RegisterController />;
+      default:
```
barrymcgee, line 15: Other panels will be added here.

src/ui/selectors.ts
```diff
@@ -1,15 +1,6 @@
 import { ReduxState } from "types";
 
 // ----- Exported functions
-
```
barrymcgee, line 4: Removing this comment for consistency as `isConfirmationModalActive ` was added without comment. It's also pretty self-explanatory.

src/app/check-auth.js
```diff
@@ -19,9 +19,7 @@ const actionWhitelist = [
   "CLEAR_CONTROLLER_DATA",
   "CLEAR_MODEL_DATA",
   "STORE_VISIT_URL",
-  "TOGGLE_COLLAPSIBLE_SIDEBAR",
```
barrymcgee, line 22: The actions are unrelated to this changeset but are redundant.

src/components/Aside/Aside.test.tsx
```diff
@@ -0,0 +1,44 @@
+import { shallow } from "enzyme";
+
+import Aside from "./Aside";
+
+describe("Aside", () => {
```
hatched, line 5: Can you add a test with `mount` to make sure that the content is rendered.

src/components/Aside/Aside.tsx
```diff
@@ -0,0 +1,28 @@
+import { ReactElement } from "react";
+import classnames from "classnames";
+
+import "./_aside.scss";
+
+type Props = {
+  children: ReactElement | ReactElement[];
```
hatched, line 7: This should probably be `ReactNode` as it can be pretty much anything.

```suggestion
  children: ReactNode | ReactNode[];
```

src/components/Aside/Aside.tsx
```diff
@@ -0,0 +1,28 @@
...
+  width?: "wide" | "narrow";
+  pinned?: boolean;
+};
+
+export default function Aside({
+  children,
+  width = undefined,
```
hatched, line 14: This would default to `undefined` anyways, even without this default value.

src/components/Aside/Aside.tsx
```diff
@@ -0,0 +1,28 @@
...
+  pinned = false,
+}: Props): ReactElement {
```
hatched, line 16: Sorry we've standardized on `JSX.Element` after some research and discussion (also the other projects use it as well) If you wouldn't mind updating these files with that 👍

src/components/Aside/Aside.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import "./_aside.scss";
+
+type Props = {
+  children: ReactElement | ReactElement[];
+  width?: "wide" | "narrow";
+  pinned?: boolean;
+};
+
+export default function Aside({
```
hatched, line 12: Should this be part of `react-components` as it's part of the vanilla app layout? 🤷‍♂️

src/animations/SlideInOut.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { motion } from "framer-motion";
+
+type Props = {
+  distanceInPx: number;
+  isActive: boolean;
+  children: ReactElement | ReactElement[];
+};
+
+export default function SlideInOut({
+  distanceInPx,
```
hatched, line 11: What is this magic number? :D I think we need some comments here about how to use this.

src/components/RegisterController/RegisterController.js
```diff
@@ -0,0 +1,269 @@
...
+        the controller
+      </a>
+    );
+  }
+
+  const [panelQs, setPanelQs] = useQueryParam("panel", StringParam);
+
+  return (
+    <Aside>
+      <SlideInOut distanceInPx={725}>
```
hatched, line 73: Should the sliding be in the `Aside` instead of the children of the aside? I feel like if I want to put content in a panel I'd just wrap it in the `Aside` and it should take care of the rest.

src/pages/Controllers/Controllers.test.js
```diff
@@ -1,8 +1,9 @@
 import cloneDeep from "clone-deep";
```
hatched, line 1: Can you add a test for this new queryparam setting functionality.

src/components/Routes/Routes.js
```diff
@@ -15,7 +15,8 @@ export const paths = {
   "/": { redirect: "/models" },
   "/models": { component: Models },
   "/models/*": { component: ModelDetails },
-  "/controllers": { component: Controllers },
+  "/controllers/": { component: Controllers },
+  "/controllers/register": { component: Controllers },
```
hatched, line 19: This is no longer used as it's using the query params now.

src/ui/index.ts
```diff
@@ -13,10 +13,10 @@ function uiReducer(
   return immerProduce(state, (draftState: UIState) => {
     switch (action.type) {
       case actionsList.userMenuActive:
-        draftState.userMenuActive = action.payload;
+        draftState.userMenuActive = action.payload as boolean;
```
hatched, line 16: I've created a follow-up issue for myself https://github.com/canonical-web-and-design/juju-squad/issues/1648

hatched (changes requested):
Thanks @barrymcgee this is looking good just some small updates requested below.

When QA'ing however I found that when the panels animate in we get a horizontal scrollbar on the page that changes as the panel slides in. Have you investigated instead animating the width of a wrapper element instead of moving the whole panel?

src/components/Aside/Aside.tsx
barrymcgee: Ok

src/components/Aside/Aside.tsx
barrymcgee: We only need to wrap it here to support Framer animations.

hatched (approved):

PR #879, opened 2021-02-18
Explicity add modelName as route param

barrymcgee:
## Done

This is explicity add `userName` and `modelName` as route params on `/models/*`

This refactor is necessary to support application details pages at `/models/:userName/:modelName/:applicationName`

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code
- Ensure all tests and checks pass


src/app/selectors.js
```diff
@@ -99,6 +99,20 @@ const getFilteredModelData = (filters) =>
 
 // ---- Utility functions
 
+/**
+  Some models can be namespaced by user so we must validate the url params
+  before confirming corrrect model name
+*/
+export const validateModelNameFromURL = (param1, param2) => {
```
hatched, line 106: Instead of doing this can we just make the username required as part of the url?

IIRC there is only a single place where we generate links in the application to models and that's in the model list tables. So by modifying [generateModelDetailsLink](https://github.com/canonical-web-and-design/jaas-dashboard/blob/master/src/components/ModelTableList/shared.js#L13) by removing the first conditional it'll always return links with the usernames in them. In doing so this function isn't necessary.

hatched (changes requested):

hatched (approved):

hatched:
@barrymcgee I've made the requested changes to this including converting some components to TS but we no longer have to go back and revisit that routing change.

PR #883, opened 2021-02-23
Added analytics pings for opening and saving charm config.

hatched:
## Done

In order to determine if this feature is being used this branch adds a couple surface level analytics pings.
- when the config is opened.
- when a config value is saved.

## QA

- Pull code.
- `yarn build`
- Using your server of choice `cd build && http-server .`
- Open the devtools to the network pane.
- Find a charm and open the config, it should send an analytics ping.
- Save a config value and it should send an analytics ping.


src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -88,6 +92,13 @@ export default function ConfigPanel({
     );
   }, [appName, modelUUID, reduxStore]);
 
+  useEffect(() => {
+    sendAnalytics({
+      category: "User",
+      action: "Config Panel opened",
```
barrymcgee, line 98: Nit: Sentence case?

barrymcgee (approved):

src/panels/ConfigPanel/ConfigPanel.tsx
hatched, line 98:
```suggestion
      action: "Config panel opened",
```

PR #897, opened 2021-03-02
Implement the breadcrumb component in the header

anthonydillon:
## Done
Implement the breadcrumb component in the header
Drive-by, fix error while awaiting modelStatusData to become available

## QA
- Load the demo and go to a model details
- See that the breadcrumb renders just the model name
- Go into an application and check the breadcrumb now matches [the design](https://zpl.io/aN63dv4)
- To test the drive-by, load the branch locally
- Go to an application details
- Refresh and see that the page loads without an error

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/891

## Screenshots
image

### DIFF
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -11,6 +11,9 @@ export function generateEntityIdentifier(
   baseAppURL,
   disableLink = false
 ) {
+  if (!namespace) {
+    return [];
+  }
   let charmStorePath = "";
   try {
     charmStorePath = URL.fromAnyString(namespace).toString().replace("cs:", "");
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -12,7 +12,7 @@ export function generateEntityIdentifier(
   disableLink = false
 ) {
   if (!namespace) {
-    return [];
+    return null;
   }
   let charmStorePath = "";
   try {
--- a/src/components/Breadcrumb/Breadcrumb.test.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.test.tsx
@@ -1,0 +1,54 @@
+import configureStore from "redux-mock-store";
+import TestRoute from "components/Routes/TestRoute";
+import { mount } from "enzyme";
+import React from "react";
+import { Provider } from "react-redux";
+import { MemoryRouter } from "react-router-dom";
+import dataDump from "testing/complete-redux-store-dump";
+
+import Breadcrumb from "./Breadcrumb";
+
+const mockStore = configureStore([]);
+
+describe("Breadcrumb", () => {
+  it("displays correctly on model details", () => {
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter initialEntries={["/models/eggman@external/group-test"]}>
+          <TestRoute path="/models/:userName/:modelName?">
+            <Breadcrumb />
+          </TestRoute>
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("[data-test='breadcrumb-application']").length).toBe(0);
+    expect(wrapper.find("[data-test='breadcrumb-model']").text()).toStrictEqual(
+      "group-test"
+    );
+  });
+
+  it("displays correctly on application details", () => {
+    const store = mockStore(dataDump);
+    const wrapper = mount(
+      <Provider store={store}>
+        <MemoryRouter
+          initialEntries={["/models/eggman@external/group-test/app/easyrsa"]}
+        >
+          <TestRoute path="/models/:userName/:modelName?/app/:appName?">
+            <Breadcrumb />
+          </TestRoute>
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("[data-test='breadcrumb-model']").text()).toStrictEqual(
+      "group-test"
+    );
+    expect(
+      wrapper.find("[data-test='breadcrumb-section']").text()
+    ).toStrictEqual("Applications");
+    expect(
+      wrapper.find("[data-test='breadcrumb-application']").text()
+    ).toStrictEqual("easyrsa");
+  });
+});
--- a/src/components/Breadcrumb/Breadcrumb.test.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.test.tsx
@@ -41,6 +41,9 @@ describe("Breadcrumb", () => {
         </MemoryRouter>
       </Provider>
     );
+    expect(wrapper.find("[data-test='breadcrumb-items']").text()).toStrictEqual(
+      "group-testApplicationseasyrsa"
+    );
     expect(wrapper.find("[data-test='breadcrumb-model']").text()).toStrictEqual(
       "group-test"
     );
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -1,11 +1,7 @@
 import { Link } from "@canonical/react-components";
 import { useParams } from "react-router-dom";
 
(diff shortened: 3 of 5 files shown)


src/app/utils/utils.js
```diff
@@ -11,6 +11,9 @@ export function generateEntityIdentifier(
   baseAppURL,
   disableLink = false
 ) {
+  if (!namespace) {
+    return [];
```
hatched, line 15: This should return `null`. An array is of a different type than a `JSX.Element` which is what this normally returns. If this was in TS it would have complained ;)

src/components/Breadcrumb/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <MemoryRouter
+          initialEntries={["/models/eggman@external/group-test/app/easyrsa"]}
+        >
+          <TestRoute path="/models/:userName/:modelName?/app/:appName?">
+            <Breadcrumb />
+          </TestRoute>
+        </MemoryRouter>
+      </Provider>
+    );
+    expect(wrapper.find("[data-test='breadcrumb-model']").text()).toStrictEqual(
```
hatched, line 44: I was expecting this test to check for something like "group-test > Applications > easyrsa" instead of grabbing each element individually. Can you move the `find` up to `p-breadcrumbs__items` to achieve this?

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,53 @@
+import { Link } from "@canonical/react-components";
+import { useParams } from "react-router-dom";
+
+export type EntityDetailsRoute = {
+  userName: string;
+  modelName: string;
+  appName: string;
+};
+
+const Breadcrumb = (): JSX.Element => {
```
hatched, line 10: Just to maintain consistency, we typically use the `function` signature for components and then export them when they are defined...

```ts
export default function Breadcrumb(): JSX.Element {
```

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,53 @@
+import { Link } from "@canonical/react-components";
+import { useParams } from "react-router-dom";
+
+export type EntityDetailsRoute = {
```
hatched, line 4: You can import this type from the `Routes` component.

```ts
import type { EntityDetailsRoute } from "components/Routes/Routes";
```

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <nav className="p-breadcrumbs" aria-label="Breadcrumb navigation">
+      <ol className="p-breadcrumbs__items">
+        {appName ? (
+          <>
+            <li className="p-breadcrumbs__item" data-test="breadcrumb-model">
+              <Link
+                href={`/models/${
+                  userName ? `${userName}/${modelName}` : `${modelName}`
```
hatched, line 21: This logic is a bit hard to follow and repeated, can you abstract it out into a utility function in this file?

hatched (changes requested):
Thanks @anthonydillon just a few small cleanups requested

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const { userName, modelName, appName } = useParams<EntityDetailsRoute>();
+
+  return (
+    <nav className="p-breadcrumbs" aria-label="Breadcrumb navigation">
+      <ol className="p-breadcrumbs__items">
+        {appName ? (
+          <>
+            <li className="p-breadcrumbs__item" data-test="breadcrumb-model">
+              <Link
+                href={`/models/${
```
hatched, line 20: Sorry i missed this in the original review...this should probably use `to`

https://reactrouter.com/web/api/Link

src/app/utils/utils.js
anthonydillon: Sure, `tableRows.js` does this a lot so thought it was the way you guys do it.

hatched: Ahh np, so the difference there is that on the 'green' path it will return an array of data. So if there is missing data it'll still return an empty array (the same type of data).

src/components/Breadcrumb/Breadcrumb.test.tsx
anthonydillon: Not sure a test to check the order is correct is that valuable but I have added it.

src/components/Breadcrumb/Breadcrumb.tsx
anthonydillon: Seems to be a mix in the codebase but no problem

src/components/Breadcrumb/Breadcrumb.tsx
anthonydillon: Nice, done

src/components/Breadcrumb/Breadcrumb.tsx
anthonydillon: Good idea

src/components/Breadcrumb/Breadcrumb.tsx
anthonydillon: Thanks done

anthonydillon:
@hatched thanks, think I have addressed all your comments.

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <nav className="p-breadcrumbs" aria-label="Breadcrumb navigation">
+      <ol className="p-breadcrumbs__items" data-test="breadcrumb-items">
+        {appName ? (
+          <>
+            <li className="p-breadcrumbs__item" data-test="breadcrumb-model">
+              <Link to={generateModelURL()}>{modelName}</Link>
+            </li>
+            <li className="p-breadcrumbs__item" data-test="breadcrumb-section">
+              <Link to={generateModelURL()}>Applications</Link>
```
barrymcgee, line 26: I find it strange that the only two clickable breadcrumbs both link to the same place?

src/components/Breadcrumb/Breadcrumb.tsx
hatched: Hmm yeah we should fix that.

src/components/Breadcrumb/Breadcrumb.tsx
hatched: This is per design but it'll change as more pages are implemented into the breadcrumb.

hatched (approved):

hatched:
QA Note:
When going from an application page to the model page the sidebar bounces open and then gets caught under the cursor. I believe this is resolved in Barry's upcoming branch.

src/components/Breadcrumb/Breadcrumb.tsx
anthonydillon: I agree, and design may change there mind now its in practise.

ziheliu214:
When I refresh the page it shows: 
Screenshot 2021-03-03 at 11 25 31

Another issue might need to be addressed separately - every time jumping pages, the left nav has the animation effect expanding then collapsing, a bit hindering the view.

Apart from that, UX is good!

barrymcgee:
@ziheliu214 Refreshing on the demo service will always produce that response, unfortunately.

I have a fix for the bouncing sidebar incoming in a separate branch.

ziheliu214:
> @ziheliu214 Refreshing on the demo service will always produce that response, unfortunately.
> 
> I have a fix for the bouncing sidebar incoming in a separate branch.

Thanks Barry :D

PR #900, opened 2021-03-03
Update dockerfile to use serve instead of http.server

hatched:
## Done

This change allows reviewers to refresh the application at any route and have the correct assets served for that path instead of having it 404 on all paths but `/`.

## QA

- Visit the demo link and navigate around the application.
- Refresh at various points and it should load the app to those pages.


Dockerfile
```diff
@@ -8,29 +8,10 @@ ADD package.json .
 ADD yarn.lock .
 RUN --mount=type=cache,target=/usr/local/share/.cache/yarn yarn install
 
-
-# Build stage: Run "yarn run build-js"
-# ===
 FROM yarn-dependencies AS build-js
```
nottrobin, line 11: It's pointless splitting this into 2 stages if the second stage is just going to inherit from the first

nottrobin: You also don't need to give your final stage a name. It's not going to be used by anything.

Dockerfile
```diff
@@ -8,29 +8,10 @@ ADD package.json .
...
-# Import code, build assets and mirror list
-ADD . .
-RUN rm -rf package.json yarn.lock .babelrc webpack.config.js
-COPY --from=build-js /srv/build .
+RUN yarn global add serve
 
 # Setup commands to run server
-ENTRYPOINT ["python3"]
-CMD ["-m", "http.server", "80"]
+ENTRYPOINT ["serve"]
```
nottrobin, line 17: As João said, you should serve on port 80 if you can.

hatched: Updated this to serve on port 80

Dockerfile
hatched: Good call, updated.

barrymcgee (approved):

PR #916, opened 2021-03-10
Panel improvements

barrymcgee:
## Done

- Remove duplicate `<main>` element from Login component
- Stop panel form overflowing
- Add back the panel box shadow
- Close panel when escape key is clicked

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Verify as above


src/components/LogIn/LogIn.js
```diff
@@ -50,7 +50,7 @@ export default function LogIn({ children }) {
           </FadeUpIn>
         </div>
       )}
-      <main>{children}</main>
+      <div className="main">{children}</div>
```
hatched, line 53: maybe a more appropriate class name would be something like `app-content` ?

src/panels/panels.tsx
```diff
@@ -1,12 +1,29 @@
...
+    }
+  },
+};
+
 export default function Panels() {
-  const panelQs = useQueryParam("panel", StringParam)[0];
+  const [panelQs, setPanelQs] = useQueryParam("panel", StringParam);
+
+  useEventListener("keydown", (e: KeyboardEvent) =>
+    close.onEscape(e, setPanelQs)
```
hatched, line 25: I'm not sure we should land this escape code as-is as we know it's going to cause issues right away if the panel content cannot prevent it from happening. 

I'd prefer it to not close on escape until we have an implementation that is cancellable.

hatched (changes requested):

src/panels/panels.tsx
```diff
@@ -1,12 +1,21 @@
...
 import RegisterController from "components/RegisterController/RegisterController";
 
 import "./_panels.scss";
 
 export default function Panels() {
-  const panelQs = useQueryParam("panel", StringParam)[0];
+  const [panelQs, setPanelQs] = useQueryParam("panel", StringParam);
+
+  // Close panel if Escape key is pressed when panel active
+  const closeOnEscape = (e: KeyboardEvent) => {
```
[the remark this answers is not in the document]
barrymcgee: Done

hatched (approved):

PR #924, opened 2021-03-15
Fix aside styling and animations

hatched:
## Done

- Horizontal scroll bars no longer appear when aside panels animate in.
- Removed fudge values to position panel content in aside panels.
- Fixed animation so that it animated the aside and not the panel.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Click to open the register controller panel, it should animate in smoothly with a drop shadow and no scrollbar appearing for the app.
- You can also use the Actions panel if you have a model with actions.

## Details

Fixes: https://github.com/canonical-web-and-design/jaas-dashboard/issues/880

### DIFF
--- a/src/animations/SlideInOut.tsx
+++ b/src/animations/SlideInOut.tsx
@@ -3,16 +3,19 @@ import { motion } from "framer-motion";
 type Props = {
   isActive: boolean;
   children: JSX.Element;
+  className: string;
 };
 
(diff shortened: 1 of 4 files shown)


anthonydillon (approved):
Works but please file an issue upstream on vanilla and XXX it here in the overrides

hatched:
Thanks, will do!

PR #926, opened 2021-03-18
Add interactivity to selecting an action in the Actions panel

hatched:
## Done

The actions list is now styled to match the designs.
- Selecting an action opens the actions option list (if one exists)
- If an option description is longer than 30 characters it'll be collapsed and allow for expansion.

Note:
It's still not functional - this is purely UI interactivity. Because of this the functionality tests are going to come in the follow-up branch which hooks this up to the API.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View a charm with actions, select an action and see the available interactions

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1663

## Screenshots
Screen Shot 2021-03-18 at 3 26 40 PM

heightanim

### DIFF
--- a/src/components/RadioInputBox/RadioInputBox.tsx
+++ b/src/components/RadioInputBox/RadioInputBox.tsx
@@ -62,7 +62,10 @@ export default function RadioInputBox({
         {options.map((option) => {
           const inputKey = `${option.name}Input`;
           return (
-            <div key={`${option.name}InputGroup`}>
+            <div
+              className="radio-input-box__input-group"
+              key={`${option.name}InputGroup`}
+            >
               <label
                 className={classnames("radio-input-box__label", {
                   "radio-input-box__label--required": option.required,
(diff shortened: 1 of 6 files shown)


src/components/RadioInputBox/RadioInputBox.tsx
```diff
@@ -1,51 +1,155 @@
...
+      // override the height set in css for the collapsed size.
+      wrapper.style.height = endHeight + "px";
+    } else if (wrapper.offsetHeight !== initialHeight) {
+      // Do not animate if the wrapper height is already in the closed state
+      // and this hook runs for the closed state again.
+      endHeight = initialHeight;
+      startHeight = container.offsetHeight + paddingNumber;
+      duration = startHeight;
+    }
+    const animation = wrapper.animate(
```
hatched, line 73: This is using [WAAPI](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Web_Animations_API_Concepts) so it's still as performant as doing it in CSS but this actually works 😆

anthonydillon (changes requested):
This looks good. Can you add some tests to the new element. 

There are a bunch of Vanilla patterns that could be used here. Such as required form elements, etc. But they can be cleaned up after.

anthonydillon (approved):
LGTM :+1:

hatched:
Thanks all

PR #951, opened 2021-04-13
Indicate the provider type throughout entity details

anthonydillon:
## Done
Add the provider type of the model to all entity info panels. So it is clear which provider you are using at all depths.

## QA
- Open https://jaas-dashboard-951.demos.haus
- Go to a model details page and see the provider info in the info panel
- Go to a app details page and see the provider info in the info panel
- Go to a unit details page and see the provider info in the info panel

## Details
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1722

## Screenshots
Model details:
image

App details:
image

Unit details:
image

### DIFF
--- a/src/pages/EntityDetails/Model/Model.js
+++ b/src/pages/EntityDetails/Model/Model.js
@@ -116,6 +116,7 @@ const Model = () => {
     "Cloud/Region": `${cloudProvider} / ${modelStatusData?.model.region}`,
     version: modelStatusData?.model.version,
     sla: modelStatusData?.model.sla,
+    provider: modelStatusData?.info?.["provider-type"],
   };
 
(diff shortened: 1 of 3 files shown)


src/pages/EntityDetails/Model/Model.js
```diff
@@ -116,6 +116,7 @@ const Model = () => {
     "Cloud/Region": `${cloudProvider} / ${modelStatusData?.model.region}`,
     version: modelStatusData?.model.version,
```
hatched, line 117: We should dynamically generate these keys based on the existence of the values so we don't have a number of empty fields in the UI if the data isn't available.

hatched (approved):
Thanks for this fix @anthonydillon. I'd like to see a follow-up sometime that would remove empty fields.

PR #957, opened 2021-04-13
Add a provider icon to the model list

anthonydillon:
## Done
Add known provider icon to the model list. Supporting "aws", "gce", "azure" and "kubernetes".

## QA
- Open the demo
- Check the provider type icons you see on the model list

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/954

## Screenshots
image


src/components/ModelTableList/StatusGroup.js
```diff
@@ -77,6 +82,67 @@ const generateModelNameCell = (model, groupLabel) => {
   );
 };
 
+/**
+  Generates the cloud adnd region info from model data.
```
hatched, line 86:
```suggestion
  Generates the cloud and region info from the model data.
```

src/components/ModelTableList/StatusGroup.test.js
```diff
@@ -59,4 +59,20 @@ describe("StatusGroup", () => {
     );
     expect(wrapper.find("tbody TableRow").length).toBe(3);
   });
+
+  it("displays the provider type icon", () => {
```
hatched, line 63: Should this be expanded to test for each icon type?

You can clone the `dataDump` to hack in different provider names if you need.

hatched (changes requested):

hatched:
I love the extra colour this adds to the page :)

However there are some qa issues:
- GCE icon doesn't line up.
- AWS icon doesn't show up.

Screen Shot 2021-04-13 at 3 12 33 PM

clagom:
Are we sure that one is the GCE logo?

hatched:
> Are we sure that one is the GCE logo?

Hah, right, that's the azure logo :D

anthonydillon:
@clagom you right it doesn't look right. @long-chung I think the google cloud platform should look like this: https://assets.ubuntu.com/v1/42299ebc-google-cloud.png

anthonydillon:
Haha, sorry about that. mixed up the imports.

anthonydillon:
@hatched also aligned stuff

src/components/ModelTableList/StatusGroup.js
anthonydillon: Done

src/components/ModelTableList/StatusGroup.test.js
anthonydillon: Happy to but not sure it really gives us that much. At that point, it's a switch case. In the future, I would probably make this its own component so it can be clearing unit tested.

long-chung:
Icon alignment issues on Chrome and FF

Screenshot 2021-04-14 at 16 13 17


Also could you add an 8px 0.5rem margin right to the icon as well pls

hatched:
Much better now @anthonydillon thanks

Screen Shot 2021-04-14 at 12 31 10 PM

I know this PR is getting long in the tooth but can you add the icons to the cells in the `Owner` grouping as well?
I'd also like to see them in the header in the `Cloud` grouping but I'm not sure how that would look :)

anthonydillon:
No prob, they are good points and I have a little code refactoring to do to tidy up some duplicated code.

anthonydillon:
@hatched refactored and shared the clouds with the owners view. Couldn't get the logo to look ok on the cloud view so will leave that to design.

src/components/ModelTableList/shared.js
```diff
@@ -129,3 +134,76 @@ export function getStatusValue(status, key) {
   }
   return returnValue;
 }
+
+/**
+  Generates the cloud and region info from model data.
+  @param {Object} model The model data.
+  @returns {Object} The React element for the model cloud and region cell.
+*/
+export const generateCloudCell = (model) => {
```
hatched, line 143: Looks like this file uses the style

`export function foo() {}` would you mind updating these to be consistent with the rest of the file?

hatched (approved):

hatched:
Looks great @anthonydillon thanks a lot! I just have one tiny request then this can land :)

PR #958, opened 2021-04-13
Display the version of an application correctly on the application list

anthonydillon:
## Done
Display the version of an application correctly on the application list

## QA
- Check out the demo
- Add an application with a version and see that you see it in the model details

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/955

## Screenshots
image

**This wrapping will be fixed with a change to the table styling in https://github.com/canonical-web-and-design/jaas-dashboard/pull/957**

### DIFF
--- a/src/tables/tableRows.js
+++ b/src/tables/tableRows.js
@@ -33,7 +33,7 @@ export function generateLocalApplicationRows(
     const rev = extractRevisionNumber(app.charm) || "-";
     const store = app.charm.indexOf("local:") === 0 ? "Local" : "CharmHub";
     const scale = app.unitsCount;
-    const version = app.workloadVersion || "-";
+    const version = app["workload-version"] || "-";
 
     return {
       columns: [
@@ -50,7 +50,6 @@ export function generateLocalApplicationRows(
         {
           "data-test-column": "version",
           content: version,
-          className: "u-align--right",
         },
         {
           "data-test-column": "scale",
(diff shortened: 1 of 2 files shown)


src/tables/tableRows.js
```diff
@@ -50,7 +50,6 @@ export function generateLocalApplicationRows(
         {
           "data-test-column": "version",
           content: version,
-          className: "u-align--right",
```
anthonydillon, line 53: Removed the right alignment as versions are not numerical

hatched: This value is user defined so it 'can' be a number. But I think there is a bug here...you'll see the app has a workload version defined but it's not rendered in the UI.

Screen Shot 2021-04-13 at 3 31 44 PM

hatched: Should we also truncate these versions and have tooltips with the full value if they can be that long in K8s land?

hatched (changes requested):

src/tables/tableRows.js
anthonydillon: Mentioned under the screenshot that once the other branch lands it will truncate the version. The title would be a good follow up branch.

anthonydillon: Well it could be in odd occasions but you wouldn't define is a number type. Even 2.3.4 is not a number.

anthonydillon:
I don't understand, that is what I changed. Could you double check the value of the version is that is being sent from the diff?

hatched:
There appears to have been a caching issue or something as when I check it out locally it works as expected. Thanks!

hatched (approved):

PR #962, opened 2021-04-15
Remove machines for k8s

anthonydillon:
## Done
- Remove the models machines tab on k8s
- Hide the Machine switcher on application view on kubernetes
- Hide machine table on units view on k8s 
- Remove machine column from app view

## QA
- You will need K8s model so ask me for that or we can pair on the QA
- Check all the above items are addressed

## Details
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1721

### DIFF
--- a/public/config.js
+++ b/public/config.js
@@ -3,15 +3,15 @@
 // eslint-disable-next-line no-unused-vars
 var jujuDashboardConfig = {
   // API host to allow app to connect and retrieve models
-  baseControllerURL: "jimm.jujucharms.com",
+  baseControllerURL: "localhost:17070",
   // Configurable base url to allow deploying to different paths.
   baseAppURL: "/",
   // If true then identity will be provided by a third party provider.
-  identityProviderAvailable: true,
+  identityProviderAvailable: false,
   // Is this application being rendered in Juju and not JAAS. This flag should
   // only be used for superficial updates like logos. Use feature detection
   // for other environment features.
-  isJuju: false,
+  isJuju: true,
   // Development flags below:
   // Show WebCli component if enabled for dev purposes, even if not natively supported.
   showWebCLI: false,
--- a/public/config.js
+++ b/public/config.js
@@ -3,15 +3,15 @@
 // eslint-disable-next-line no-unused-vars
 var jujuDashboardConfig = {
   // API host to allow app to connect and retrieve models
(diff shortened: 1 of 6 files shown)


anthonydillon:
@hatched if you get any time could you commit how you would conditionally remove the machine row data and header from the application view. The condition is: `modelStatusData.info["provider-type"] !== "kubernetes"`

hatched:
There you go @anthonydillon It's a bit of a hacky approach but it gets the job done for now. I'm ok with it because we'll be refactoring all this with the new modular table in the near future.

anthonydillon:
Thanks Jeff. Haha I wrote it this way and thought it was too hacky. I was worried that if someone changed the table it would hide the wrong thing and we would never spot it. 

But I think this could work if I add IDs to the columns and loop over and remove the one with machine as a key.

src/pages/EntityDetails/App/App.tsx
```diff
@@ -204,11 +216,13 @@ export default function App(): JSX.Element {
         </>
       </div>
       <div className="entity-details__main u-overflow--scroll">
-        <ButtonGroup
-          buttons={["units", "machines"]}
-          activeButton={tableView}
-          setActiveButton={setTableView}
-        />
+        {modelStatusData?.info?.["provider-type"] !== "kubernetes" && (
```
hatched, line 219: Can you replace this check with the `hideMachines` value from above?

public/config.js
```diff
@@ -3,15 +3,15 @@
 // eslint-disable-next-line no-unused-vars
 var jujuDashboardConfig = {
   // API host to allow app to connect and retrieve models
-  baseControllerURL: "jimm.jujucharms.com",
+  baseControllerURL: "localhost:17070",
```
hatched, line 6: Oops :) Can you remove these changes plz.

hatched (changes requested):
Minor changes requested. Nice improvement on my hack :)

public/config.js
anthonydillon: Done

src/pages/EntityDetails/App/App.tsx
anthonydillon: Good spot, done

hatched (approved):
Thanks for keeping up with this 👍

PR #975, opened 2021-04-27
Fetch charm icon from charmhub for charms deployed from there

### LINKED ISSUE #972: Icons are not rendered in Juju 2.9
When deployed to a controller with Juju 2.9 icons are no longer displayed. This is because the charm url that is provided is now the charmhub style `ch:amd64/xenial/ceph-457` and not the charmstore style.

There is no way to get to the charm url directly without going through the API first:
`curl -s "https://api.charmhub.io/v2/charms/info/ceph-mon?fields=result.media" | jq .result.media[]?.url`. 

We should fetch the icons from the controller directly instead of going to charmhub so that we can support fetching public as well as local charms icons from a single location. This will simplify the code but also reduce errors as the user will always need access to the controller but may be in an environment that does not permit access to a public domain.

hatched:
## Done

Fetch charm icon from charmhub for charms deployed from there.

## QA

- In the demo icons should be rendered as expected.
- Bootstrap your own controller with Juju 2.9 
- Point the dashboard to it using [these instructions](https://github.com/canonical-web-and-design/jaas-dashboard#developing-while-connected-to-a-juju-controller)

## Details

Fixes #972 

## Screenshots

Screen Shot 2021-04-27 at 2 11 00 PM

### DIFF
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -251,6 +251,11 @@ export const generateIconPath = (charmId) => {
     charmId = charmId.replace("cs:", "");
     return `https://api.jujucharms.com/charmstore/v5/${charmId}/icon.svg`;
   }
+  if (charmId.indexOf("ch:") === 0) {
+    // XXX This is a temporary fix until charmhub.io can provide us with an
+    // icon endpoint.
+    return defaultCharmIcon;
+  }
   return "";
 };
 
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -1,42 +1,17 @@
-import { URL } from "@canonical/jaaslib/lib/urls";
-
 import cloneDeep from "clone-deep";
 
 import defaultCharmIcon from "static/images/icons/default-charm-icon.svg";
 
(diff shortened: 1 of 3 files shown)


src/app/utils/utils.js
```diff
@@ -265,16 +240,21 @@ export const extractRevisionNumber = (charmName) => charmName.split("-").pop();
...
     return defaultCharmIcon;
   }
-  if (namespace.indexOf("cs:") === 0) {
-    namespace = namespace.replace("cs:", "");
-    return `https://api.jujucharms.com/charmstore/v5/${namespace}/icon.svg`;
+  if (charmId.indexOf("cs:") === 0) {
+    charmId = charmId.replace("cs:", "");
+    return `https://api.jujucharms.com/charmstore/v5/${charmId}/icon.svg`;
+  }
+  if (charmId.indexOf("ch:") === 0) {
```
hatched, line 254: All the lines of code changed outside of this conditional are a driveby to clean-up the code.

barrymcgee (approved):

PR #983, opened 2021-04-29
Add Sharing panel to model detail pages

barrymcgee:
## Done

Add Sharing panel to model detail pages, showing who a model has already been shared with.

## QA

- Pull code
- Click on a model 
- Click on 'Share model' under the topology
- Verify the share model panel opens and lists the users the model has been shared with

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1737,
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1739

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "classnames": "2.3.1",
     "clone-deep": "4.0.1",
     "d3": "6.7.0",
+    "date-fns": "^2.21.1",
     "formik": "2.2.6",
     "framer-motion": "4.1.9",
     "immer": "9.0.2",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "classnames": "2.3.1",
     "clone-deep": "4.0.1",
     "d3": "6.7.0",
-    "date-fns": "^2.21.1",
+    "date-fns": "2.21.1",
     "formik": "2.2.6",
     "framer-motion": "4.1.9",
     "immer": "9.0.2",
--- a/src/app/selectors.js
+++ b/src/app/selectors.js
@@ -141,7 +141,7 @@ const getDecodedMacaroons = (macaroons) => {
 const getModelUUIDByName = (name, modelData) => {
   let owner = null;
   let modelName = null;
-  if (name.includes("/")) {
+  if (name?.includes("/")) {
     [owner, modelName] = name.split("/");
   } else {
     modelName = name;
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -1,4 +1,3 @@
-import { URL } from "@canonical/jaaslib/lib/urls";
 import { parseISO, formatDistanceToNow } from "date-fns";
 import cloneDeep from "clone-deep";
 
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -1,3 +1,5 @@
+import { URL } from "@canonical/jaaslib/lib/urls";
+import { parseISO, formatDistanceToNow } from "date-fns";
 import cloneDeep from "clone-deep";
 
 import defaultCharmIcon from "static/images/icons/default-charm-icon.svg";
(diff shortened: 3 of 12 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "classnames": "2.3.1",
     "clone-deep": "4.0.1",
     "d3": "6.7.0",
+    "date-fns": "^2.21.1",
```
hatched: can you remove the ^ plz

src/app/selectors.js
```diff
@@ -141,7 +141,7 @@ const getDecodedMacaroons = (macaroons) => {
 const getModelUUIDByName = (name, modelData) => {
   let owner = null;
   let modelName = null;
-  if (name.includes("/")) {
+  if (name?.includes("/")) {
```
hatched, line 144: Why this was being called without a model name?

src/app/utils/utils.js
```diff
@@ -1,3 +1,5 @@
+import { URL } from "@canonical/jaaslib/lib/urls";
```
hatched, line 1: This doesn't look like it's used?

src/app/utils/utils.js
```diff
@@ -403,3 +405,7 @@ export const generateRelationIconImage = (applicationName, modelStatusData) => {
   }
   return generateIconImg(applicationName, application.charm);
 };
+
+export const formatFriendlyDateToNow = (date) => {
+  return formatDistanceToNow(parseISO(date), { addSuffix: true });
```
hatched, line 410: Should we have this support a tooltip which has the actual value in it?

src/components/Aside/Aside.tsx
```diff
@@ -23,7 +27,13 @@ export default function Aside({
         "is-pinned": pinned === true,
       })}
     >
-      {children}
+      {!loading ? (
+        children
+      ) : (
+        <div className="loading">
```
hatched, line 33: The loading indicator is only rendered in the content where data is missing so in the config case, only the left column gets a spinner and in this case, the same, the 'add a user' column would not get a spinner so this should be handled in the panel itself, not the aside.

src/panels/ShareModelPanel/ShareModel.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+            <QueryParamProvider ReactRouterRoute={Route}>
+              <ShareModel />
+            </QueryParamProvider>
+          </TestRoute>
+        </Provider>
+      </MemoryRouter>
+    );
+    const firstUserCard = wrapper.find(".share-model__card").first();
+    const firstUserName = firstUserCard.find(".share-model__card-title strong");
+    expect(firstUserName.text()).toBe("eggman@external");
```
hatched, line 30: Are there multiple users here? Might be best to check that all are listed.

hatched (changes requested):

src/app/selectors.js
barrymcgee: The test was calling it without a model name before I set up the params provider.

src/app/utils/utils.js
barrymcgee: I didn't explicity add that.. must have been VS Code - removing.

src/app/utils/utils.js
barrymcgee: There's already an issue to do that [here](https://github.com/canonical-web-and-design/jaas-dashboard/issues/965).. that why I made this a common util.

src/panels/ShareModelPanel/ShareModel.test.tsx
barrymcgee: In the data dump, there is only one.

src/panels/ShareModelPanel/ShareModel.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export default function ShareModel() {
+  const { modelName } = useParams<EntityDetailsRoute>();
+
+  const modelStatusData: TSFixMe = useModelStatus() || null;
+
+  const users = modelStatusData?.info?.users;
+
+  const isOwner = (user: string) => {
+    return user === modelStatusData?.info["owner-tag"].split("-")[1];
```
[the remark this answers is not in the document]
barrymcgee: Good catch - refactored.

hatched (approved):

PR #1003, opened 2021-05-20
Add an empty column to a subordinate unit table

anthonydillon:
## Done
Add an empty column to a subordinate unit table
Also drive by, to update the empty unit table message to be application now model.

## QA
- Open the demo
- Go to a model with a subordinate (hadoopspark(gomboli) > slave)
- Check the subordinate columns now align correctly

## Details
Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1765

## Screenshots
image

### DIFF
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -9,7 +9,7 @@ export function generateEntityIdentifier(charmId, name, subordinate) {
   }
 
   return (
-    <div className="entity-name">
+    <div className="entity-name u-truncate" title={name}>
       {subordinate && <span className="subordinate"></span>}
       {charmId && generateIconImg(name, charmId)}
       {name}
--- a/src/pages/EntityDetails/App/App.test.js
+++ b/src/pages/EntityDetails/App/App.test.js
@@ -207,7 +207,7 @@ describe("Entity Details App", () => {
     ].applications.etcd.units = null;
     const wrapper = await generateComponent(tweakedData);
     expect(wrapper.find("MainTable caption").text()).toBe(
-      "There are no units in this model"
+      "There are no units in this application"
     );
   });
 });
--- a/src/pages/EntityDetails/App/App.tsx
+++ b/src/pages/EntityDetails/App/App.tsx
@@ -284,7 +284,7 @@ export default function App(): JSX.Element {
                     rows={unitPanelRows}
                     className="entity-details__units p-main-table panel__table has-checkbox"
                     sortable
-                    emptyStateMsg={"There are no units in this model"}
+                    emptyStateMsg={"There are no units in this application"}
                     data-test="units-table"
                   />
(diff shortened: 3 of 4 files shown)


hatched (approved):

long-chung:
Text is still cut off. NB this was on a 1680px browser width

Screenshot 2021-05-21 at 09 33 58

anthonydillon:
@long-chung thanks. What would you like it to do in that situation?

long-chung:
@anthonydillon Would it be possible to ellipses it and add a title attribute so it shows the full text on hover?

ziheliu214:
Thanks Ant.

The alignment is right, as @long-chung said maybe a tooltip for full text would be good.

Not sure if its related to this issue, but the status pill (2 allocating, 1 lost) doesn't reflect the workload status (2 waiting, 1 unknown)
Screenshot 2021-05-21 at 11 22 53

The subnet unit 'ganglia-node/0' should link to its unit page just as the principle units. 
Also its weird as it seems ganlia-node app has a unit 0, but on the application list its scale is 0. Same to other subodinate apps.
Screenshot 2021-05-21 at 11 25 24

anthonydillon:
> Would it be possible to ellipses it and add a title attribute so it shows the full text on hover?

@long-chung sure.

> Not sure if its related to this issue, but the status pill (2 allocating, 1 lost) doesn't reflect the workload status (2 waiting, 1 unknown)

@ziheliu214 There are a bunch of issues created by @hatched to cover this miss count of subordinate entities. This count is infact correct (minus the subordinate units) when referring to the agent not workload.

ziheliu214:
> > Would it be possible to ellipses it and add a title attribute so it shows the full text on hover?
> 
> @long-chung sure.
> 
> > Not sure if its related to this issue, but the status pill (2 allocating, 1 lost) doesn't reflect the workload status (2 waiting, 1 unknown)
> 
> @ziheliu214 There are a bunch of issues created by @hatched to cover this miss count of subordinate entities. This count is infact correct (minus the subordinate units) when referring to the agent not workload.

Interesting, I may have overlooked that units have 2 statuses. But then do we know which is the key indicator of a working unit that users look at? Maybe we should move the status colour to the 'agent' column instead to match the pills?

anthonydillon:
@long-chung added the truncation and title

anthonydillon:
@ziheliu214 can you suggest some fix for the issue in this PR?

ziheliu214:
> @ziheliu214 can you suggest some fix for the issue in this PR?

Hi I just asked an Openstack engineer who uses juju, so the status value of unit ( https://juju.is/docs/olm/status-values) refers to the workload column, though they would check all 3 columns (workload, agent, notes) to understand if this unit is working normally.

So I suggest to change the pill data to mirror the workload value instead, so they will have same colour indication too.

PR #1021, opened 2021-05-31
Fail gracefully when action logs includes missing applications

### LINKED ISSUE #1020: ActionLogs page fails when charm has been removed
To reproduce:
- Add a charm and run a few actions.
- Then remove the charm.
- View the Action Logs page.

` Uncaught TypeError: modelStatusData.applications[appName] is undefined`

hatched:
## Done

If the action logs contains applications that are no longer in the model then fail gracefully by only rendering the application name.

## QA

- Contact me for access to a test model.... or...
- Create a new model and `juju deploy cs:ceph-mon`.
- Run a simple action like `get-health`.
- View the action logs page.
- `juju remove-application ceph-mon`
- View the action logs page.
- It shouldn't 💥 

## Details

Fixes #1020 

## Screenshots

Without application data:
Screen Shot 2021-05-31 at 11 33 36 AM

With application data:
Screen Shot 2021-05-31 at 11 36 09 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "juju-dashboard",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "description": "A dashboard for Juju and JAAS (Juju as a service)",
   "bugs": {
     "url": "https://github.com/canonical-web-and-design/jaas-dashboard/issues"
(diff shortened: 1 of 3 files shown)


huwshimi (approved):
Looks good!

ziheliu214:
So after removing the application, its action log disappears:
Screenshot 2021-06-01 at 16 48 24

ziheliu214:
Had a call with @hatched, the edge case above happened when the `run-action` is run on CLI.

hatched:
Just to expand on this a bit more...
The dashboard uses the new actions spec (tasks & operations) whereas the CLI uses the old actions spec. If the actions are run on the CLI and then the application is removed the actions are stored as part of the application and so they are not returned as part of the operations list. However if you execute the same actions with the new spec (even on the cli using the actions-v2 flag) then the data gets stored in the operations list and is returned.

We think that this is acceptable as it's an edge case to begin with and this is even further down that rabbit hole.

PR #1026, opened 2021-06-04
Add Toast notifications to sharing UI

barrymcgee:
## Done

Add Toast notifications to sharing UI

## QA

- Go to model
- Open sharing panel
- Add new user / change an existing users access / delete a user
- See positive action toast
- Try to add a user which already exists
- See negative (error) toast

### DIFF
--- a/src/components/ToastCard/ToastCard.test.tsx
+++ b/src/components/ToastCard/ToastCard.test.tsx
@@ -1,9 +1,10 @@
 import { mount } from "enzyme";
+import cloneDeep from "clone-deep";
 
 import ToastCard from "./ToastCard";
 
 describe("Toast Card", () => {
-  const t = {
+  const toastInstanceExample = {
     createdAt: 1623162274616,
     duration: 5000,
     id: "2",
@@ -15,6 +16,7 @@ describe("Toast Card", () => {
   };
 
   it("should display message", () => {
+    const t = cloneDeep(toastInstanceExample);
     const wrapper = mount(
       <ToastCard
         type="positive"
@@ -28,6 +30,7 @@ describe("Toast Card", () => {
   });
 
   it("should display as correct type", () => {
+    const t = cloneDeep(toastInstanceExample);
     const wrapper = mount(
       <ToastCard
         type="positive"
@@ -39,6 +42,7 @@ describe("Toast Card", () => {
   });
 
   it("should display correct success icon", () => {
+    const t = cloneDeep(toastInstanceExample);
     const wrapper = mount(
       <ToastCard
         type="positive"
@@ -50,6 +54,7 @@ describe("Toast Card", () => {
   });
 
   it("should display correct error icon", () => {
+    const t = cloneDeep(toastInstanceExample);
     const wrapper = mount(
       <ToastCard
         type="negative"
@@ -61,6 +66,7 @@ describe("Toast Card", () => {
   });
 
   it("should display close icon", () => {
+    const t = cloneDeep(toastInstanceExample);
     const wrapper = mount(
       <ToastCard
         type="negative"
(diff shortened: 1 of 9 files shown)


hatched:
QA Notes:

The background of the toast is transparent
Screen Shot 2021-06-07 at 2 01 17 PM

There is no way to dismiss the toast. This is an issue when it covers up an action element.

When removing someone it doesn't appear to animate them closed any longer.
Screen Shot 2021-06-07 at 2 03 16 PM

src/components/ToastCard/ToastCard.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <ToastCard type="positive" message="I am a toast message" />
+    );
+    expect(wrapper.find("[data-type='positive']").exists()).toBe(true);
+  });
+
+  it("should display correct success icon", () => {
+    const wrapper = mount(
+      <ToastCard type="positive" message="I am a toast message" />
+    );
+    expect(wrapper.find("[data-type='positive']").exists()).toBe(true);
```
hatched, line 26: This doesn't check for the existence of an icon.

src/components/ToastCard/ToastCard.tsx
```diff
@@ -0,0 +1,32 @@
+import "./_toast-card.scss";
+
+type Props = {
+  type: string;
+  message: string;
+};
+
+export default function ToastCard({ type, message }: Props) {
+  let iconName;
```
hatched, line 9: Can you add a default here.

src/components/ToastCard/ToastCard.tsx
```diff
@@ -0,0 +1,32 @@
...
+      break;
+    case "negative":
+      iconName = "error";
+      break;
+    default:
+      break;
+  }
+  return (
+    <div className="toast-card" data-type={type}>
+      <i className={`p-icon--${iconName}`}>Success</i>
```
hatched, line 25: Without a default this would resolve to `p-icon--`

src/components/ToastCard/_toast-card.scss
```diff
@@ -0,0 +1,29 @@
+@import "vanilla-framework/scss/vanilla";
```
hatched, line 1: Vanilla has toast styles already. Did those not work for some reason?

hatched (changes requested):

src/components/ToastCard/_toast-card.scss
barrymcgee: As discussed, Vanilla does not have toast but notification styles which are different.

barrymcgee:
@hatched I've added a dismiss button to the toast notifications.

For the other bug, I've added a separate issue here: https://github.com/canonical-web-and-design/jaas-dashboard/issues/1038

src/components/ToastCard/ToastCard.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { mount } from "enzyme";
+
+import ToastCard from "./ToastCard";
+
+describe("Toast Card", () => {
+  const t = {
```
hatched, line 6: This object is shared between all of the test instances. While unlikely in this case it's a good idea to always clone the data for each test so that you don't end up with cascading or spurious failures. In this case it's an easy fix, just use the `cloneDeep` library we have and clone it for each test.

src/components/ToastCard/ToastCard.tsx
hatched: You could also use an enum (object) which would return the correct value for the class below....a lot less lines :)

src/components/ToastCard/_toast-card.scss
```diff
@@ -0,0 +1,35 @@
...
+
+.toast-card {
+  background-color: $color-x-light;
+  border: 1px solid $color-mid-light;
+  border-radius: 5px;
+  box-shadow: 0 0 2rem 0 rgba(51, 51, 51, 0.2);
```
hatched, line 7: Do we have a variable for a drop shadows throughout the app?

hatched (approved):

src/components/ToastCard/_toast-card.scss
barrymcgee: Seems `$box-shadow` is available in Vanilla 🤙🏻

PR #1029, opened 2021-06-06
Update Local Apps title to Local App

### LINKED ISSUE #937: Local Apps title should be Local App

satyampgt4:
## Done

update Local Apps title to Local App

## QA

- Pull code
- Open http://0.0.0.0:8036/
- [Add additional steps]

## Details

fix #937

### DIFF
--- a/src/tables/tableHeaders.tsx
+++ b/src/tables/tableHeaders.tsx
@@ -10,7 +10,7 @@ type HeaderRow = {
 };
 
 export const localApplicationTableHeaders: Header = [
-  { content: "local apps", sortKey: "local-apps" },
+  { content: "local app", sortKey: "local-app" },
   { content: "status", sortKey: "status" },
   { content: "version", sortKey: "version" },
   { content: "scale", className: "u-align--right", sortKey: "scale" },


barrymcgee:
@satyampgt4 Thank you for taking the time to propose this fix 🙏🏻

However, the fix is not as simple as it first seems as this `localApplicationTableHeaders` array is used in more than one place:

Screenshot 2021-06-07 at 09 09 24

Therefore, you'll need to either find a way to determine if the table has more than one row before pluralizing the header or if that adds too much complexity, it may be preferential in this case to simply add another table header definition to the `tableHeaders.tsx` to cater specifically for this single use case.

barrymcgee (changes requested):
@satyampgt4 Thank you for taking the time to propose this fix 🙏🏻

However, the fix is not as simple as it first seems as this `localApplicationTableHeaders` array is used in more than one place:

Screenshot 2021-06-07 at 09 09 24

Therefore, you'll need to either find a way to determine if the table has more than one row before pluralizing the header or if that adds too much complexity, it may be preferential in this case to simply add another table header definition to the `tableHeaders.tsx` to cater specifically for this single use case.

barrymcgee:
@satyampgt4 As this PR has turned stale, I'm going to close it for now.

PR #1045, opened 2021-06-10
Sharing view toggles on small screens

### LINKED ISSUE #1046: [Sharing] Add small screen view toggles
Screenshot 2021-06-10 at 12 15 53
Screenshot 2021-06-10 at 12 16 03

barrymcgee:
## Done

Sharing view toggles on small screens

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Decrease screen width to 500px wide
- Go to a model
- Click the sharing panel
- Verify that 'Add new user' button appears
- Click it.. to see form
- Verify that 'Back' button appears
- Click it to return to the sharing cards

Fixes #1046

### DIFF
--- a/src/components/Aside/_aside.scss
+++ b/src/components/Aside/_aside.scss
@@ -37,17 +37,10 @@
       .aside-split-col {
         width: 100%;
 
-        &:first-child {
-          border-bottom: 1px solid #ccc;
-          margin-bottom: 1rem;
-          padding-bottom: 1rem;
-        }
-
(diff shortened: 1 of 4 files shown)


hatched:
QA Note:
After adding a new user it would be nice if the 'add user' page would close and switch to the list.

hatched (approved):
While I'd like to see this done exclusively in CSS so that we could do some nicer transitions between the two views this is sufficient and functional for now.

PR #1050, opened 2021-06-14
Update button text from 'Share model' to 'Model access'

### LINKED ISSUE #1036: Finding "model access" in the gui is not intuitive
GUI: 0.7.1

I would suggest introducing "Model Access" instead of having to press "share model".

Its not intuitive and difficult to find that awesome feature.

image

barrymcgee:
## Done

Update button text from 'Share model' to 'Model access'

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to a model
- Verify button text under info panel is updated

## Details

Fixes #1036

### DIFF
--- a/src/pages/EntityDetails/Model/Model.test.js
+++ b/src/pages/EntityDetails/Model/Model.test.js
@@ -252,6 +252,24 @@ describe("Model", () => {
         </MemoryRouter>
       </Provider>
     );
-    expect(wrapper.find("Topology").length).toBe(1);
+    expect(wrapper.find("Topology").exists()).toBe(true);
+  });
+
+  it("should have a link for model access panel", () => {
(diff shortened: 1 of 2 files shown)


src/pages/EntityDetails/Model/Model.test.js
```diff
@@ -252,6 +252,24 @@ describe("Model", () => {
         </MemoryRouter>
       </Provider>
     );
-    expect(wrapper.find("Topology").length).toBe(1);
+    expect(wrapper.find("Topology").exists()).toBe(true);
+  });
+
+  it("should have a link for model access panel", () => {
```
hatched, line 258: Should probably test that this does what it's supposed to do on click.

hatched (approved):
Looks good, just would like expansion on the test here.

PR #1056, opened 2021-06-17
Share from models index page

barrymcgee:
## Done

Share from models index page on large and small screens

## QA

- Pull code
- Ensure you can open the correct model access panel from different screen widths

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1748,
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/1034

### DIFF
--- a/src/components/ModelTableList/CloudGroup.test.js
+++ b/src/components/ModelTableList/CloudGroup.test.js
@@ -87,5 +87,7 @@ describe("CloudGroup", () => {
     const firstContentRow = wrapper.find(".cloud-group tr").at(1);
     const modelAccessButton = firstContentRow.find(".model-access");
     expect(modelAccessButton.length).toBe(2);
+    expect(firstContentRow.find(".sm-screen-access-cell").exists()).toBe(true);
+    expect(firstContentRow.find(".lrg-screen-access-cell").exists()).toBe(true);
   });
 });
--- a/src/components/ModelTableList/CloudGroup.test.js
+++ b/src/components/ModelTableList/CloudGroup.test.js
@@ -69,4 +69,23 @@ describe("CloudGroup", () => {
     expect(tables.length).toBe(1);
     expect(tables.get(0).props.rows.length).toBe(3);
   });
+
+  it("model access buttons are present in cloud group", () => {
+    const store = mockStore(dataDump);
+    const filters = {
+      cloud: ["aws"],
+    };
+    const wrapper = mount(
+      <MemoryRouter>
+        <Provider store={store}>
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <CloudGroup filters={filters} />
+          </QueryParamProvider>
+        </Provider>
+      </MemoryRouter>
+    );
+    const firstContentRow = wrapper.find(".cloud-group tr").at(1);
+    const modelAccessButton = firstContentRow.find(".model-access");
+    expect(modelAccessButton.length).toBe(2);
+  });
 });
--- a/src/components/ModelTableList/CloudGroup.test.js
+++ b/src/components/ModelTableList/CloudGroup.test.js
@@ -1,7 +1,8 @@
-import { MemoryRouter } from "react-router";
+import { MemoryRouter, Route } from "react-router";
 import { mount } from "enzyme";
 import { Provider } from "react-redux";
 import configureStore from "redux-mock-store";
+import { QueryParamProvider } from "use-query-params";
 
 import CloudGroup from "./CloudGroup";
 
@@ -21,9 +22,13 @@ describe("CloudGroup", () => {
       },
     });
     const wrapper = mount(
-      <Provider store={store}>
-        <CloudGroup />
-      </Provider>
+      <MemoryRouter>
+        <Provider store={store}>
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <CloudGroup />
+          </QueryParamProvider>
+        </Provider>
+      </MemoryRouter>
     );
     const tables = wrapper.find("MainTable");
     expect(tables.length).toBe(0);
@@ -34,7 +39,9 @@ describe("CloudGroup", () => {
     const wrapper = mount(
       <MemoryRouter>
         <Provider store={store}>
-          <CloudGroup />
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <CloudGroup />
+          </QueryParamProvider>
         </Provider>
       </MemoryRouter>
     );
@@ -52,7 +59,9 @@ describe("CloudGroup", () => {
     const wrapper = mount(
       <MemoryRouter>
         <Provider store={store}>
-          <CloudGroup filters={filters} />
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <CloudGroup filters={filters} />
+          </QueryParamProvider>
         </Provider>
       </MemoryRouter>
     );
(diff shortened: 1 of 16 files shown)


hatched:
QA Notes:

- [ ] The button is way to large. 
- [ ] It pushes down the rows without an error message causing each row to jump around as you move your mouse.
- [ ] It should also vertically center in the row with the error message.
- [ ] Closing the share panel doesn't clear the 'model' query param from the url.
- [ ] Mobile view doesn't match the design.


Screen Shot 2021-06-17 at 10 15 51 AM
Screen Shot 2021-06-17 at 10 16 01 AM

Design:
Screen Shot 2021-06-17 at 10 31 50 AM

barrymcgee:
@hatched 

- I spent a couple of hours trying to centering the button but it isn't trivial. Flexbox doesn't apply to table cells. Absolutely positioning isn't reliable due to the variable heights of the rows. If it's a hard requirement, I'll make a separate issue that can be prioritised accordingly. 
- How does it not match the mobile design? If you mean, the overlapping status icon, I consider that outside the scope of this change.

barrymcgee:
Also.. I've added an animation transition on the row to make the increased space needed for the hover less jarring. However, the button itself will need to be made smaller if we want to move the increased row size entirely but this will affect the size of the touch target on small screens. It's currently 28px high, already falling well short of the recommended 44px. Thoughts @long-chung ?

hatched:
Sorry it appears that I had a stuck cache, the mobile view looks good.

As far as the height issue what I was thinking would be that the error rows would have an additional class on them so you could apply some margin to the button in those cases. It's not a big deal for now though it can be a follow-up.

I think that the button needs to be made smaller than the vanilla is-dense in order to match the design and prevent the non-error rows from jumping around.

src/components/PanelHeader/PanelHeader.tsx
```diff
@@ -1,19 +1,32 @@
...
+};
+
 export default function PanelHeader({ title }: Props): JSX.Element {
-  const [panelQs, setPanelQs] = useQueryParam("panel", StringParam);
+  const [panelQs, setPanelQs] = useQueryParams({
+    panel: StringParam,
+    model: StringParam,
+  });
+
+  // Remove all query strings when close button is clicked
```
hatched, line 18: I'm not 100% sure but I think `panelQs` will only have `panel` and `model` regardless of the rest of the query params. If so then this is working correctly and this comment just needs to be updated. If this isn't the case then you'll only want to clear those 2 params and not any other in there.

Can you verify?

src/components/PanelHeader/PanelHeader.tsx
```diff
@@ -1,19 +1,32 @@
...
   title: ReactNode;
 };
 
+type QueryString = {
+  [key: string]: string | undefined;
+};
+
 export default function PanelHeader({ title }: Props): JSX.Element {
-  const [panelQs, setPanelQs] = useQueryParam("panel", StringParam);
+  const [panelQs, setPanelQs] = useQueryParams({
```
hatched, line 13: Some new tests need to be added for this component to handle the fetching and resetting of the query params.

src/components/ModelTableList/shared.js
```diff
@@ -207,3 +207,27 @@ export function generateCloudAndRegion(model) {
...
+
+/**
+  Returns the model access button or an alternative value
+  @param {Function} setPanelQs A function to set query strings
+  @param {String} modelName the name of the model
+  @returns {Object} The markup for the table cell
+*/
+export function generateAccessButton(setPanelQs, modelName) {
+  return (
+    <>
```
hatched, line 219: This doesn't need the wrapper fragment.

src/components/ModelTableList/shared.js
```diff
@@ -207,3 +207,27 @@ export function generateCloudAndRegion(model) {
...
+      <button
+        onClick={() => {
+          setPanelQs({
+            model: modelName,
+            panel: "share-model",
+          });
+        }}
+        className="model-access  p-button--neutral is-dense"
```
hatched, line 227: two spaces :)

src/components/ModelTableList/StatusGroup.js
```diff
@@ -104,7 +111,7 @@ function generateModelTableDataByStatus(groupedModels) {
       const cloud = generateCloudCell(model);
       const credential = getStatusValue(model.info, "cloud-credential-tag");
       const controller = getStatusValue(model.info, "controllerName");
-      const lastUpdated = getStatusValue(model.info, "status.since");
+      const lastUpdated = getStatusValue(model.info, "status.since")?.slice(2);
```
hatched, line 114: Can you add a quick comment about what the format of the data looks like here before you slice it.

src/components/ModelTableList/StatusGroup.test.js
```diff
@@ -75,4 +86,23 @@ describe("StatusGroup", () => {
...
+        <Provider store={store}>
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <StatusGroup filters={filters} />
+          </QueryParamProvider>
+        </Provider>
+      </MemoryRouter>
+    );
+    const firstContentRow = wrapper.find(".status-group tr").at(1);
+    const modelAccessButton = firstContentRow.find(".model-access");
+    expect(modelAccessButton.length).toBe(2);
```
hatched, line 106: Can you add one more assertion that both of the columns are there? One for mobile and desktop?

src/components/ModelTableList/OwnerGroup.test.js
```diff
@@ -54,10 +61,31 @@ describe("OwnerGroup", () => {
...
+        <Provider store={store}>
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <OwnerGroup filters={filters} />
+          </QueryParamProvider>
+        </Provider>
+      </MemoryRouter>
+    );
+    const firstContentRow = wrapper.find(".owners-group tr").at(1);
+    const modelAccessButton = firstContentRow.find(".model-access");
+    expect(modelAccessButton.length).toBe(2);
```
hatched, line 89: Can you add an assertion that the extra column is here too?

src/components/ModelTableList/CloudGroup.test.js
```diff
@@ -52,12 +59,33 @@ describe("CloudGroup", () => {
...
+        <Provider store={store}>
+          <QueryParamProvider ReactRouterRoute={Route}>
+            <CloudGroup filters={filters} />
+          </QueryParamProvider>
+        </Provider>
+      </MemoryRouter>
+    );
+    const firstContentRow = wrapper.find(".cloud-group tr").at(1);
+    const modelAccessButton = firstContentRow.find(".model-access");
+    expect(modelAccessButton.length).toBe(2);
```
hatched, line 89: Can you add an assertion here for that extra row on mobile?

hatched (approved):
Looks good! Thanks for keeping up on this one.

Just need a few test assertions and if you could verify that one comment (or update it).

long-chung:
looks much better with the shallower button. There's enough clear space so a miss hit shouldn't be a problem and the hit area is still bigger than a standard text link so it should be ok

src/components/ModelTableList/shared.js
barrymcgee: Surprised Prettier didn't catch that.

src/components/PanelHeader/PanelHeader.tsx
barrymcgee: It will only have panel and model but I purposely did it like this on the assumption that other panels may need other query params so clearing the panel would clear them all? If you still think this should only clear the two query params, ping me and I'll change.

PR #1062, opened 2021-06-22
Conditionally render sharing model access button only if access

barrymcgee:
## Done

Conditionally render sharing model access button only if access

## QA

- Pull code
- Click to models you now you have owner/admin/write access on and verify model access button appears under the info panel
- Click to model you do not have the above permissions on and verify the button does not appear

## Details

Fixes https://github.com/canonical-web-and-design/app-squad/issues/112


hatched:
We'll want the same conditional behavior on the model list page so we should probably wait for that PR to land first.

hatched (approved):
Looks good, lets just be sure to add a similar conditional to the model list. That fn will probably need to be abstracted to a selector or util.

PR #1065, opened 2021-06-23
Restrict model index visibility to Access button on row hover

barrymcgee:
## Done

Restrict model index visibility to Access button on row hover

## QA

- Pull code
- Verify you can only open the access panel for those models you have the correct permissions for

## Details

Fixes https://github.com/canonical-web-and-design/juju-squad/issues/1779

### DIFF
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -411,3 +411,15 @@ export const formatFriendlyDateToNow = (date) => {
   });
   return formattedDate.concat(" ago");
 };
+
+export const canAdministerModelAccess = (userName, modelStatusData) => {
+  let hasPermission = false;
+  const modelUsers = modelStatusData?.info?.users;
+  const sharingAccess = ["admin", "write", "owner"];
+  modelUsers.forEach((userObj) => {
+    if (userObj.user === userName && sharingAccess.includes(userObj.access)) {
+      hasPermission = true;
+    }
+  });
+  return hasPermission;
+};
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -416,10 +416,11 @@ export const canAdministerModelAccess = (userName, modelStatusData) => {
   let hasPermission = false;
   const modelUsers = modelStatusData?.info?.users;
   const sharingAccess = ["admin", "write", "owner"];
-  modelUsers.forEach((userObj) => {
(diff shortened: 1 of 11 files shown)


src/app/utils/utils.js
```diff
@@ -411,3 +411,16 @@ export const formatFriendlyDateToNow = (date) => {
   });
   return formattedDate.concat(" ago");
 };
+
+export const canAdministerModelAccess = (userName, modelStatusData) => {
```
hatched, line 415: As this is a util lets change `modelStatusData` here to just be the `users` value as we don't use anything other than this.

Can you then also add a docbloc of what the values look like for both arguments.

src/components/ModelTableList/_model-table-list.scss
```diff
@@ -65,7 +65,7 @@
         width: 100%;
       }
 
-      .model-access-alt {
+      .has-permission .model-access-alt {
```
hatched, line 68: if the user `has-permission` then don't show it? :)

src/hooks/useActiveUser.tsx
```diff
@@ -0,0 +1,11 @@
+import { useSelector, useStore } from "react-redux";
+import { getActiveUserTag, getWSControllerURL } from "app/selectors";
+
+export default function useActiveUser() {
```
hatched, line 4: Nice.......but ;)

This will return the active user on the root controller for the application, not the controller that the model is on. While these CAN be the same they don't have to be, like when you register an additional controller. So I'm not sure that this is going to be the best approach for this problem as you need to query the active user for every row in the model list.

hatched (changes requested):

src/components/ModelTableList/_model-table-list.scss
barrymcgee: Don't show the model access button alternative? Yes, because we don't want to show the button _and_ the alternative if user has permission.

src/hooks/useActiveUser.tsx
barrymcgee: Created separate issue here: https://github.com/canonical-web-and-design/jaas-dashboard/issues/1066

hatched (approved):
Approved but lets make sure we do that follow-up right away so that these changes don't go too far down the rabbit hole.

PR #1067, opened 2021-06-25
Check for model users before calling info

barrymcgee:
## Done

Check for model users before calling info otherwise the app can crash after a hard refresh

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Sanity check code

Fixes https://github.com/canonical-web-and-design/app-squad/issues/143


anthonydillon (approved):
Seems reasonable 👍

PR #1082, opened 2021-07-08
Refactor the icon in the sharing heading

anthonydillon:
## Done
Remove the redefinition of the icon mixin
Moved the icon within the heading so to remove the positioning styling

## QA
- Check out the demo
- Go to a model and open the share panel
- See that the heading and icon look the same as they did before

## Screenshots
image


hatched (approved):

hatched:
Looks great, thanks for the cleanup.

PR #1099, opened 2021-07-26
Connect to the all model watcher when viewing a model details page 

hatched:
## Done

When viewing a model details page connect to the all model watcher.
When leaving the model details page, stop and disconnect from the all model watcher.

This is step 1 of many to convert this page from using the timer based poller to the watcher for displaying model data.

## QA

- View a model details page and then inspect the network tab, it should connect to the watcher and disconnect upon leaving.

## Details

Fixes: https://github.com/canonical-web-and-design/juju-squad/issues/1542

### DIFF
--- a/src/pages/EntityDetails/Model/Model.js
+++ b/src/pages/EntityDetails/Model/Model.js
@@ -47,7 +47,7 @@ import useActiveUser from "hooks/useActiveUser";
 
 import ChipGroup from "components/ChipGroup/ChipGroup";
 
-import { startModelWatcher } from "juju/index";
+import { startModelWatcher, stopModelWatcher } from "juju/index";
 
 import { renderCounts } from "../counts";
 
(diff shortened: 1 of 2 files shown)


src/pages/EntityDetails/Model/Model.js
```diff
@@ -78,6 +82,26 @@ const Model = () => {
     activeView: withDefault(StringParam, "apps"),
   });
 
+  const uuid = modelStatusData?.info?.uuid;
+
+  useEffect(() => {
+    let watcherHandle = null;
+    let conn = null;
+    async function startWatcher() {
+      ({ conn, watcherHandle } = await startModelWatcher(uuid, appState));
```
huwshimi, line 91: Is it possible that appState might not have the data you need at this point? I know there's a comment below about fixing this up, but if you end up with outdated data sent to `startModelWatcher` then it'll be because this useEffect was run before `appState` had all the data.

huwshimi (approved):
Looks good! One comment.

src/pages/EntityDetails/Model/Model.js
hatched: It's not clear here but for a `uuid` to be defined then the necessary in `appState` will be populated. This is opaque and the only way to know is to "know" :) But yeah as per the following comment this needs to be updated. Thanks for calling it out 👍

PR #1100, opened 2021-07-27
Type AllWatcher deltas and setup dispatch of the deltas.

hatched:
## Done

- Type the AllWatcher deltas .
- Dispatch the deltas.
- Placeholder file started for the `processDelta`.

## QA

This doesn't really do anything so no QA required. 
As long as the app doesn't explode when you use it.

## Details

Fixes: https://github.com/canonical-web-and-design/app-squad/issues/173

### DIFF
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -7,6 +7,7 @@ export interface ModelWatcherData {
 
 export type AllWatcherDelta =
   | ["action", "change", ActionChangeDelta]
+  | ["charm", "change", CharmChangeDelta]
   | ["unit", "change", UnitChangeDelta]
   | ["machine", "change", MachineChangeDelta]
   | ["model", "change", ModelChangeDelta]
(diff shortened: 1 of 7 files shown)


hatched:
https://jaas-dashboard-1100.demos.haus/

src/juju/types.ts
```diff
@@ -0,0 +1,217 @@
...
+    };
+  } | null; // lxdprofile
+  config?: { [key: string]: string | boolean };
+}
+
+interface MachineChangeDelta {
+  addresses: AddressData | null;
+  "agent-status": MachineAgentStatus;
+  "container-type": string;
+  "hardware-characteristics": HardwareCharacteristics | undefined;
```
huwshimi, line 99: Should this use a `?` instead of `| undefined` e.g. `"hardware-characteristics"?: HardwareCharacteristics`?

src/juju/watchers.ts
```diff
@@ -0,0 +1,24 @@
+import { AllWatcherDelta, ModelWatcherData } from "./types";
+
+export function processDeltas(
+  state: ModelWatcherData,
+  deltas: AllWatcherDelta[]
+) {
+  if (!deltas) {
+    return;
+  }
+  // console.log(deltas);
```
huwshimi, line 10: `console.log`

huwshimi (approved):
I think this looks pretty good. I can't say I understand all the changes but the code looks fine with one question.

src/juju/types.ts
hatched: Yep! good catch.

hatched:
Thanks @huwshimi! I've applied those changes to the follow-up branch already in progress as it's diverged a bit too much already.

PR #1110, opened 2021-08-10
Store and display model info delta

hatched:
## Done

Store and display the model info delta from the all watcher.

This branch disables the tests for the `Model` component as the component uses data from two different redux data sets. The new dataset is being setup using mocks via `Fishery`. Once the conversion to the mocks is complete then these tests can be re-enabled pointing to the new data.

## QA

- View a model's details page, does the left column show the model info fields.


src/pages/EntityDetails/Model/Model.test.js
```diff
@@ -15,14 +17,36 @@ jest.mock("components/Topology/Topology", () => {
 
 const mockStore = configureStore([]);
 
-describe("Model", () => {
+describe.skip("Model", () => {
```
hatched, line 20: Skipped test suite, see the comment in the PR description.

src/testing/redux-factory.ts
```diff
@@ -0,0 +1,96 @@
+import { Factory } from "fishery";
```
hatched, line 1: This is the first pass setting up the mock for the model watcher data for the model info data. This is not currently being used but was tested as functional with the (now disabled) Model tests.

src/juju/watchers.ts
```diff
@@ -1,24 +1,57 @@
...
-      case "unit":
+      case "model":
         switch (delta[1]) {
           case "change":
-            // Handle unit change delta
-            console.log(modelUUID); // temp quiet the linter for modelUUID :)
+            if (!modelWatcherData[modelUUID]) {
+              modelWatcherData[modelUUID] = generateModelWatcherBase();
+            }
+            mergeWith(modelWatcherData[modelUUID].model, delta[2]);
```
anthonydillon, line 50: Could you add a comment about this index on the delta array.

src/pages/EntityDetails/Model/Model.js
```diff
@@ -69,8 +71,16 @@ const shouldShow = (segment, activeView) => {
...
+  if (cloudTag && region) {
+    return `${extractCloudName(cloudTag)} / ${region}`;
+  }
+  return "";
+};
+
 const Model = () => {
-  const appState = useStore().getState();
+  const store = useStore();
+  const appState = store.getState();
```
anthonydillon, line 83: What's the point in this. Doesn't look like you use store anywhere else.

anthonydillon (commented):
Code looks ok but a few comments.

src/pages/EntityDetails/Model/Model.js
hatched: Good catch, that was just a change from some investigation that ended up not being used. I've reverted that change.

src/juju/watchers.ts
hatched: done 👍

src/pages/EntityDetails/Model/Model.js
```diff
@@ -153,17 +169,8 @@ const Model = () => {
...
-  const cloudProvider = modelStatusData
-    ? extractCloudName(modelStatusData.model["cloud-tag"])
-    : "";
-
-  const ModelEntityData = {
-    controller: modelStatusData?.model.type,
-    "Cloud/Region": `${cloudProvider} / ${modelStatusData?.model.region}`,
-    version: modelStatusData?.model.version,
-    sla: modelStatusData?.model.sla,
-    provider: modelStatusData?.info?.["provider-type"],
```
Caleb-Ellis, line 165: It seems like this is no longer being represented in the model details page. Is that intentional?

src/juju/watchers.ts
```diff
@@ -1,24 +1,58 @@
...
-  state: ModelWatcherData,
+  modelWatcherData: ModelWatcherData,
   deltas: AllWatcherDelta[]
 ) {
   if (!deltas) {
     return;
   }
-  // console.log(deltas);
   deltas.forEach((delta) => {
+    // The delta messages are in the format [entity, action, data].
```
Caleb-Ellis, line 42: You could destructure the array so you wouldn't need to refer to this comment if you were further down the file, say.
``` typescript
const [entity, action, data] = delta;
```

src/pages/EntityDetails/Model/Model.js
Caleb-Ellis, line 165: It seems like this is no longer being represented in the model details page. Is that intentional?

Caleb-Ellis (approved):
Just one question about the model provider but otherwise LGTM!

src/juju/watchers.ts
hatched: Good call I'll do this in the follow-up as I've already moved on from this one :)

src/pages/EntityDetails/Model/Model.js
hatched: Yes sorry I meant to add a comment about this. The `provider` is an internal implementation detail of Juju and is a 1:1 mapping to the cloud provider so knowing what the provider name is is irrelevant.

src/juju/watchers.ts
hatched: In doing this as a follow-up I found what I believe to be a TypeScript 'bug'...

I added another case for the `charm` deltas as follows...

```ts
...
      case "charm":
        switch (delta[1]) {
          case "change":            
            const formatted = {
              [delta[2]["charm-url"]]: delta[2],
            };
            mergeWith(modelWatcherData[modelUUID].charms, formatted);
            break;
        }
        break;
...
```

now changing `delta[2]` to `data` as per `const [entityType, actionType, data] = delta`

```ts
...
            const formatted = {
              [delta[2]["charm-url"]]: delta[2],
            };
...
```

gives the TS error...

```
Element implicitly has an 'any' type because expression of type '"charm-url"' can't be used to index type 'ModelChangeDelta | ActionChangeDelta | ApplicationChangeDelta | CharmChangeDelta | UnitChangeDelta | MachineChangeDelta | RelationChangeDelta'.
  Property 'charm-url' does not exist on type 'ModelChangeDelta | ActionChangeDelta | ApplicationChangeDelta | CharmChangeDelta | UnitChangeDelta | MachineChangeDelta | RelationChangeDelta'.ts(7053)
```

So it appears that it cannot correctly infer the available data if it's been deconstructed as it only thinks it has a single key, `model-uuid`.

PR #1112, opened 2021-08-11
Updates docker yarn run to use --production

petesfrench:
## Done

- Updates docker yarn run to use --production

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check nothing is broken


Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt update && \
     curl -fsSL https://deb.nodesource.com/setup_16.x | bash - && \
     apt update && \
     apt install nodejs --yes
-RUN npm install -g yarn
+RUN npm install -g yarn --production
```
hatched, line 11: This change is a noop. the `--production` flag only does something when it's installing from a package.json. You probably meant to add this to line 15 instead.

hatched (changes requested):

Dockerfile
hatched: Sorry ignore that previous comment, this is still a noop but if you move it to 15 then it won't have the necessary dependencies to actually build the project :)

petesfrench:
@hatched Hi, I moved the production tag to the other line as per your suggestion and played around with the dependencies to get it to build. I think it should work fine now?

hatched (approved):

hatched:
Thanks for getting this working it's much faster to build.

After some digging the reason these random deps had to be moved up is because CRA runs lint on build.

PR #1115, opened 2021-08-16
Store the remaining change deltas in the redux store.

hatched:
## Done

Store remaining change deltas in the redux store.

## QA

Open up an existing model and inspect the redux store for juju/modelWatcherData and ensure that it has all of the top level entity keys... action, application, charm, machine, model, relation, unit.


src/juju/watchers.ts
```diff
@@ -39,19 +45,79 @@ export function processDeltas(
     return;
   }
   deltas.forEach((delta) => {
-    // The delta messages are in the format [entity, action, data].
+    // Delta is in the format [entityType, actionType, data].
```
anthonydillon, line 48:
```suggestion
    // Delta is in the format of [entityType, actionType, data].
```

anthonydillon: Or you could map it to these object names to make the code more readable?

src/juju/watchers.ts
```diff
@@ -39,19 +45,79 @@ export function processDeltas(
...
     const modelUUID = delta[2]["model-uuid"];
+    if (!modelWatcherData[modelUUID]) {
+      modelWatcherData[modelUUID] = generateModelWatcherBase();
+    }
     switch (delta[0]) {
+      case "action":
+        switch (delta[1]) {
```
anthonydillon, line 55: Nested switch feel wrong and will grow messy. Why not check and switch the `entityType` function handlers with switch/case?

anthonydillon (approved):
LGTM :+1: with some questions/suggestions in the code

src/juju/watchers.ts
hatched: I actually thought the exact same thing :) As there are a few other delta types to add I was going to wait until those ones are completed and then look for a way to clean this up. I expect that I'll have a couple utility functions that are called instead of this repetition.

src/juju/watchers.ts
hatched: I applied your change. I had also attempted using destructuring but there were issues with TypeScript. See the discussion [here](https://github.com/canonical-web-and-design/jaas-dashboard/pull/1110#discussion_r688669051) for more information.

hatched:
Thanks for the review!

PR #1117, opened 2021-08-16
Add missing deltas and simplify delta parsing

hatched:
## Done

- Add support for the "remove" deltas.
- Simplify delta parsing code.

## QA

- The redux store should have all data available for the model
- When you remove an application it should also remove it from the store.

### DIFF
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -43,6 +43,23 @@ export type AllWatcherDelta =
   | ["relation", "change", RelationChangeDelta]
   | ["relation", "remove", RelationChangeDelta];
 
+export type DeltaEntityTypes =
+  | "actions"
+  | "applications"
+  | "charms"
+  | "machines"
+  | "relations"
+  | "units";
+
+export type DeltaMessageData =
+  | ActionChangeDelta
+  | ApplicationChangeDelta
+  | CharmChangeDelta
+  | UnitChangeDelta
+  | MachineChangeDelta
+  | ModelChangeDelta
+  | RelationChangeDelta;
+
 export interface ModelData {
   actions: ActionData;
   applications: ApplicationData;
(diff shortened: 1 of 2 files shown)


src/juju/types.ts
```diff
@@ -32,11 +32,33 @@ export interface UnitData {
...
   | ["unit", "change", UnitChangeDelta]
+  | ["unit", "remove", UnitChangeDelta]
   | ["machine", "change", MachineChangeDelta]
+  | ["machine", "remove", MachineChangeDelta]
   | ["model", "change", ModelChangeDelta]
-  | ["relation", "change", RelationChangeDelta];
+  | ["relation", "change", RelationChangeDelta]
+  | ["relation", "remove", RelationChangeDelta];
+
+export type DeltaEntityTypes =
```
huwshimi, line 46: These could be more useful as an enum that you could then use when defining the `AllWatcherDelta` tuples above.

src/juju/watchers.ts
```diff
@@ -37,6 +43,23 @@ function generateModelWatcherBase(): ModelData {
...
 }
 
+function _processDelta(
+  actionType: string,
+  delta: DeltaMessageData,
+  modelData: ModelData,
+  entityType: DeltaEntityTypes,
+  key: string
+): void {
+  if (actionType === "change") {
```
huwshimi, line 53: "change", "remove" etc. could also be an enum that you use here and when defining the `AllWatcherDelta` tuples.

src/juju/watchers.ts
```diff
@@ -50,74 +73,30 @@ export function processDeltas(
...
-            mergeWith(modelWatcherData[modelUUID].units, formatted);
-            break;
-        }
-        break;
+    const modelData = modelWatcherData[modelUUID];
+    const _process = _processDelta.bind(null, delta[1], delta[2], modelData);
+    if (delta[0] === "action") {
+      _process("actions", delta[2].id);
+    }
+    if (delta[0] === "application") {
```
huwshimi, line 81: `delta[0]` can't equal multiple entity types, so these could all be `else if`. Also here's another spot you could use the enums.

huwshimi (approved):
Looking good. Just a few small suggestions.

PR #1120, opened 2021-08-17
Persist the connection for the model watcher when viewing the model details

hatched:
## Done

- When viewing the model details pages persist the same instance of the model watcher connection.
- Disconnect the model watcher connection when routing away from a model details page.
- The routing needed to be nested so that we have a persistent view for all of the model details pages as a place to put the watcher management code.

## QA

- Open the network tab and view the model list, once it goes through all of your models then clear the network list and view a model.
- It should connect the watcher, click the element in the network pane to see the frames.
- Navigate away from the model details view and you should see the connection terminate.


src/components/Routes/TestRoute.tsx
```diff
@@ -1,16 +1,10 @@
...
 
-import { paths } from "./Routes";
-
 type Props = {
   path: string;
   children: JSX.Element;
 };
 
 export default function TestRoute({ path, children }: Props) {
-  // Validate paths
```
hatched, line 11: An issue was created to re-implement this check.

src/juju/index.js
```diff
@@ -75,7 +75,7 @@ function determineLoginParams(credentials, identityProviderAvailable) {
 
 function startPingerLoop(conn) {
   // Ping to keep the connection alive.
-  const intervalId = setInterval(() => {
+  const intervalId = window.setInterval(() => {
```
hatched, line 78: This fixes an issue with TS types as the node.js version of `setInterval` returns a `Timer` object whereas the browser only returns an integer.

src/pages/EntityDetails/Model/Model.js
```diff
@@ -94,39 +91,6 @@ const Model = () => {
 
   const uuid = modelStatusData?.info?.uuid;
 
-  useEffect(() => {
```
hatched, line 97: This was moved to the `ModelDetails` component.

PR #1124, opened 2021-08-25
Convert the model details pages to use the new watcher data

hatched:
## Done

The data on the model details pages is now being shown from the watcher data

## QA

- View a model and play around, make sure nothing breaks.

### DIFF
--- a/src/components/EntityInfo/EntityInfo.tsx
+++ b/src/components/EntityInfo/EntityInfo.tsx
@@ -1,7 +1,7 @@
 import "./_entity-info.scss";
 
 type Props = {
-  data: { [key: string]: string | JSX.Element };
+  data: { [key: string]: string | number | JSX.Element };
 };
 
 export default function EntityInfo({ data }: Props): JSX.Element {
--- a/src/components/EntityInfo/EntityInfo.tsx
+++ b/src/components/EntityInfo/EntityInfo.tsx
@@ -1,7 +1,9 @@
+import type { ReactNode } from "react";
+
 import "./_entity-info.scss";
 
 type Props = {
-  data: { [key: string]: string | number | JSX.Element };
+  data: { [key: string]: ReactNode };
 };
 
 export default function EntityInfo({ data }: Props): JSX.Element {
--- a/src/juju/model-selectors.ts
+++ b/src/juju/model-selectors.ts
@@ -5,6 +5,7 @@ import type {
   ApplicationData,
   ModelInfo,
   ModelWatcherData,
+  RelationData,
   UnitData,
 } from "./types";
 
@@ -65,6 +66,18 @@ export function getModelUnits(modelUUID: string) {
     (modelWatcherData): UnitData | null => {
       if (modelWatcherData) {
         return modelWatcherData.units;
+      }
+      return null;
+    }
+  );
+}
+
+export function getModelRelations(modelUUID: string) {
+  return createSelector(
+    getModelWatcherDataByUUID(modelUUID),
+    (getModelWatcherData): RelationData | null => {
+      if (getModelWatcherData) {
+        return getModelWatcherData.relations;
       }
       return null;
     }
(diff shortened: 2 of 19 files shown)


src/components/EntityInfo/EntityInfo.tsx
```diff
@@ -1,7 +1,7 @@
 import "./_entity-info.scss";
 
 type Props = {
-  data: { [key: string]: string | JSX.Element };
+  data: { [key: string]: string | number | JSX.Element };
```
huwshimi, line 4: It looks like you could probably just use `data: { [key: string]: ReactNode };`.

src/juju/model-selectors.ts
```diff
@@ -1,27 +1,43 @@
...
   state.juju.modelWatcherData;
 const getModelList = (state: ReduxState): ModelsList => state.juju.models;
 
-export function getModelInfoByUUID(modelUUID: string) {
+export function getModelWatcherDataByUUID(modelUUID: string) {
+  return createSelector(getModelWatcherData, (modelWatcherData) => {
+    if (modelWatcherData[modelUUID]) {
+      return modelWatcherData[modelUUID];
+    }
+    return null;
```
huwshimi, line 22: This could be `return modelWatcherData[modelUUID] ?? null` if you want to be fancy.

src/juju/model-selectors.ts
```diff
@@ -32,3 +48,51 @@ export function getModelUUIDByName(modelName: string, ownerName: string) {
...
 }
+
+export function getModelApplications(modelUUID: string) {
+  return createSelector(
+    getModelWatcherDataByUUID(modelUUID),
+    (modelWatcherData): ApplicationData | null => {
+      if (modelWatcherData) {
+        return modelWatcherData.applications;
+      }
+      return null;
```
huwshimi, line 59: Again with all these you could save some precious lines with `return modelWatcherData?.applications ?? null`.

src/juju/watchers.ts
```diff
@@ -86,7 +87,13 @@ export function processDeltas(
     if (delta[0] === DeltaEntityTypes.ACTION) {
       _process(ReduxDeltaEntityTypes.ACTIONS, delta[2].id);
     } else if (delta[0] === DeltaEntityTypes.APPLICATION) {
-      _process(ReduxDeltaEntityTypes.APPLICATIONS, delta[2].name);
+      const formatted: ApplicationData = {
+        [delta[2].name]: delta[2],
```
huwshimi, line 91: These would probably be a lot easier to follow with some more descriptive names. You could use a destructure in the forEach, something like `deltas.forEach(([action, change, delta]) => {`.

src/juju/watchers.ts
```diff
@@ -99,6 +106,28 @@ export function processDeltas(
...
+        const applicationName = key.split("/")[0];
+        if (applicationUnitCounts[applicationName] === undefined) {
+          applicationUnitCounts[applicationName] = 0;
+        }
+        applicationUnitCounts[applicationName] += 1;
+      });
+      Object.entries(applicationUnitCounts).forEach(
+        ([applicationName, count]) => {
+          if (!modelData.applications[applicationName]) {
+            // @ts-ignore
```
huwshimi, line 125: Probably worth leaving a comment for why this is necessary.

src/pages/EntityDetails/App/App.test.js
```diff
@@ -20,7 +20,7 @@ jest.mock("components/Topology/Topology", () => {
   return Topology;
 });
 
-describe("Entity Details App", () => {
+describe.skip("Entity Details App", () => {
```
huwshimi, line 23: :eye: :nose: :eye:

huwshimi (approved):
I think this looks pretty good. Just some small suggestions that you can ignore if you wish.

src/juju/model-selectors.ts
hatched: Got to save those precious bytes!

src/juju/watchers.ts
hatched: Unfortunately I'm not able to destructure due to a bug in TS. The discussion is in one of the previous PR's here.

PR #1131, opened 2021-09-01
Generate the app status from an aggregate of the unit statuses

hatched:
## Done

- The all watcher doesn't contain an app status so we have to generate an aggregate app status from parsing the units.
- Remove the `refreshModel` call as a driveby from the webCLI as it's no longer needed now that the model details page uses the watcher data.

## QA

- Contact me for access to a model to QA this with.
- It should show the application status on the Application list, Machine details page, Unit details page.

## Screenshot

Screen Shot 2021-09-01 at 11 36 08 AM

### DIFF
--- a/src/juju/reducer.js
+++ b/src/juju/reducer.js
@@ -86,6 +86,9 @@ export default function jujuReducer(state = defaultState, action) {
         break;
       case actionsList.populateMissingAllWatcherData:
         if (!draftState.modelWatcherData?.[payload.uuid]?.model) {
+          if (!draftState.modelWatcherData?.[payload.uuid]) {
+            draftState.modelWatcherData[payload.uuid] = {};
+          }
           draftState.modelWatcherData[payload.uuid].model = {};
         }
         mergeWith(draftState.modelWatcherData[payload.uuid].model, {
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -17,7 +17,7 @@ export function generateEntityIdentifier(charmId, name, subordinate) {
 }
 
 export const generateStatusElement = (
-  status,
+  status = "unknown",
   count,
   useIcon = true,
   actionsLogs = false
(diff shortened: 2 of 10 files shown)


huwshimi:
This looks fine for some models:

Screen Shot 2021-09-02 at 9 05 59 am

But for some models (I think that have apps with no units) it isn't right:

Screen Shot 2021-09-02 at 9 04 38 am

hatched:
Good find, thanks @huwshimi that is resolved now.

huwshimi:
QA is good now, thanks @hatched

Screen Shot 2021-09-03 at 9 31 42 am

src/juju/reducer.js
```diff
@@ -86,6 +86,9 @@ export default function jujuReducer(state = defaultState, action) {
         break;
       case actionsList.populateMissingAllWatcherData:
         if (!draftState.modelWatcherData?.[payload.uuid]?.model) {
+          if (!draftState.modelWatcherData?.[payload.uuid]) {
+            draftState.modelWatcherData[payload.uuid] = {};
+          }
           draftState.modelWatcherData[payload.uuid].model = {};
         }
```
Caleb-Ellis, line 93: This combination of negative logic and optional chaining is a bit hard to follow. I wonder if this could be rewritten in the positive, or some descriptive intermediary variables could be used like:
``` typescript
const existingUUID = draftState.modelWatcherData?.[payload.uuid];
const existingModel = existingUUID?.model;

if (!existingModel) {
  if (!existingUUID) {
    draftState.modelWatcherData[payload.uuid] = {};
  }
  draftState.modelWatcherData[payload.uuid].model = {};
}
```

Caleb-Ellis (approved):
LGTM! Just one comment on a bit of hard-to-follow logic

src/juju/reducer.js
hatched: Good call, in the follow-up I've got some changes here so I'll apply this there to avoid conflicts.

PR #1135, opened 2021-09-02
Convert EntityDetails to use the model watcher data

hatched:
## Done

Convert the EntityDetails component to use the model watcher data.

## QA

- If you can view a model details then this is working as expected.


anthonydillon (approved):
Code and QA look good to me 👍

PR #1142, opened 2021-09-09
Convert the InfoPanel and Topology components to new watcher data structure

hatched:
## Done

- Converted the InfoPanel and Topology components to use the new watcher data structure.
- Added a delta handler for the annotation delta events.

## QA

Make sure that the side InfoPanel is shown with the topology.

### DIFF
--- a/src/components/InfoPanel/InfoPanel.test.js
+++ b/src/components/InfoPanel/InfoPanel.test.js
@@ -18,16 +18,7 @@ jest.mock("components/Topology/Topology", () => {
 
 describe("Info Panel", () => {
   it("renders the topology", () => {
-    const mockState = reduxStateFactory().build(
-      {},
-      {
-        transient: {
-          models: [
-            { name: "enterprise", owner: "kirk@external", type: "kubernetes" },
-          ],
-        },
-      }
-    );
+    const mockState = reduxStateFactory().build();
     const store = mockStore(mockState);
     render(
       <Provider store={store}>
@@ -44,16 +35,7 @@ describe("Info Panel", () => {
   });
 
   it("renders the expanded topology on click", () => {
-    const mockState = reduxStateFactory().build(
-      {},
-      {
-        transient: {
-          models: [
-            { name: "enterprise", owner: "kirk@external", type: "kubernetes" },
-          ],
-        },
-      }
-    );
+    const mockState = reduxStateFactory().build();
     const store = mockStore(mockState);
     render(
       <Provider store={store}>
--- a/src/components/InfoPanel/InfoPanel.test.js
+++ b/src/components/InfoPanel/InfoPanel.test.js
@@ -1,10 +1,10 @@
+import { render, screen } from "@testing-library/react";
 import configureStore from "redux-mock-store";
 import { Provider } from "react-redux";
-import { mount } from "enzyme";
 import { MemoryRouter, Route } from "react-router";
 import { QueryParamProvider } from "use-query-params";
-import dataDump from "testing/complete-redux-store-dump";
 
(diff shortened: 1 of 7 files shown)


src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -1,29 +1,37 @@
+import { render, screen } from "@testing-library/react";
 import configureStore from "redux-mock-store";
 import { Provider } from "react-redux";
-import { mount } from "enzyme";
```
anthonydillon, line 3: Was it difficult to make the switch?

src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -1,29 +1,37 @@
...
+import { reduxStateFactory } from "testing/redux-factory";
 import TestRoute from "components/Routes/TestRoute";
 
 import InfoPanel from "./InfoPanel";
 
 const mockStore = configureStore([]);
 
 jest.mock("components/Topology/Topology", () => {
-  const Topology = () => <div className="topology"></div>;
+  const Topology = () => <div data-testid="topology"></div>;
```
anthonydillon, line 15: I believe the standard is `data-test`

src/components/InfoPanel/InfoPanel.test.js
```diff
@@ -32,16 +40,24 @@ describe("Info Panel", () => {
...
+            { name: "enterprise", owner: "kirk@external", type: "kubernetes" },
+          ],
+        },
+      }
+    );
+    const store = mockStore(mockState);
+    render(
       <Provider store={store}>
-        <MemoryRouter
-          initialEntries={["/models/user-eggman@external/group-test"]}
```
anthonydillon, line 43: Bye bye user-eggman :wave:

anthonydillon (commented):
Few comments in the code

src/components/InfoPanel/InfoPanel.test.js
hatched: Honestly it was trivial. The only issue being that we have to have a mindset change from querying by class name to things like `role`, `text`, or the escape hatch of `testid`.

src/components/InfoPanel/InfoPanel.test.js
hatched: Yeah that is(was?) the standard but RTL has a method `getByTestId` in which it requires `data-testid` vs `data-test`.

src/components/InfoPanel/InfoPanel.test.js
hatched: Poor eggman, he has been dethroned by Captain Kirk.

anthonydillon:
You are missing provider from the infopanel:
image

src/components/InfoPanel/InfoPanel.test.js
anthonydillon: I see, good to know.

anthonydillon: Better to follow the testing frameworks standard then ours

anthonydillon (changes requested):
QA issue above

hatched:
Oh good catch :D I actually removed that as it's an implementation detail of Juju. It's dependent on the cloud that you're on so the user knowing what the provider is called is irrelevant. 

The data is also not available in the watcher so we'd have to make extra requests for it.

anthonydillon:
+1 after a discussion on mm

anthonydillon (approved):

hatched:
Thanks, much appreciated 👍

PR #1152, opened 2021-09-29
Add support for relative controller api endpoint paths.

hatched:
## Done

Added support for relative controller API endpoint paths. This is necessary because the charm can provide a relative path when it's proxying the connections to the same host. This allows the dashboard to connect to the correct endpoint even when we don't know the ultimate endpoint of the proxy.

## QA

- The app should continue to work as expected.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -55,6 +55,17 @@ checkConfigExists();
 
 function bootstrap() {
   const config = window.jujuDashboardConfig;
+  // It's possible that the charm is generating a relative path for the
+  // websocket because it is providing the API on the same host as the
+  // application assets.
+  // If we were provided with a relative path for the endpoint then we need
+  // to build the full correct path for the websocket to connect to.
+  const controllerAPIEndpoint = config.controllerAPIEndpoint;
+  if (controllerAPIEndpoint.indexOf("://") === -1) {
+    const protocol =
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -55,6 +55,17 @@ checkConfigExists();
 
 function bootstrap() {
   const config = window.jujuDashboardConfig;
+  // It's possible that the charm is generating a relative path for the
+  // websocket because it is providing the API on the same host as the
+  // application assets.
+  // If we were provided with a relative path for the endpoint then we need
+  // to build the full correct path for the websocket to connect to.
+  const controllerAPIEndpoint = config.controllerAPIEndpoint;
+  if (controllerAPIEndpoint.indexOf("://") === -1) {
```
anthonydillon, line 64: Can you use `includes`? It's clearer.

anthonydillon (changes requested):
Simple suggestion in the code. Should there be a test for this?

src/index.js
hatched: Good call.

hatched:
There should definitely be some tests, unfortunately this file isn't tested at all so we'll have to swing back around and add some.

jpmartinspt (approved):

PR #1165, opened 2021-10-21
Revert to the older beta of the jujulib as beta7 only supports Juju 3.0

hatched:
## Done

The previous renovate update bumped the jujulib version. That version only supports Juju 3.0 so we had to roll it back to beta6.

## QA

- The app should continue to work as expected.


hatched:
Thanks!

Issue #1170, opened 2021-11-03
Cannot open the dashboard after `juju upgrade-model`


heitorPB:
**Describe the bug**

I keep seeing the error `Something has gone wrong. If this issue persists, please raise an issue on GitHub.`

**To Reproduce**
1. Go to jimm.xxxxxxxx.com
2. Click on any model
3. See error

**Expected behavior**
No error

**Additional context**
I `juju upgrade-model` and after that I can't open it in the Juju Dashboard.

heitorPB:
This error only appears to the models that i have `juju upgrade-model`. The models that I did not run the command work fine.

hatched:
Hmm I'm not sure what this could be, what versions did you go between? So that I can try and reproduce the problem.

heitorPB:
Upgraded from 2.8.x to 2.9.16:

Our controller was upgraded to 2.9.16
Then dashboard installed and all models worked.
Then I upgraded some models to 2.9.16
Then those models don't open in the dashboard anymore.

hatched:
Thanks, I'll try and repro locally shortly and get back to you as off the top of my head there should be no difference between connecting to those different model versions. If you wouldn't mind looking in the browser console (ctrl+shift+i) and pasting here any error messages that you see when this happens.

heitorPB:
There's this error here:

```
instrument.ts:129 DOMException: Failed to construct 'WebSocket': The URL 'null' is invalid.
    at Ls.connect (https://jimm.example.com/static/js/main.5d3f7958.chunk.js:1:84689)
    at https://jimm.example.com/static/js/main.5d3f7958.chunk.js:1:85805
    at Ns (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:685299)
    at t.unstable_runWithPriority (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:703110)
    at Wi (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:625470)
    at Cs (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:684760)
    at https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:684671
    at B (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:702169)
    at MessagePort._.port1.onmessage (https://jimm.example.com/static/js/2.a32b748d.chunk.js:2:700891)
```

[...]

PR #1181, opened 2021-11-24
gracefully fail if the web cli is not passed the necessary data to construct the websocket address.

hatched:
## Done

If the WebCLI component isn't passed the necessary data to contruct the websocket address then do not try and render the WebCLI. This is a bandaid for the root cause, which can be seen in #1170.

## QA

- Contact me to get access to a known broken model.
- View the demo link and view the model details, the app shouldn't 💥 .


Caleb-Ellis (approved):
Can confirm: 
- App breaks in production when viewing a model without a ws address
- App does not break in this branch when viewing a model without a ws address

hatched:
Thanks 👍

PR #1193, opened 2021-12-07
If no description is provided for an action don't try and render one.

### LINKED ISSUE #1164: Run Action crash for Nextcloud charm
**Describe the bug**
When trying to use "Run Action" from nextcloud (Using 0.8.1 dashboard)

https://charmhub.io/nextcloud  (edge 611d74d)

**To Reproduce**
1. Go to 'Application' for nextcloud charm'
2. Click on 'Run Action'

**Expected behavior**
The right hand pane opens with actions available.

**Screenshots**
image

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser: Chrome 
 - Version Version 94.0.4606.71 (Official Build) (64-bit)

hatched:
## Done

Support actions without a description

## QA

- Ask me for access to a model with a charm with actions without descriptions.

## Details

Fixes #1164

### DIFF
--- a/src/components/RadioInputBox/DescriptionSummary.tsx
+++ b/src/components/RadioInputBox/DescriptionSummary.tsx
@@ -1,5 +1,5 @@
 type Props = {
-  description: string;
+  description: string | undefined;
 };
 
 export default function DescriptionSummary({ description }: Props) {
--- a/src/components/RadioInputBox/DescriptionSummary.tsx
+++ b/src/components/RadioInputBox/DescriptionSummary.tsx
@@ -2,9 +2,10 @@ type Props = {
   description: string;
 };
 
-export default function DescriptionSummary({
-  description,
-}: Props): JSX.Element {
+export default function DescriptionSummary({ description }: Props) {
+  if (!description) {
+    return null;
+  }
   // 30 is a magic number, the width of the available text area of the field
   // If the width of the actions area increases then this number will need
   // to be adjusted accordingly.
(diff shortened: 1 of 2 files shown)


src/components/RadioInputBox/DescriptionSummary.tsx
```diff
@@ -2,9 +2,10 @@ type Props = {
   description: string;
```
Caleb-Ellis, line 2: The `Props` type should be updated to reflect what `description` can be

src/components/RadioInputBox/DescriptionSummary.tsx
```diff
@@ -2,9 +2,10 @@ type Props = {
...
 };
 
-export default function DescriptionSummary({
-  description,
-}: Props): JSX.Element {
+export default function DescriptionSummary({ description }: Props) {
```
Caleb-Ellis, line 5: The `Props` type should be updated to include the falsy types that `description` can be. In MAAS we typically refer to the actual API type itself (e.g. `Machine["name"]` instead of `string`) so that any changes to the shape of the API are known by all the components that consume it.

Caleb-Ellis (approved):
LGTM! Just one comment on types.

src/components/RadioInputBox/DescriptionSummary.tsx
hatched: Oops, that'll teach me to be distracted :) Updated the type.

Yeah the API structure doesn't change, but what I want to do is import the types from libjuju just haven't yet had the time yet.

PR #1225, opened 2022-03-29
Revert "Update dependency framer-motion to v6"

anthonydillon:
Reverts canonical-web-and-design/jaas-dashboard#1207

This upgrade is causing a `Can't import the named export 'Children' from non EcmaScript module (only default export is available)` error when building.


petermakowski (approved):
Sorry, my bad! Thanks for fixing this 🙏

PR #1226, opened 2022-03-29
Parse and load the cs: icons from charmhub

anthonydillon:
## Done
If loading a `cs:` charm. Parse and generate the charmhub icon URL to load the image from 

## QA
- Pull code
- Open http://localhost:3001/
- Go to a model and see the icons load from charmhub.io
- _Not that etcd is not in charmhub and so will 404_

## Details
Fixes https://github.com/canonical-web-and-design/jaas-dashboard/issues/1210

## Screenshots
image

### DIFF
--- a/src/app/utils/utils.js
+++ b/src/app/utils/utils.js
@@ -249,8 +249,32 @@ export const generateIconPath = (charmId) => {
     return defaultCharmIcon;
   }
   if (charmId.indexOf("cs:") === 0) {
+    // Strip unnessesary prefixes
+    // before: cs:~containers/lxd-container-47
+    // after:  containers/lxd-container-47
+    charmId = charmId.replace("cs:~", "");
     charmId = charmId.replace("cs:", "");
-    return `https://api.jujucharms.com/charmstore/v5/${charmId}/icon.svg`;
+
+    // Remove release from the charmId
+    // before: containers/trusty/lxd-container-47
+    // after:  containers/lxd-container-47
+    charmId = charmId.replace("precise/", "");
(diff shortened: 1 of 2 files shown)


huwshimi:
With one of the models I looked at all but one of them was missing. How do I know which ones should load or not (they all say "Charmhub" in the store column of the table:

Screen Shot 2022-03-30 at 10 00 36 am

I'm guess this is unrelated, but the icons that do load are strangely cropped (I'm using Chrome): 

Screen Shot 2022-03-30 at 10 00 09 am

anthonydillon:
Thanks @huwshimi I added some tests to cover the expected charmId's. Fixed the parsing to cover each of them.

anthonydillon:
In regards to the icon it seems like a difference in browser as I am using Chrome and looks ok for me. Probably another issue to resolve in the future.

src/app/utils/utils.js
```diff
@@ -249,8 +249,32 @@ export const generateIconPath = (charmId) => {
...
+
+    // Combine owner and charm name
+    // before: containers/lxd-container-47
+    // after:  containers-lxd-container-47
+    charmId = charmId.replace("/", "-");
+
+    // Strip the revision number from the end
+    // before: containers-lxd-container-47
+    // after:  containers-lxd-container
+    charmId = charmId.substr(0, charmId.lastIndexOf("-", charmId.length));
```
huwshimi, line 275: Great comments here, will certainly make it easier if this code gets revisited in the future.

huwshimi (approved):
Thanks Ant those icons are all loading now.

Issue #1035, opened 2021-06-07
Populating "Action Message" field not possible


erik78se:
GUI 0.7.1

Can't find any information on how to populate the "Action Message" field from the actions pane.

* The operator example documentation doesn't cover this 
* The only way to get something in here is to fail the action: 

`event.fail("some message")
`
![image](https://user-images.githubusercontent.com/850603/121096697-fddc0500-c7f2-11eb-9d77-ac31a18c7fff.png)

* No info in the docs here: https://discourse.charmhub.io/t/actions-for-the-charm-author/1113
* No info in the docs here: https://juju.is/docs/sdk/actions  (The example doesn't not produce the desired outcome e.g. setting any value)
* Also, no information that event.fail("some text")  actually sets that field. It should be mentioned to allow a developer to use that information to set meaningful messages for the operator.

hatched:
You're right the "Action Message" field is a bit confusing right now and we may rename it or remove it entirely in upcoming releases as it doesn't work as one would expect it to. This is actually just the first step towards the full implementation. You can see the end goal in the mockup below You'll note that in this we have the various output options split out in the drop down as this fits the output from the various action formats better.

![III  Action logs - Default view](https://user-images.githubusercontent.com/532033/121119328-a6c93680-c7d8-11eb-8ffd-c11db7722bb7.png)

erik78se:
I've logged duplicates of this. There must be something I can do to get something in there?

PR #1236, opened 2022-09-08
Get missing action logs

### LINKED ISSUE #1199: Documentation on actions in ops missing information on how to get info out from actions
I'm reading the docs to try figuring out how to get the "message" field in the GUI to show the result from my actions.

But its nothing there that helps.

What should I do to get anything displayed there?

This is the docs I'm referring to: https://operator-framework.readthedocs.io/en/latest/_modules/ops/charm.html?highlight=set_results#

goulinkh:
## Done

- Show the output of the message
- Add the ability to filter the logs by `stderr`, `stdout` or both

### TODO
- Add a button maybe in `controls`, that will show the payload of `set_results` 

## QA

- Run the dashboard locally
- Open http://0.0.0.0:3000/
- Run a charm that has an action with logs, you can run this one (for some reason this charm successful action fail on juju 3.0 beta4): https://github.com/goulinkh/qa-juju-charm
- Make sure that the logs are present in the dashboard and match the defined behavior in [Figma](https://www.figma.com/file/u9v1jvPc3TwlfwHSWsysn9/JAAS-Baseline?node-id=530%3A38081)

## Details

Fixes #1199, #1035 

## Screenshots
image

### DIFF
--- a/src/pages/EntityDetails/Model/ActionLogs/ActionLogs.test.js
+++ b/src/pages/EntityDetails/Model/ActionLogs/ActionLogs.test.js
@@ -9,12 +9,19 @@ import dataDump from "testing/complete-redux-store-dump";
 import ActionLogs from "pages/EntityDetails/Model/ActionLogs/ActionLogs";
 import TestRoute from "components/Routes/TestRoute";
 import { waitForComponentToPaint } from "testing/utils";
+import { act } from "react-dom/test-utils";
 
 jest.mock("juju", () => {
   return {
     queryOperationsList: () => {
       return new Promise((resolve) => {
         const apiData = require("testing/list-operations-api-response.json");
+        resolve(apiData.response);
+      });
+    },
+    queryActionsList: () => {
+      return new Promise((resolve) => {
+        const apiData = require("testing/list-actions-api-response.json");
         resolve(apiData.response);
       });
     },
@@ -46,9 +53,25 @@ describe("Action Logs", () => {
   it("requests the action logs data on load", async () => {
     const wrapper = await generateComponent();
     const expected = [
-      ["easyrsa", "1/list-disks", "completed", "", "", ""],
-      ["└easyrsa/0", "", "completed", "2", "-", "2021-04-14T20:27:57Z"],
-      ["└easyrsa/1", "", "completed", "3", "-", "2021-04-14T20:27:57Z"],
+      ["easyrsa", "1/list-disks", "completed", "", "", "", ""],
+      [
+        "└easyrsa/0",
+        "",
+        "completed",
+        "2",
+        "log message 1",
+        new Date("2021-04-14T20:27:57Z").toLocaleString(),
+        "Output",
+      ],
+      [
+        "└easyrsa/1",
+        "",
+        "completed",
+        "3",
+        "log message 1log message 2error message",
+        new Date("2021-04-14T20:27:57Z").toLocaleString(),
+        "Output",
+      ],
     ];
     const result = [];
     wrapper.find("tbody TableRow").forEach((row, rowIndex) => {
@@ -57,7 +80,7 @@ describe("Action Logs", () => {
         result[rowIndex].push(cell.text());
       });
     });
-    expect(expected).toEqual(result);
+    expect(result).toEqual(expected);
   });
 
   it("fails gracefully if app does not exist in model data", async () => {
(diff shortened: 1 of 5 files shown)


huwshimi (commented):
I've just done a QA. One little thing I spotted is that if you a row set to "STDOUT/STDERR" and there is only an error you end up with a bit of space above the error:

Screen Shot 2022-10-06 at 10 46 11 am

When just showing STDERR it looks fine:

Screen Shot 2022-10-06 at 10 46 20 am

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.test.js
```diff
@@ -78,6 +116,6 @@ describe("Action Logs", () => {
         result[rowIndex].push(cell.text());
       });
     });
-    expect(expected).toEqual(result);
+    expect(result).toEqual(expected);
   });
```
huwshimi, line 120: It'd probably be good to add a test for checking that changing the the output selector changes the message col.

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
```diff
@@ -102,6 +124,39 @@ export default function ActionLogs() {
 
   const applicationList = () => Object.keys(modelStatusData.applications);
 
+  const OutputSelector = ({
```
huwshimi, line 127: This is defined inside the `ActionLogs` component which means it'll get reconstructed every time `ActionLogs` is rerendered. It'd probably be best in its own file (though I know at the moment this project sometimes defines more than one component per file).

Though you could probably just use `ContextualMenu` inside the row object as this component is not too large and doesn't have any state or logic to test in isolation.

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
```diff
@@ -122,6 +188,11 @@ export default function ActionLogs() {
   }, [modelUUID]);
 
   const tableData = useMemo(() => {
+    const handleOutputSelect = (actionTag: string, type: string) => {
```
huwshimi, line 191: `type` is a reserved word in typescript strict mode so it might be good to rename this to be ready for the future.

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
```diff
@@ -102,6 +124,39 @@ export default function ActionLogs() {
...
+      hasToggleIcon
+      toggleLabel={type === "STDOUT/STDERR" ? "Output" : type}
+      links={[
+        {
+          children: "STDOUT/STDERR",
+          onClick: () => onSelect("STDOUT/STDERR"),
+        },
+        {
+          children: "STDERR",
+          onClick: () => onSelect("STDERR"),
```
huwshimi, line 148: These states rely on specific strings so it'd be a little less fragile if these were to use an enum, something like:

```typescript
enum Output {
    ALL = "STDOUT/STDERR",
    STDERR = "STDERR",
    STDOUT = "STDOUT",
}
```

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
```diff
@@ -135,6 +206,22 @@ export default function ActionLogs() {
...
+          if (!actionFullDetails) return;
+          const stdout = (actionFullDetails.log || [])
+            .map((m) => m.message)
+            .join("\n");
+          const stderr =
+            actionFullDetails.status === "failed"
+              ? actionFullDetails.message
+              : "";
+          const StdOut = () => <span>{stdout}</span>;
+          const StdErr = () => <span style={{ color: "red" }}>{stderr}</span>;
```
huwshimi, line 224: It'd probably be good do use the Vanilla red for this. You'll need to create a new class and set `color: $color-negative;`.

huwshimi (approved):
This is looking good, I've made a few suggestions.

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
```diff
@@ -112,6 +167,17 @@ export default function ActionLogs() {
         appStore.getState()
       );
       setOperations(operationList.results);
+      const actionsTags = operationList.results
+        .flatMap((operation: Operation) =>
```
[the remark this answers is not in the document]
goulinkh: Good catch!

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
goulinkh: Good idea :)

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
goulinkh: To be honest I was surprised that it worked!

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
goulinkh: I thought that creating a new scss file just for one rule is too much, I will do that then thanks!

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
goulinkh: Makes sense thanks for the explanation, this is very helpful!

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.tsx
goulinkh: Having a consistent way of styling is good :+1:

src/pages/EntityDetails/Model/ActionLogs/ActionLogs.test.js
goulinkh: Good call!

PR #1238, opened 2022-09-15
Revert renovate config

huwshimi:
## Done

- Revert https://github.com/canonical/jaas-dashboard/pull/1224 so that we get more frequent updates (renovate was changed while this project wasn't getting a lot of attention).
- I made one change to the config which is to change the schedule from weekly to monthly (I thought it could be worth trying that cadence for a while).

## QA
n/a


goulinkh (approved):
LGTM

PR #1240, opened 2022-09-16
Add a local config file.

huwshimi:
## Done

- Provide a way to configure the dashboard locally while keeping a clean git tree.

## QA

- Follow the modified instructions in the readme to configure your dashboard.
- Run `git status` and your tree should be clean.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -13,6 +13,9 @@
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/app-icon.png" />
     <script src="/config.js"></script>
+    <% if (process.env.REACT_APP_USE_LOCAL_CONFIG) { %>
+    <script src="/config.local.js"></script>
+    <% } %>
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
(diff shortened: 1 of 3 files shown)


huwshimi:
@goulinkh not sure if you've run into this yet, but I kept checking in my config.js. There might be a better way to do this, so open to any suggestions you have!

public/index.html
```diff
@@ -13,6 +13,9 @@
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/app-icon.png" />
     <script src="/config.js"></script>
+    <% if (process.env.REACT_APP_USE_LOCAL_CONFIG) { %>
```
huwshimi, line 16: I initially used `NODE_ENV === "development"` but then if you don't need a config.local.js this will result in an error when it tries to load a file that doesn't exist.

goulinkh: Yes very good point! I'm doing the same thing stash the file `config.js` often.
The downside of having a new env variable is that we have to add it locally, I found a way to make this work automatically that may work as well:
```html
    <% if (process.env.NODE_ENV === "development" &&
    require("fs").existsSync("public/config.local.js")) { %>
...
```

goulinkh (approved):
LGTM :rocket:

public/index.html
huwshimi: Thanks! That's much nice, I'll update it.

Issue #965, opened 2021-04-20
Humanise date/time


anthonydillon:
**Describe the bug**
It would be nice to humanise the date/time table data and provide a label with the actual value.

**To Reproduce**
1. Go to the new actions list
2. Scroll to the  right hand column
3. See that date/time is not very useful

**Expected behavior**
I would expect something like "2 days ago" with a hover that shows the time and date.

**Screenshots**
![image](https://user-images.githubusercontent.com/1413534/115411055-0f3d6380-a1eb-11eb-9e75-2215f97f535d.png)

hatched:
Do we have a library that we use across any other projects for this feature?

huwshimi:
> Do we have a library that we use across any other projects for this feature?

We use https://date-fns.org/ in MAAS.

PR #1282, opened 2022-10-24
Use relative time in action log

huwshimi:
## Done

- Change time in the action log to be relative and display the timestamp in a tooltip.
- Update tests to TypeScript/RTL.

## QA

- Go to a model that has action logs (you might need to perform an action for an app).
- Check that the completion time column shows relative dates.
- Hover the date and you should see the exact date/time.

## Details

Fixes: #965.

## Screenshots

Screen Shot 2022-10-24 at 12 19 45 pm


goulinkh:
Upon running an action, I get this 
image

goulinkh:
For some context here is object in the redux store:
image

huwshimi:
@goulinkh it looks like this branch is displaying the correct relative date, but for some reason the action timestamps returned by Juju (or jujulib) are `0001-01-01T00:00:00Z`.

It would probably be good to check where this date is coming from. I wasn't able to reproduce it (with Juju 2 or 3) so if you still have that model would you mind checking your websocket connection to see if that is what Juju is returning? I don't think we do any date manipulation in jujulib, but it would be good to rule that out.

goulinkh (approved):
LGTM, this is way more readable :+1: 

FYI I filed an issue for the date edge case value: https://github.com/canonical/jaas-dashboard/issues/1351

Issue #922, opened 2021-03-12
Machines tab missing 'empty' message


hatched:
![Screen Shot 2021-03-12 at 4.33.20 PM.png](https://images.zenhubusercontent.com/5bf56fa65f3ffb2e1680062c/01d18ffd-2826-4a38-a309-d677fb74ec96)

vs the apps & integrations

![Screen Shot 2021-03-12 at 4.33.58 PM.png](https://images.zenhubusercontent.com/5bf56fa65f3ffb2e1680062c/532a398e-e958-42a5-9170-c0e8ad56b732)

PR #1283, opened 2022-10-25
Display a message when there are no machines

huwshimi:
## Done

- Display a message when there are no machines in a model.
- Reenable Model tests and update to TypeScript and RTL.

## QA

- Click on a model that doesn't have any machines and then go to the 'Machines' tab.
- You should see a message and a link to the docs.

## Details

Fixes: #922.

## Screenshots

Screen Shot 2022-10-25 at 2 23 55 pm


src/pages/EntityDetails/Model/Model.test.tsx
```diff
@@ -0,0 +1,521 @@
...
+        const apiData = require("testing/list-actions-api-response.json");
+        resolve(apiData.response);
+      });
+    },
+  };
+});
+
+const mockStore = configureStore([]);
+
+describe("Model", () => {
```
huwshimi, line 46: These tests were skipped for some reason. I've done a rough update to RTL but this really needs the factories before it can be cleaned up further.

src/pages/EntityDetails/Model/Model.test.tsx
```diff
@@ -0,0 +1,521 @@
...
+      </Provider>
+    );
+
+    const noRelationsMsg = document.querySelector(
+      "[data-testid='no-integrations-msg']"
+    );
+    expect(noRelationsMsg).toBeInTheDocument();
+  });
+
+  it("should show a message if a model has no machines", () => {
```
huwshimi, line 300: This is the only test that was added.

goulinkh (approved):
LGTM

Issue #1280, opened 2022-10-23
Error configuring charms


huwshimi:
**Describe the bug**
Saving a charm configuration form results in:

```
index.js:440 Uncaught (in promise) TypeError: conn.facades.application.set is not a function
    at setApplicationConfig (index.js:440:1)
    at async _submitToJuju (ConfigPanel.tsx:173:1)
```

**To Reproduce**
1. Click on a model -> click an app -> Click 'Configure' button.
2. Change a value, click 'Save and apply' and then 'Yes, apply changes'.
3. The button will get stuck on saving and in your browser console you'll have the error.

**Expected behavior**
The form should save.

**Additional context**
I tested this with both Juju 2.9 and 3.0 and both gave the same error.

PR #1288, opened 2022-10-27
Fix setting config for an app

huwshimi:
## Done

- Fix setting config for an app.

## QA

- Click on a model -> click an app -> Click 'Configure' button.
- Change a value, click 'Save and apply' and then 'Yes, apply changes'.
- The form should save correctly.
- Refresh the page and the form should contain the new values.

## Details

Fixes: #1280.

### DIFF
--- a/src/juju/index.js
+++ b/src/juju/index.js
@@ -437,9 +437,13 @@ export async function setApplicationConfig(
       setValues[key] = `${config[key].newValue}`;
     }
   });
-  const resp = await conn.facades.application.set({
-    application: appName,
-    options: setValues,
+  const resp = await conn.facades.application.setConfigs({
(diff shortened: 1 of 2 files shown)


src/juju/index.js
```diff
@@ -437,9 +437,13 @@ export async function setApplicationConfig(
       setValues[key] = `${config[key].newValue}`;
     }
   });
-  const resp = await conn.facades.application.set({
-    application: appName,
-    options: setValues,
+  const resp = await conn.facades.application.setConfigs({
```
huwshimi, line 440: This method has changed for recent facade versions.

goulinkh (approved):
LGTM :+1:

PR #1342, opened 2022-11-20
WD-481 Store and fetch the bakery outside of the Redux store

huwshimi:
## Done

- Move bakery to a separate file.
- Replace selectors and reducers with bakery import.
- Update tests to use the new bakery setup.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check that you can still log in/out and display models.

### DIFF
--- a/src/@canonical/macaroon-bakery.d.ts
+++ b/src/@canonical/macaroon-bakery.d.ts
@@ -1,0 +1,1 @@
+declare module "@canonical/macaroon-bakery";
--- a/src/app/root.js
+++ b/src/app/root.js
@@ -1,7 +1,9 @@
 import immerProduce from "immer";
 import cloneDeep from "clone-deep";
 
-import { actionsList } from "./actions";
+import bakery from "app/bakery";
+
+import { actionsList } from "./action-types";
 
 function rootReducer(state = {}, action) {
   return immerProduce(state, (draftState) => {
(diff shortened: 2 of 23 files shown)


src/@canonical/macaroon-bakery.d.ts
```diff
@@ -0,0 +1 @@
+declare module "@canonical/macaroon-bakery";
```
goulinkh, line 1: Sorry I missed your PR on the bakery lib to move to Typescript!

src/app/root.js
```diff
@@ -32,7 +31,7 @@ function rootReducer(state = {}, action) {
         draftState.visitURL = action.payload;
         break;
       case actionsList.logOut:
-        delete draftState.bakery.storage._store.identity;
+        delete bakery.storage._store.identity;
```
goulinkh, line 34: I'm not sure why we are doing this at 2 places: here & `actions/logOut`?

goulinkh (approved):
I suppose we don't need a state for bakery so this makes sense :+1:

src/@canonical/macaroon-bakery.d.ts
huwshimi: No worries, thanks for the review!

src/app/root.js
huwshimi: Ah yeah, well spotted. I can't see any reason for that so I'll remove it from here.

PR #1355, opened 2022-12-06
WD-912 - Rename root state

huwshimi:
## Done

- Rename root state (that's not actually the real root state, but a subsection of state) to 'general'.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The UI should load as normal.

## Details

Fixes: https://warthogs.atlassian.net/browse/WD-912

### DIFF
--- a/src/app/actions.js
+++ b/src/app/actions.js
@@ -107,7 +107,7 @@ export function logOut(store) {
   async function logOut(dispatch) {
     const state = store.getState();
     const identityProviderAvailable =
-      state?.root?.config?.identityProviderAvailable;
+      state?.general?.config?.identityProviderAvailable;
     const pingerIntervalIds = getPingerIntervalIds(state);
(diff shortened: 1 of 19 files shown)


src/app/actions.js
```diff
@@ -107,7 +107,7 @@ export function logOut(store) {
   async function logOut(dispatch) {
     const state = store.getState();
     const identityProviderAvailable =
-      state?.root?.config?.identityProviderAvailable;
+      state?.general?.config?.identityProviderAvailable;
```
huwshimi, line 110: I'd be happy with another name if you have any other suggestions @goulinkh?

goulinkh: `general` seems fine to be, but you could also name it: `connection`, `config`,`login` based on the list of actions that the reducer handles.

goulinkh (approved):
LGTM

PR #1360, opened 2023-01-05
WD-922 - Handle updating model permissions

huwshimi:
## Done

Support updating model permissions in the model poller middleware.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Open the list of models.
- Click the "Access" button for one of the models.
- Enter a username for a real user and click "Add user" and you should get a notification that the user was added and they should appear in the list of users.
- Now enter an incorrect username and click "Add user" and you should get an error.

[WD-922]: https://warthogs.atlassian.net/browse/WD-922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/store/middleware/check-auth.js
+++ b/src/store/middleware/check-auth.js
@@ -71,8 +71,7 @@ export default ({ getState }) =>
         checkLoggedIn(state, wsControllerURL)
       ) {
         // Await the next to support async thunks
-        await next(action);
-        return;
+        return await next(action);
       } else {
         error(action.NAME, wsControllerURL);
       }
@@ -81,8 +80,7 @@ export default ({ getState }) =>
         actionAllowlist.includes(action.type) ||
         checkLoggedIn(state, wsControllerURL)
       ) {
-        next(action);
-        return;
+        return next(action);
       } else {
         error(action.type, wsControllerURL);
       }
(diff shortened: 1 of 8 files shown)


src/store/middleware/check-auth.js
```diff
@@ -70,8 +70,7 @@ export default ({ getState }) =>
         checkLoggedIn(state, wsControllerURL)
       ) {
         // Await the next to support async thunks
-        await next(action);
-        return;
+        return await next(action);
```
huwshimi, line 73: I was stumped for ages about why the promises were getting lost in the middleware and it turned out that this middleware was not returning the next promise.

src/store/middleware/model-poller.ts
```diff
@@ -138,13 +145,29 @@ export const modelPollerMiddleware: Middleware = (reduxStore) => {
...
+        }
+      );
+      return;
+    } else if (action.type === actionsList.logOut) {
+      jujus.forEach((juju) => {
+        juju.logout();
+      });
       return next(action);
-    };
+    } else if (action.type === actionsList.updatePermissions) {
```
huwshimi, line 156: This now needs to be handled in the middleware as it need to have access to the controller connections which are only stored in the middleware (previously they were stored in the redux store).

src/store/middleware/check-auth.js
goulinkh: Wow I can imagine how frustrating this is!

goulinkh (approved):
LGTM :+1:
Good job! I was about to file an issue before seeing this PR!

PR #1379, opened 2023-02-15
WD-2030 - Replace controller test data with factories

huwshimi:
## Done

- Add controller factories.
- Update tests to use factories instead of data dump.
- Removed unused test data.

## Details

https://warthogs.atlassian.net/browse/WD-2030

### DIFF
--- a/src/components/ModelTableList/CloudGroup.test.tsx
+++ b/src/components/ModelTableList/CloudGroup.test.tsx
@@ -6,6 +6,7 @@ import { QueryParamProvider } from "use-query-params";
 import { ReactRouter6Adapter } from "use-query-params/adapters/react-router-6";
 
 import { rootStateFactory } from "testing/factories/root";
+import { generalStateFactory, configFactory } from "testing/factories/general";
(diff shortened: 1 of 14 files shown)


src/components/ModelTableList/CloudGroup.test.tsx
```diff
@@ -72,7 +73,26 @@ describe("CloudGroup", () => {
...
+              user: {
+                "display-name": "eggman",
+                identity: "user-eggman@external",
+                "controller-access": "",
+                "model-access": "",
+              },
+            },
+          },
+        }),
+        juju: dataDump.juju,
```
goulinkh, line 93: shouldn't this be replaced with a factory as well?

huwshimi: Yes! I just haven't created the juju factories yet (that's the next thing to do). There actually is a juju state factory but it includes outdated data so it's not applicable for all tests.

goulinkh (approved):

Issue #1237, opened 2022-09-14
It is trivial to present the dashboard with a "working" config that cannot access models


pengale:
**Describe the bug**

The call to String.replace in https://github.com/canonical/jaas-dashboard/blob/ce5dae02f35f605989d1c6a5bcd633b94df6047e/src/juju/index.js#L188 depends on the controller url being specified with a trailing `/api`. For the overall toolchain, that parameter is optional. 

The dashboard code contains a little gotcha when those four characters are omitted in the value dropped into config.js. In this case, the dashboard will successfully connect to the controller, but then use the unmodified controller url to connect to the individual model apis. This produces a failure a few lines down, when the dashboard attempts to use the facades that it believes that it has fetched from the models' connection.

**To Reproduce**
Test the dashboard against a local controller, per the REAME. In config.js, leave out the "/api" from the controller url. The dashboard will error out on login.

**Expected behavior**
The dashboard normally connects individually to each model in order to fetch model facades. This is a little bit of a hack, but the controller will not expose those facades on the main connection.

The dashboard code should be refactored to detect whether or not it is actually connecting to a model, and error out when it cannot. Additionally, the code should be more careful with the .replace :-)

**Additional context**
Two devs reading the code line by line managed to miss this, and it caused some rather awkward delays in the process of developing the dashboard charm.

PR #1390, opened 2023-02-28
WD-2351- Display controller endpoint configuration errors.

huwshimi:
## Done

- Validate and display controller endpoint configuration errors.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Change the value of controllerAPIEndpoint in your config.local.js to an invalid address (e.g. remove the /api or the wss://) and reload the dashboard.
- You should see an error message.

## Details

Fixes: #1237.
https://warthogs.atlassian.net/browse/WD-2351

## Screenshots

Screenshot 2023-02-28 at 11 25 32 am

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,4 +1,5 @@
-import { StrictMode } from "react";
+import { Notification, Strip } from "@canonical/react-components";
+import { ReactNode, StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import { Provider } from "react-redux";
 import process from "process";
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -53,8 +62,32 @@ checkConfigExists();
...
+  let error: string | null = null;
   if (!config) {
-    console.error("No configuration found.");
+    error = "No configuration found.";
+  }
+  const addressRegex = new RegExp(/^ws[s]?:\/\/(\S+)\/api$/);
+  if (!config?.controllerAPIEndpoint) {
+    error = `controllerAPIEndpoint is not set.`;
+  } else if (!config?.controllerAPIEndpoint.endsWith("/api")) {
+    error = `controllerAPIEndpoint (${config?.controllerAPIEndpoint}) must end with /api.`;
```
goulinkh, line 73: Another possibility here is to postfix the API endpoint with `/api`,  but it's up to you whether you want it to be specifically set by the user or not.

goulinkh (approved):
LGTM

src/index.tsx
huwshimi: I did wonder about this, but the controllerAPIEndpoint needs to match values returned by Juju so I thought it was probably best (and less error prone) if this value is set correctly in the config (e.g. we could easily forget to append "/api" somewhere else in the code if the value is read straight from the config).

Issue #1352, opened 2022-12-02
WebCLI not showing on Juju 3.0


goulinkh:
The WebCLI is only showing on Juju 2.9

PR #1393, opened 2023-02-28
Show Juju CLI on Juju 3

huwshimi:
## Done

- Fix for the CLI not showing on Juju 3.

## QA

- Connect your dashboard to a Juju 3 instance.
- Load the dashboard and go to a model.
- The CLI input should appear at the bottom and you should be able to run commands.

## Details

Fixes: #1352.
https://warthogs.atlassian.net/browse/WD-2354

### DIFF
--- a/src/pages/EntityDetails/EntityDetails.test.tsx
+++ b/src/pages/EntityDetails/EntityDetails.test.tsx
@@ -212,6 +212,25 @@ describe("Entity Details Container", () => {
     });
   });
 
+  it("shows the CLI in juju higher than 2.9", async () => {
+    state.juju.modelWatcherData = {
+      abc123: modelWatcherModelDataFactory.build({
+        applications: {
+          "ceph-mon": applicationInfoFactory.build(),
+        },
+        model: modelWatcherModelInfoFactory.build({
+          name: "enterprise",
+          owner: "kirk@external",
+          version: "3.0.7",
+        }),
+      }),
+    };
+    renderComponent();
+    await waitFor(() => {
+      expect(screen.queryByTestId("webcli")).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


goulinkh:
The commands endpoint isn't working in my case:
Selection_001

Where I'm running Juju 3.0 on a k8s env:
Selection_002

huwshimi:
@goulinkh that's strange. Somehow your address starts with "wss://ws//". I wonder if this is somehow related to when you were testing my other PR about breaking the controller address? Though if that was the case I'd expect the model not to load at all.

I've pushed a fix related to the ws protocol, but I don't think that is causing this issue. Could you check your Redux state and check what your controller URL is set to in the juju.controllers object key?

Screenshot 2023-03-03 at 10 33 32 am

goulinkh:
@huwshimi The URL issue is solved with your fixes, but the server is still not working, have you tried it on  a Juju 3.0?
image

goulinkh:
I managed to find the issue in the dashboard's k8s charm: https://github.com/canonical/juju-dashboard-charm/pull/22
The one thing to change in this PR, is to get the protocol `ws://` or `wss://` from the config. Currently it's always `wss://`

huwshimi:
> I managed to find the issue in the dashboard's k8s charm: [canonical/juju-dashboard-charm#22](https://github.com/canonical/juju-dashboard-charm/pull/22) The one thing to change in this PR, is to get the protocol `ws://` or `wss://` from the config. Currently it's always `wss://`

Ah that would explain why it was working for me as I'm using Juju without k8s. I've updated this PR to get the correct protocol.

goulinkh (approved):
LGTM :+1:

Issue #1204, opened 2022-01-13
Boolean Toggle is a textbox


JamesYangLim:
Hi, in ceph-osd zap-disk action i-really-mean-it field should be a toggle box instead of a textbox.

![image](https://user-images.githubusercontent.com/30457655/149286916-17dfc056-ebab-4843-bbee-ec66ade9bd02.png)

huwshimi:
It looks like everything is currently being set to a text input: https://github.com/canonical/jaas-dashboard/blob/main/src/components/RadioInputBox/OptionInputs.tsx#L52.

PR #1394, opened 2023-03-01
WD-2356 - Support boolean action options

huwshimi:
## Done

- Support boolean action options.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Deploy and app that has boolean action options e.g. ceph-osd.
- Navigate to the unit list for that app and tick some units.
- Click 'Run action' and choose one of the actions that has a boolean option.
- Check that the boolean option has a checkbox instead of a text field.

## Details

#1204.
https://warthogs.atlassian.net/browse/WD-2356

## Screenshots

Screenshot 2023-03-01 at 4 27 37 pm

### DIFF
--- a/src/components/RadioInputBox/OptionInputs.test.tsx
+++ b/src/components/RadioInputBox/OptionInputs.test.tsx
@@ -34,6 +34,41 @@ describe("OptionInputs", () => {
     expect(container).toMatchSnapshot();
   });
 
+  it("can display a text input", () => {
+    const onValuesChange = jest.fn();
+    render(
+      <Formik initialValues={{}} onSubmit={jest.fn()}>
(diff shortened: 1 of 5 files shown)


goulinkh (approved):
LGTM, one tiny comment, should we use the [Switch component](https://canonical.github.io/react-components/?path=/story/switch--default-story) instead as it's already used in other places?

huwshimi:
> LGTM, one tiny comment, should we use the [Switch component](https://canonical.github.io/react-components/?path=/story/switch--default-story) instead as it's already used in other places?

As I understand it, the Switch component is designed for immediate changes, like changing what is visible on a page or maybe settings, but in a form we should be using a checkbox.

Issue #725, opened 2020-10-24
Public address should be a link


anthonydillon:
When exposing an application the next thing you are likely to want to do is visit the public address which we should make an easy accees link in the details. Which opens in a new tab.

hatched:
There has been a lot of discussion over the years about allowing a charm to define it's addressable endpoints so that we could create links to them in the UI. So far nothing has materialized. People can also expose an application over a range of ports so we won't have just "one" link necessarily. It's also unlikely that the application gets exposed just on the IP and not on some additional path.

So this was a long way of saying we can definitely turn it into a link but it probably won't be super useful :)

anthonydillon:
If we are displaying an IP, that is an address. Therefore seems reasonable to make it clickable instead making the user copy the text and entering it into another tab. Often the address is truncated and so the only way to reach the IP address is to inspect the table cell and copy the contents. 

We are not promising that the IP address will result in an application or the right place. That is an issue in which would be solved by charm authors providing ports and paths to there application UI. But, until then let's not make people inspect a page to get the address out of it, just to see.

PR #1398, opened 2023-03-03
WD-2409- Display copy button for ip addresses

huwshimi:
## Done

- Update the unit list with a hover button to copy the address.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Navigate to an application that has units with exposed IPs.
- Hover an IP address and you should see a copy button. Click the button and the address should get copied.

## Details

Fixes: #725.
https://warthogs.atlassian.net/browse/WD-2409

## Screenshots
Screen Recording 2023-03-09 at 9 52 04 am

### DIFF
--- a/src/tables/tableRows.tsx
+++ b/src/tables/tableRows.tsx
@@ -11,8 +11,9 @@ import {
   generateRelationIconImage,
   generateStatusElement,
   generateIconImg,
+  copyToClipboard,
 } from "components/utils";
-import { Tooltip } from "@canonical/react-components";
+import { Button, Icon, Tooltip } from "@canonical/react-components";
 import {
   MainTableCell,
   MainTableRow,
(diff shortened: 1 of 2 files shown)


src/tables/tableRows.tsx
```diff
@@ -34,6 +34,23 @@ export type Query = {
   activeView?: string | null;
 };
 
+const generateLink = (address?: string | null) =>
+  address ? (
+    <a
+      href={`http://${address}`}
```
goulinkh, line 40: A lot of  units are running an HTTP server however sometimes it could be custom protocols e.g. `postgres://`, `redis://`,...
Another option could be to show a copy to clipboard button to have the IP in one click and then have the possibility to paste it to a new tab (browser will add http:// prefix automatically) or paste it to program or terminal easily.

https://user-images.githubusercontent.com/36013798/222744307-136a4f10-8e9a-442c-9897-a2016606bd65.mov

huwshimi: This is a really good point. I've updated it with a copy button.

goulinkh (approved):
LGTM

Issue #849, opened 2021-01-29
Config inputs do not validate and fail silently


barrymcgee:
- Go to CMR model and click Logstash app
- Click Configure
- Enter a letter in 'beats_port' field
- Click Save
- Saving fails silently

<img width="1280" alt="Screenshot 2021-01-29 at 10 15 27" src="https://user-images.githubusercontent.com/505570/106267355-6a317100-6221-11eb-91cd-46ab1f3872ee.png">

hatched:
Yep this is known, validation and error reporting haven't yet been implemented. I'll leave this open until that has happened.

PR #1411, opened 2023-03-08
WD-2459 - config errors and validation

huwshimi:
## Done

- Display config API errors
- Dedupe config field shared code.
- Add number field.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Open the config panel for an app with a number config and check that it appears as a number field and that you can't input text.
- The easiest way to get a config error seems to be to inspect a number input and change its type to text and then enter in some non number characters. Save the form and it should display the errors at the top of the form.

## Details

Fixes: #849.
https://warthogs.atlassian.net/browse/WD-2459

## Screenshots

Screenshot 2023-03-07 at 2 46 53 pm

### DIFF
--- a/src/panels/ConfigPanel/ConfigField.tsx
+++ b/src/panels/ConfigPanel/ConfigField.tsx
@@ -1,0 +1,141 @@
+import { ReactNode, useEffect, useRef, useState } from "react";
+import classnames from "classnames";
+
+import { isSet } from "components/utils";
+import { ConfigData } from "juju/api";
+
+export type SetNewValue = (name: string, value: any) => void;
+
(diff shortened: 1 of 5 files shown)


src/panels/ConfigPanel/ConfigField.tsx
```diff
@@ -0,0 +1,141 @@
+import { ReactNode, useEffect, useRef, useState } from "react";
```
huwshimi, line 1: This component contains everything that was duplicated between `TextAreaConfig` and `BooleanConfig`.

goulinkh (approved):

Issue #525, opened 2020-05-28
Improve error boundary notice to include more feedback


barrymcgee:
When users hit an this error, we should do more to inform them of the specifics of what went wrong, if possible.

Possibly include a stack trace which could be copied/pasted into Github issue?

PR #1416, opened 2023-03-14
WD-2610 - Improve the error boundary

huwshimi:
## Done

- Migrate ErrorBoundary to TypeScript.
- Display error and stack trace on error page.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- The easiest way to see the error boundary is to throw an error e.g add `throw new Error("It's definitely broken");` to ModelsIndex.tsx.
- Check that the error and stack trace are shown.
- Click the "raise an issue" link and check that the stack trace is included in the issue.

## Details

Fixes: #525.
https://warthogs.atlassian.net/browse/WD-2610

## Screenshots

Screenshot 2023-03-14 at 12 41 12 pm
Screenshot 2023-03-14 at 11 38 52 am


goulinkh (approved):
This is really cool, one small improvement overall could be to support source map so that the stack trace shows Typescript file names with correct line numbers

huwshimi:
> This is really cool, one small improvement overall could be to support source map so that the stack trace shows Typescript file names with correct line numbers

That's a good idea, I've added it as a new issue so that we can make sure it doesn't increase the bundle size too much as well as make sure it gets served correctly: https://github.com/canonical/jaas-dashboard/issues/1421

PR #1426, opened 2023-03-16
WD-2663 - Fix lockup

huwshimi:
## Done

- Fix lockup caused by API call on each render.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Load the dashboard.
- Make sure that you can interact with the dashboard and that it doesn't lock up almost immediately.

## Details

https://warthogs.atlassian.net/browse/WD-2663

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.tsx
+++ b/src/components/PrimaryNav/PrimaryNav.tsx
@@ -3,7 +3,7 @@ import { Icon, StatusLabel, Tooltip } from "@canonical/react-components";
 import classNames from "classnames";
 import Logo from "components/Logo/Logo";
 import UserMenu from "components/UserMenu/UserMenu";
-import { useMemo, useState } from "react";
+import { useEffect, useMemo, useState } from "react";
 import { useSelector } from "react-redux";
 import { NavLink } from "react-router-dom";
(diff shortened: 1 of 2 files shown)


src/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -73,10 +73,18 @@ const ControllersLink = () => {
 const PrimaryNav = () => {
   const appVersion = useSelector(getAppVersion);
   const [updateAvailable, setUpdateAvailable] = useState(false);
+  const [versionRequested, setVersionRequested] = useState(false);
 
-  dashboardUpdateAvailable(appVersion || "").then((e) => {
```
huwshimi, line 77: I missed this in review, but this call was causing an infinite loop as it would be called on every render, and after each API response it would update state, causing a new render which started the process again.

goulinkh (approved):

src/components/PrimaryNav/PrimaryNav.tsx
goulinkh: Sorry about that!

goulinkh: Also thanks for the mock! for some reason `jest.mock` was just not working in this scenario.

Issue #943, opened 2021-04-06
Investigate converting font-size: `0.875rem;` to `%small-text`


hatched:
ref: https://github.com/canonical-web-and-design/jaas-dashboard/pull/940/files#r607645978

PR #1430, opened 2023-03-17
WD-2691 - Use small Vanilla font size

huwshimi:
## Done

- Get the small font size from Vanilla rather than hardcoding it.
- Remove unused Button component.

## QA

This is a straight swap of the value so probably no need to QA?

## Details

Fixes: #943.
https://warthogs.atlassian.net/browse/WD-2691


src/components/Button/Button.tsx
huwshimi, line 1: When I was testing these changes I realised this component was not used.

goulinkh (approved):

Issue #724, opened 2020-10-24
wCLI history


anthonydillon:
We should store the latest commands run in the wCLI which is recalled with the arrow keys.

huwshimi:
This could potentially be stored in local storage so these commands are available when you revisit the app.

PR #1433, opened 2023-03-21
WD-2757- WebCLI history

huwshimi:
## Done

- Add history to the WebCLI.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Using a local controller go to a model details page.
- Using the CLI at the bottom enter a few commands.
- Press the up and down arrows and you should be able to navigate through your history.
- Refresh the page and you should still be able to navigate through your history.

## Details

https://warthogs.atlassian.net/browse/WD-2757
Fixes: #724.

## Screenshots

Screen Recording 2023-03-21 at 12 36 22 pm

### DIFF
--- a/src/components/WebCLI/WebCLI.tsx
+++ b/src/components/WebCLI/WebCLI.tsx
@@ -60,7 +60,7 @@ const WebCLI = ({
     wsMessageStore.current = "";
     setOutput(""); // Clear the output when sending a new message.
   };
-  const keyListener = useCallback(
+  const keydownListener = useCallback(
     (event: KeyboardEvent) => {
       if (event.key === "ArrowUp" || event.key === "ArrowDown") {
         let newPosition = historyPosition;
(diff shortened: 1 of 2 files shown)


src/components/WebCLI/WebCLI.tsx
```diff
@@ -35,16 +43,53 @@ const WebCLI = ({
 }: Props) => {
   const [connection, setConnection] = useState<Connection | null>(null);
   const [shouldShowHelp, setShouldShowHelp] = useState(false);
+  const [historyPosition, setHistoryPosition] = useState(0);
   const inputRef = useRef<HTMLInputElement>(null);
   const wsMessageStore = useRef<string>("");
   let [output, setOutput] = useState("");
   const sendAnalytics = useAnalytics();
   const storeState = useStore().getState();
+  const [cliHistory, setCLIHistory] = useLocalStorage<string[]>(
```
goulinkh, line 52: We should consider adding a size limit to the size of history commands, where the size limit of the local storage is ~5-10 Mb which is huge however parsing and stringifying for example a 5Mb of JSON may not be ideal for performance.
having a limit around 200 items should be fine.

huwshimi: Good idea! I've updated this.

goulinkh (approved):
This works really well, I would just set the cursor at the end of the line if possible as sometimes it's at the beginning

huwshimi:
> This works really well, I would just set the cursor at the end of the line if possible as sometimes it's at the beginning

Done!

Issue #862, opened 2021-02-05
[Config form] Keyboard users must press tab twice to move through form fields


barrymcgee:
Users should only have to press the tab key once to progress through fields in a form.

PR #1439, opened 2023-03-24
WD-2832 - Fixed extra tab over hidden config buttons

huwshimi:
## Done

- Fixed extra tab over hidden config buttons.
- Improved accessibility of config items.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to an app and open the config panel.
- Click inside an input.
- Press tab once and the next config item should be highlighted.
- Press enter or space and the help on the right should update.
- Press tab again and (assuming you haven't changed the value from the default) the focus should move to the input.
- Change one of the values so that the 'use default' button appears and check that you can tab to that button.

## Details

Fixes: #862.
https://warthogs.atlassian.net/browse/WD-2832

### DIFF
--- a/src/components/DivButton/DivButton.test.tsx
+++ b/src/components/DivButton/DivButton.test.tsx
@@ -1,0 +1,36 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import DivButton from "./DivButton";
+
+describe("DivButton", () => {
+  it("calls the on-click handler when clicked", async () => {
+    const onClick = jest.fn();
+    render(<DivButton onClick={onClick} />);
+    await userEvent.click(screen.getByRole("button"));
+    expect(onClick).toHaveBeenCalled();
+  });
+
+  it("calls the on-click handler when space is pressed", async () => {
+    const onClick = jest.fn();
+    render(<DivButton onClick={onClick} />);
+    await userEvent.type(screen.getByRole("button"), " ");
+    expect(onClick).toHaveBeenCalled();
+  });
+
+  it("calls the on-click handler when enter is pressed", async () => {
+    const onClick = jest.fn();
+    render(<DivButton onClick={onClick} />);
+    await userEvent.type(screen.getByRole("button"), "{enter}");
+    expect(onClick).toHaveBeenCalled();
+  });
+
+  it("does not call the on-click handler when another key is pressed", async () => {
+    const onClick = jest.fn();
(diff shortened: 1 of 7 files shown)


huwshimi:
> Code seems good to me, however one of the QA steps wasn't working: When I type something in an empty input, and then click on the default button that appears, the default button then dissapears but the input value doesn't reset.

Strangely I can't seem to reproduce this issue. Can you see if there is a specific scenario or any other details that might help tracking this down?

Screen Recording 2023-03-28 at 9 19 43 am

goulinkh:
Sorry I couldn't find the reason why it is not rerendering the text area after the config value reset, but in my case clicking use default is changing the value except for the textarea element in the DOM

huwshimi:
> Sorry I couldn't find the reason why it is not rerendering the text area after the config value reset, but in my case clicking use default is changing the value except for the textarea element in the DOM

If you don't mind could I grab a couple more details?
- which browser/version are you using?
- can you tell me a specific charm and config value that this is happening with?

goulinkh:
I've tried it on the latest version of chrome, the boolean configs reset works but not the text ones, here is the charm that isn't working: ch:amd64/jammy/mysql-k8s-31

huwshimi:
@goulinkh is it working correctly on main for you?

Also could you try changing the TextAreaConfig [input prop](https://github.com/canonical/jaas-dashboard/blob/main/src/panels/ConfigPanel/TextAreaConfig.tsx#L18) to the following:

```jsx
input={(value) => {
        console.log("render", config.name, " with value", value);
        return (
          <textarea
            ref={inputRef}
            value={value}
            onFocus={() => setSelectedConfig(config)}
            onChange={(e) => {
              console.log("onChange called", e.target.value);
              setNewValue(config.name, e.target.value);
            }}
          ></textarea>
        );
      }}
```

And check that when you click to reset to default the textarea gets the correct `value` and that the `onChange` does NOT fire (I'm wondering if somehow there's a loop that's updating the value right after it gets reset).

goulinkh:
@huwshimi Here is logs after typing `test` in the textfield and then clicking the default button:

```
render cluster-name  with value test
render cluster-name  with value undefined
```

However the textfield in the DOM still has the value test shown.

huwshimi:
@goulinkh I wonder if it doesn't like receiving `undefined` so I've updated this to set an empty string if there is no value. Could you try again with that change?

goulinkh:
@huwshimi Nice! it's working now

PR #1466, opened 2023-05-02
Fix production panels

huwshimi:
## Done

- Fix panels that weren't opening in in the built dashboard.

## QA

- Open the demo.
- Check that each panel opens:
 - Go to the controller list and click "Register a controller"
 - Go to the model list and click on a model and then click "Modify access"
 - On a model page click on an application and then click "Configure"
 - On the application page tick a unit and then click "Run action"


Lukewh (approved):
:+1: LGTM

PR #1504, opened 2023-06-28
WD-4697 - Use new types from jujulib

huwshimi:
## Done

- Update to use types supplied by patternProperties in jujulib.
- Remove TSFixMes for types that are no handled by jujulib.
- Replace factories with correct ones (these were exposed by the type updates).
- Add factories for new types.

## Details

https://warthogs.atlassian.net/browse/WD-4697

### DIFF
--- a/src/components/ModelTableList/ModelSummary/ModelSummary.test.tsx
+++ b/src/components/ModelTableList/ModelSummary/ModelSummary.test.tsx
@@ -4,8 +4,8 @@ import {
   modelDataApplicationFactory,
   modelDataFactory,
   modelDataUnitFactory,
+  modelDataMachineFactory,
 } from "testing/factories/juju/juju";
-import { machineChangeDeltaFactory } from "testing/factories/juju/model-watcher";
 import { renderComponent } from "testing/utils";
 
 import ModelSummary, { Label } from "./ModelSummary";
@@ -23,8 +23,8 @@ describe("ModelSummary", () => {
         }),
       },
       machines: {
-        "0": machineChangeDeltaFactory.build(),
-        "1": machineChangeDeltaFactory.build(),
+        "0": modelDataMachineFactory.build(),
+        "1": modelDataMachineFactory.build(),
       },
     });
     renderComponent(
--- a/src/components/RelationIcon/RelationIcon.tsx
+++ b/src/components/RelationIcon/RelationIcon.tsx
@@ -16,7 +16,12 @@ const RelationIcon = ({ applicationName, applications }: Props) => {
   const charmId =
     "charm-url" in application
       ? application["charm-url"]
-      : application["charm"];
+      : "charm" in application
+      ? application["charm"]
+      : null;
+  if (!charmId) {
+    return null;
+  }
   return <CharmIcon name={applicationName} charmId={charmId} />;
 };
 
--- a/src/juju/api.test.ts
+++ b/src/juju/api.test.ts
@@ -32,7 +32,6 @@ import {
   applicationInfoFactory,
   machineChangeDeltaFactory,
 } from "testing/factories/juju/model-watcher";
-import type { TSFixMe } from "types";
 
 import type { Config } from "./api";
 import {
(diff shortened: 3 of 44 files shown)


src/components/ModelTableList/ModelSummary/ModelSummary.test.tsx
```diff
@@ -23,8 +23,8 @@ describe("ModelSummary", () => {
         }),
       },
       machines: {
-        "0": machineChangeDeltaFactory.build(),
-        "1": machineChangeDeltaFactory.build(),
+        "0": modelDataMachineFactory.build(),
+        "1": modelDataMachineFactory.build(),
```
huwshimi, line 27: We were using the wrong type of factories for a bunch of these tests but now we get warned if they're wrong.

src/components/RelationIcon/RelationIcon.tsx
```diff
@@ -16,7 +16,12 @@ const RelationIcon = ({ applicationName, applications }: Props) => {
   const charmId =
     "charm-url" in application
       ? application["charm-url"]
-      : application["charm"];
+      : "charm" in application
```
huwshimi, line 19: Previously the application type was marked as always having all attributes, but there was a case where this might not be true, so I had to update everywhere to check if the application was the full object.

src/juju/api.test.ts
```diff
@@ -558,35 +557,13 @@ describe("Juju API", () => {
...
-            results: [
-              {
-                ...abc123.results[0],
-                result: {
-                  ...abc123.results[0].result,
-                  "agent-version": "2.3.4",
-                },
-              },
-            ],
-          },
```
huwshimi, line 572: The `"agent-version"` type was fixed on the jujulib side so we no longer have to do these merges.

src/juju/api.ts
```diff
@@ -324,28 +320,6 @@ async function fetchModelInfo(conn: ConnectionWithFacades, modelUUID: string) {
   return modelInfo;
 }
 
-const versionToString = (agentVersion: ModelManagerNumber) =>
-  `${agentVersion.Major}.${agentVersion.Minor}.${agentVersion.Patch}`;
-
-const toModelInfo = (modelInfo: JujuModelInfoResults): ModelInfoResults => ({
```
huwshimi, line 330: This isn't needed anymore as the "agent-version" type was fixed in jujulib.

src/juju/watchers.ts
```diff
@@ -49,23 +49,24 @@ export function generateModelWatcherBase(): ModelData {
   };
 }
 
-function _processDelta(
+function _processDelta<M extends ModelData, E extends keyof M>(
```
huwshimi, line 52: This method has been updated to infer the types of the deltas from the new types provided by jujulib.

src/juju/watchers.ts
```diff
@@ -128,18 +134,14 @@ export function processDeltas(
...
           if (!modelData.applications[applicationName]) {
-            // The following line is ignored because the type requires that
-            // the application has a number of pre-defined values.
-            // This is only here because sometimes the unit delta is parsed
-            // before the application delta arrives so it needs somewhere to
-            // store this information.
-            // This can be resolved by computing the unit-count in the
-            // selector instead of when the unit delta arrives.
-            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-            // @ts-ignore TSFixMe XXX
```
huwshimi, line 139: Instead of ignore this, the application type has been updated to either be the full application object or just the unit count, as this ignore meant that if the application only had the unit count it could break the UI.

src/store/general/selectors.ts
```diff
@@ -89,9 +88,7 @@ export const getActiveUserTag = createSelector(
     slice,
     (state, wsControllerURL) => getControllerConnection(state, wsControllerURL),
   ],
-  (_sliceState, controllerConnection) =>
-    // TSFixMe: jujulib types user as `object`.
-    (controllerConnection?.user as TSFixMe)?.identity
+  (_sliceState, controllerConnection) => controllerConnection?.user?.identity
```
huwshimi, line 91: These types are now correct in jujulib.

src/store/juju/types.ts
```diff
@@ -25,19 +28,14 @@ export type AdditionalController = {
 
 export type Controllers = Record<string, (Controller | AdditionalController)[]>;
 
-export type ModelInfo =
-  | Omit<JujuModelInfo, "agent-version"> & {
-      "agent-version": string;
```
huwshimi, line 30: "agent-version" is now set to a string in jujulib.

src/store/juju/types.ts
```diff
@@ -25,19 +28,14 @@ export type AdditionalController = {
...
-    };
-
-export type ModelData = {
-  applications: FullStatus["applications"];
+// There is some model data that we don't want to store from the full status because it changes
+// too often causing needless re-renders and is currently irrelevant
+// like controllerTimestamp.
+export type ModelData = Omit<
+  FullStatusWithAnnotations,
```
huwshimi, line 35: We can now use the model data type from jujulib, minus the bits of data we don't want to store.

PR #1509, opened 2023-06-30
Remove AdditionalProperties types

huwshimi:
## Done

- Replace AdditionalProperties types with references from parent types so that changes to types will be picked up.

### DIFF
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -1,7 +1,7 @@
 import type { Connection } from "@canonical/jujulib";
 import type ActionV7 from "@canonical/jujulib/dist/api/facades/action/ActionV7";
 import type AllWatcherV3 from "@canonical/jujulib/dist/api/facades/all-watcher/AllWatcherV3";
-import type { AdditionalProperties as AnnotationsAdditionalProperties } from "@canonical/jujulib/dist/api/facades/annotations/AnnotationsV2";
+import type { AnnotationsGetResult } from "@canonical/jujulib/dist/api/facades/annotations/AnnotationsV2";
 import type AnnotationsV2 from "@canonical/jujulib/dist/api/facades/annotations/AnnotationsV2";
 import type ApplicationV15 from "@canonical/jujulib/dist/api/facades/application/ApplicationV15";
 import type CharmsV6 from "@canonical/jujulib/dist/api/facades/charms/CharmsV6";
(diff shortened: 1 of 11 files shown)


src/testing/factories/juju/Application.ts
```diff
@@ -1,33 +1,34 @@
 import type {
-  AdditionalProperties,
   ApplicationGetResults,
   Base,
   Value,
   ErrorResults,
 } from "@canonical/jujulib/dist/api/facades/application/ApplicationV15";
```
vladimir-cucu, line 6: I've looked at the source code of `ApplicationV18` and it should work instead of `ApplicationV15`, while being the up to date version (as far as I understand). The only difference I've observed between V15 and V18 is that the `Base` type has an extra key `"image-id"`, which I am not sure how much would affect the dashboard.

src/testing/factories/juju/Application.ts
```diff
@@ -1,33 +1,34 @@
...
 import { Factory } from "fishery";
 
-export const applicationConfigFactory = Factory.define<AdditionalProperties>(
+export const applicationConfigFactory = Factory.define<
+  NonNullable<ApplicationGetResults["application-config"]>
+>(() => ({
+  trust: configFactory.build({
+    default: false,
+    description: "Does this application have access to trusted credentials",
+    source: "default",
```
vladimir-cucu, line 15: I'm not sure, but I think we could remove this line, as `source` will be inffered to be `"default"` from factory.

vladimir-cucu (commented):
Looks good to merge! The changes run without issues locally. I've tried to provided a few comments, however, I am unsure how relevant/correct they are. 😅

src/juju/types.ts
```diff
@@ -103,28 +103,20 @@ export enum ReduxDeltaEntityTypes {
...
-export type DeltaMessageData =
-  | ActionChangeDelta
-  | AnnotationChangeDelta
-  | ApplicationChangeDelta
-  | CharmChangeDelta
-  | UnitChangeDelta
-  | MachineChangeDelta
-  | ModelChangeDelta
-  | RelationChangeDelta;
+export type DeltaMessageData = AllWatcherDelta[2];
```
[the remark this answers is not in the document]
huwshimi: Great suggestion, I'll fix those up!

src/testing/factories/juju/Application.ts
huwshimi: Well spotted, as you say this won't have much impact so I've updated these to the latest types.

src/testing/factories/juju/Application.ts
huwshimi: Well spotted, I've fixed this too.

Issue #1506, opened 2023-06-29
Canvas modals are not in the foreground


huwshimi:
**Describe the bug**
For some reason the z-index on the modals has changed which has made them appear behind other elements.

**To Reproduce**
1. Go to a model details page
2. Click on the button to expand the canvas.
3. See that the modal is not at the front.

**Screenshots**
<img width="803" alt="Screenshot 2023-06-29 at 12 11 54 pm" src="https://github.com/canonical/juju-dashboard/assets/361637/cdd4c9e5-883d-4064-9824-fbeb7d64c03d">

PR #1521, opened 2023-07-10
WD-5043 - fix: Modals not appearing fully in the foreground

vladimir-cucu:
## Done

- Canvas modals appear completely on the foreground

## Details

Fixes: #1506 
https://warthogs.atlassian.net/browse/WD-5043

## Screenshots
Screenshot 2023-07-10 at 17 11 35

### DIFF
--- a/src/pages/EntityDetails/App/App.tsx
+++ b/src/pages/EntityDetails/App/App.tsx
@@ -277,7 +277,7 @@ export default function App(): JSX.Element {
 
   return (
     <>
-      <div>
+      <div className="entity-details__sidebar">
         <InfoPanel />
         <>
           <div className="entity-details__actions">
--- a/src/pages/EntityDetails/Model/Model.tsx
+++ b/src/pages/EntityDetails/Model/Model.tsx
@@ -135,45 +135,40 @@ const Model = () => {
 
   return (
     <>
-      <div>
-        {/* The sidebar needs to be within a wrapping div so that it does not
-            get given the full height of the grid so that the sidebar can be sticky.
-        */}
-        <div className="entity-details__sidebar">
-          <InfoPanel />
-          <div className="entity-details__actions">
-            {activeUser &&
-              canAdministerModelAccess(
-                activeUser,
-                modelStatusData?.info?.users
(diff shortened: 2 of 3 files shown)


huwshimi (commented):
@vladimir-cucu it looks like this approach breaks the sticky side panel. When there is long content the side panel should stay sticky like this:

Screen Recording 2023-07-11 at 10 37 19 am

Whereas this branch makes them scroll together:

Screen Recording 2023-07-11 at 10 40 00 am

A different approach would be to use a portal to render the modal to the top of the DOM so that the modal doesn't get cut off by the parent `overflow: hidden` (I presume that's what's happening here).

We tend to use the `usePortal` hook for portals, like this: https://github.com/canonical/react-components/blob/d7fe94f2f663ab98526aff07e51b87965a01694f/src/components/ConfirmationButton/ConfirmationButton.tsx#L39.

src/components/InfoPanel/_info-panel.scss
```diff
@@ -51,9 +42,13 @@
...
-    font-size: 0.75rem;
-    margin-bottom: 0.25rem;
-  }
+.p-muted-heading {
+  font-size: 0.75rem !important;
+  margin-bottom: 0.25rem !important;
+}
+
+.p-modal .p-modal__dialog {
+  width: 90%;
```
huwshimi, line 53: These should probably be scoped inside a class, something like `.info-panel__modal { ... }` and then we'd need to add the `info-panel__modal` class to the `<Modal>` component.

huwshimi (approved):
Nice work, just one little suggestion to keep the CSS from potentially effecting other components.

src/components/InfoPanel/_info-panel.scss
vladimir-cucu: Done! ✅  Also changed the location of `.p-muted-heading` to be inside `_entity-info.css` and slightly changed `_entity-info.css` to better represent the component hierarchy. Let me know if I should change it back.

huwshimi: That looks good!

Issue #1499, opened 2023-06-13
Clicking outside action confirmation has error


huwshimi:
**Describe the bug**
When you try to click outside a confirmation for performing an action it has an error: `Reduce of empty array with no initial value`.

**To Reproduce**
1. Go to an application inside a model
2. Tick a unit and click "Run action"
3. Choose an action and click "Run action"
4. Click outside the confirmation and it will display the error.

PR #1524, opened 2023-07-11
WD-5050 - fix: Clicking outside action confirmation has error

vladimir-cucu:
## Done

- Clicking outside action confirmation will close the action confirmation screen

## Details

Fixes: #1499 
https://warthogs.atlassian.net/browse/WD-5050


huwshimi (commented):
This has fixed it for one of the action panels, but it looks like the other panel needs updating as well.

If you go to a model and search for an app, then tick one of the apps and then click 'Run action' you'll get the other panel.

huwshimi (approved):
Works great!

Issue #1525, opened 2023-07-12
Skip charm selection if only one charm


huwshimi:
In the multiple app actions panel it should skip the charm/revision selection if there is only one option.

1. Go to the apps tab for a model.
2. Search for an app.
3. Tick an app (or multiple apps that have the same charm and revision).
4. Click 'Run action'.
5. The "Choose applications of charm" panel will appear with only one option.

If only one application was ticked, or if multiple applications were ticked that have the same charm and revision it should skip the "Choose applications of charm" step and go straight to the application selection.

<img width="1089" alt="Screenshot 2023-07-12 at 10 06 04 am" src="https://github.com/canonical/juju-dashboard/assets/361637/7307a9fc-24b9-4df4-8cab-daffebd59b08">

PR #1531, opened 2023-07-14
WD-5134 - feat: Skip charm selection if only one charm is available

vladimir-cucu:
## Done

- Skips charm selection panel if only one charm is available.

## Details
Fixes: #1525 
https://warthogs.atlassian.net/browse/WD-5134


src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+      data-testid={TestId.PANEL}
+      title={
+        charmURL ? (
+          <CharmActionsPanelTitle charmURL={charmURL} />
+        ) : (
+          Label.CHARMS_PANEL_TITLE
+        )
+      }
+      onRemovePanelQueryParams={handleRemovePanelQueryParams}
+      loading={charmURL === undefined}
```
huwshimi, line 85: It might be nice to leave a comment here about why this needs to be `undefined` not any falsey value.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+    >
+      <>
+        {charmURL && <CharmActionsPanel charmURL={charmURL} />}
+        {!charmURL && <CharmsPanel onCharmURLChange={setCharmURL} />}
```
huwshimi, line 89: This could be `{charmURL ? <CharmActionsPanel charmURL={charmURL} /> : <CharmsPanel onCharmURLChange={setCharmURL} />}`.

src/panels/CharmsPanel/CharmsPanel.test.tsx
```diff
@@ -39,27 +39,41 @@ describe("CharmsPanel", () => {
...
+      path,
+      url,
+      state,
+    });
     expect(screen.getByRole("button", { name: "Next" })).toBeDisabled();
   });
 
   it("next button is enabled when a charm is selected", () => {
-    renderComponent(<CharmsPanel />, { path, url, state });
+    renderComponent(<CharmsPanel onCharmURLChange={() => {}} />, {
```
huwshimi, line 60: This will currently make no difference to these tests, but the convention is to use `jest.fn()` to provide functions for props in tests. The benefit will be that if the function type changes (e.g. a return type is added) the `() => {}` will no longer satisfy the type, but `jest.fn()` will.

huwshimi (approved):
The experience is so much nicer now. I've left a couple of really small comments.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.tsx
vladimir-cucu: Good remark! We check for `charmURL === undefined` in two different places, thus I think that creating `const isPanelLoading = !isSet(charmURL)` would improve the code.

src/panels/CharmsPanel/CharmsPanel.test.tsx
vladimir-cucu: I never thought about using `jest.fn()` instead of `() => {}` from a "maintainability" standpoint. Thanks for the remark!

PR #1535, opened 2023-07-18
Rename Logs tab

huwshimi:
## Done

- Rename Actions Logs tab to Logs in preparation for adding Audit logs.

## QA

- Go to a model and you should see a "Logs" tab where it used to say "Action Logs".
- Click on the Logs tab and you should see the action logs with a switcher with one item "Action Logs".

## Details

https://warthogs.atlassian.net/browse/WD-5142


huwshimi:
> Looks good to me! 🙂
> 
> One thing that I've observed (which is really nit-picky and not really worth going into) is that the loading icon for `Action logs` screen looks out of place. I am not sure if it is intended for it to be this way.

Thanks, I've fixed it!

PR #1537, opened 2023-07-20
Add yarn proxy config

samhotep:
## Done

- Added squid.internal as a yarn proxy

## QA

- Check that demo has installed
- Pull code
- Open http://0.0.0.0:8036/

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,3 +1,5 @@
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-3.6.0.cjs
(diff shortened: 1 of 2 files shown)


samhotep:
@huwshimi One issue is that locally the proxy isn't needed so will have to be manually edited out (unless using VPN). Atm I'm not certain of an elegant way to differentiate between the environments

huwshimi:
Thanks so much @samhotep! I've found a way to set the config based off the env vars: https://github.com/canonical/juju-dashboard/pull/1530.

PR #1545, opened 2023-07-27
WD-5215 - feat: Hide model column from table in model logs view

vladimir-cucu:
## Done

- Updated the AuditLogsTable to take a prop to hide the Model column.
- Set the prop to hide the model column on the model level audit logs view.

## Details

https://warthogs.atlassian.net/browse/WD-5215

### DIFF
--- a/src/components/AuditLogsTable/AuditLogsTable.tsx
+++ b/src/components/AuditLogsTable/AuditLogsTable.tsx
@@ -9,7 +9,7 @@ import { generateFakeAuditLogs } from "pages/Logs/audit-logs-fake-data";
 import getUserName from "utils/getUserName";
 
 type Props = {
-  showModel: boolean;
+  showModel?: boolean;
 };
 
 const COLUMN_DATA: Column[] = [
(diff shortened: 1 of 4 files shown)


src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -8,6 +8,10 @@ import { formatFriendlyDateToNow } from "components/utils";
 import { generateFakeAuditLogs } from "pages/Logs/audit-logs-fake-data";
 import getUserName from "utils/getUserName";
 
+type Props = {
+  showModel: boolean;
```
huwshimi, line 12: This could be an optional prop so that we don't have to provide it when we don't want to show the model.

src/pages/EntityDetails/Model/ApplicationsTab/ApplicationsTab.test.tsx
```diff
@@ -1,6 +1,5 @@
 import { screen } from "@testing-library/react";
 
-import { actions as jujuActions } from "store/juju";
```
huwshimi, line 3: Thanks!

huwshimi (approved):
One small suggestion.

PR #1547, opened 2023-07-28
WD-5211 - feat: Add Logs table to top level Logs page

vladimir-cucu:
## Done

- Added Audit Logs table to top level Audit Logs page.
- Fixed main content UI overflow on desktop and mobile.

## Details

https://warthogs.atlassian.net/browse/WD-5211

## Screenshots

Screenshot 2023-07-28 at 17 01 41


src/scss/_breakpoints.scss
```diff
@@ -2,7 +2,7 @@
 
 // 0px -> mobile
 @mixin mobile {
-  @media screen and (max-width: $breakpoint-small) {
+  @media screen and (max-width: $breakpoint-small - 1px) {
```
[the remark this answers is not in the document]
huwshimi: That seems fine in my testing.

src/layout/BaseLayout/_base-layout.scss
```diff
@@ -61,6 +61,21 @@
...
+      display: flex;
+      flex-direction: column;
+      height: 100%;
+
+      .l-content {
+        flex: 1;
+        min-height: 0;
+      }
+    }
+  }
```
huwshimi, line 78:
```suggestion
```


Instead add the following after line [162](https://github.com/canonical/juju-dashboard/pull/1547/files#diff-fd8132c704f92ac7d42cd11b60ec0393b0d08e6655e2e72c8d24c99d21909572R162):

```css
  .l-navigation.is-pinned {
    @include desktop {
      position: sticky !important;
    }
  }
```

src/pages/Logs/_logs.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.logs {
+  @include mobile {
+    overflow-x: scroll;
+  }
+
+  @include desktop {
+    overflow-y: scroll;
+  }
+}
```
huwshimi, line 12:
```suggestion
```

src/pages/Logs/Logs.tsx
```diff
@@ -1,11 +1,20 @@
+import FadeIn from "animations/FadeIn";
+import AuditLogsTable from "components/AuditLogsTable/AuditLogsTable";
 import Header from "components/Header/Header";
 import BaseLayout from "layout/BaseLayout/BaseLayout";
 
+import "./_logs.scss";
+
```
huwshimi, line 7:
```suggestion
```

huwshimi (commented):
I've made some suggestions here to improve the scrolling a little by making the sidebar sticky instead of changing the content scrolling (hopefully I've understood the issue you were fixing).

There were two issues I spotted, one was that the scroll area was nested inside the content a little, like this:


Screenshot 2023-07-31 at 10 11 55 am


Whereas with my suggestion the content extends to the edge of the screen:


Screenshot 2023-07-31 at 10 30 31 am


I also noticed that for some reason the border under the model tabs was hidden:


Screenshot 2023-07-31 at 10 12 51 am


And lastly the changes to make the logs scroll would also need to be applied to the controllers and models lists, whereas by making the sidebar sticky it applies to all pages without changes.

src/layout/BaseLayout/_base-layout.scss
vladimir-cucu: Thanks for pointing out this change! It is definitely simpler and better than my approach, and scales well with different contents. I've added also `overflow-x: scroll` to `.l-main` in order for the user to be able to scroll horizontally.

src/layout/BaseLayout/_base-layout.scss
```diff
@@ -122,6 +123,12 @@
       }
     }
   }
+
+  &.is-pinned {
+    @include desktop {
+      position: sticky !important;
+    }
+  }
```
huwshimi, line 131: I might have misled you slightly, to prevent having to use `!important` I think you can put the following inside the `.l-application { ... }` below instead of here:

```css
  .l-navigation.is-pinned {
    @include desktop {
      position: sticky;
    }
  }
```

src/layout/BaseLayout/_base-layout.scss
```diff
@@ -51,6 +51,7 @@
   // Override the Vanilla default that sets overflow-y to auto but this prevents
   // sticky from being used by the the entity details sidebar.
   overflow-y: visible !important;
+  overflow-x: scroll;
```
huwshimi, line 54: Making the scrolling change here means that everything in the content scrolls, including the header:

Screenshot 2023-08-01 at 10 08 12 am

The other tables get wrapped in an element with `u-overflow--auto` so that only the table scrolls (I'd be keen for us to investigate if we can move away from horizontal scrolling of the tables at some stage but it would require global changes). I've made another comment about where to add this.

src/pages/Logs/Logs.tsx
```diff
@@ -6,6 +8,11 @@ const Logs = (): JSX.Element => (
     <Header>
       <b>Audit logs</b>
     </Header>
+    <div className="l-content logs">
+      <FadeIn isActive={true}>
+        <AuditLogsTable showModel />
```
huwshimi, line 13: To make only the table scroll this needs to be wrapped in `<div className="u-overflow--auto">...</div>`.

huwshimi (approved):
Looking good, just a few little improvements.

PR #1551, opened 2023-08-01
WD-5218 - Fetch audit logs

huwshimi:
## Done

- Fetch audit logs in the table component.
- Create a reusable spinner component.

## QA

- Set your config.local.js to use staging.
- Open `src/pages/Logs/Logs.tsx`.
- Add `<AuditLogsTable />` below the header.
- Load the app and go to the top level logs page.
- You should see the logs that get fetched from JIMM.

## Details

https://warthogs.atlassian.net/browse/WD-5218


src/components/LoadingSpinner/_loading-spinner.scss
```diff
@@ -0,0 +1,8 @@
+@import "../../scss/variables";
+
+.loading-spinner {
```
huwshimi, line 3: Similar loading alignment was used in a few places. I think this will need a followup PR to make some tweaks after https://github.com/canonical/juju-dashboard/pull/1547 lands.

src/store/store.ts
```diff
@@ -50,8 +51,11 @@ export const useAppDispatch: () => AppDispatch = useDispatch;
 // This hook can be used in place of useDispatch to get correctly typed dispatches that return promises.
 export const usePromiseDispatch = () => {
   const dispatch = useAppDispatch();
-  return <Result>(action: AnyAction) =>
-    (dispatch as (action: AnyAction) => Promise<Result>)(action);
+  return useCallback(
+    <Result>(action: AnyAction) =>
+      (dispatch as (action: AnyAction) => Promise<Result>)(action),
+    [dispatch]
+  );
```
huwshimi, line 58: This was causing the useEffect in the table component to rerun as it was recreating the function on each render.

src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -41,15 +49,37 @@ const COLUMN_DATA: Column[] = [
 
 const AuditLogsTable = ({ showModel = false }: Props) => {
   const { modelName } = useParams<EntityDetailsRoute>();
-  const additionalEmptyMsg = showModel ? ` for ${modelName}` : "";
+  const [auditLogs, setAuditLogs] = useState<AuditEvent[]>([]);
+  const [loading, setLoading] = useState(false);
```
vladimir-cucu, line 53: There is a small flicker (showing table with no data and then the `LoadingSpinner`) before the first `useEffect` is triggered and makes `loading = true`. Maybe we could make `true` to be the default state of `loading`?

vladimir-cucu (approved):
Looks good to me! Adding the `LoadingSpinner` seems like a really good idea. 🙂

src/components/AuditLogsTable/AuditLogsTable.test.tsx
```diff
@@ -1,40 +1,128 @@
...
+    });
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("should have model as second header when showing the model data", async () => {
+    renderComponent(<AuditLogsTable showModel={true} />, { state });
+    expect((await screen.findAllByRole("columnheader"))[1]).toHaveTextContent(
```
huwshimi, line 50: Once I rebased I had to update a number of tests to fetch the audit logs and wait for the promise to resolve.

src/components/AuditLogsTable/AuditLogsTable.tsx
huwshimi: Done!

PR #1554, opened 2023-08-03
Add AutocompleteInput component

huwshimi:
## Done

- Add a component that uses `datalist` to provide suggestions to an input.


vladimir-cucu (approved):
LGTM!

PR #1561, opened 2023-08-10
WD-5220 - Add audit logs filter panel

huwshimi:
## Done

- Add a panel for filtering audit logs (this doesn't refetch from the API yet as it needs a couple of other PRs to land first).

## QA

- Connect your dashboard to the staging JIMM.
- Go to the Logs page.
- Click the filter button.
- Click on some of the fields and depending on what data you have it might display suggestions.
- Fill out some fields and click the Filter button.
- You should see the currently applied filter chips.
- Dismiss some chips and the filters should change.
- Connect your dashboard to the production JIMM (the default JAAS websocket).
- Go to a model and click the Logs tab and then the Audit logs toggle.
- You should be able to open the filter panel from the new action bar.

## Details

https://warthogs.atlassian.net/browse/WD-5220


src/components/AuditLogsTable/AuditLogsTable.test.tsx
```diff
@@ -84,30 +83,6 @@ describe("AuditLogsTable", () => {
     expect(columnHeaders[5]).toHaveTextContent("facade version");
   });
 
-  it("should not fetch audit logs if there is no websocket", () => {
```
huwshimi, line 87: These tests have been moved to the `useFetchAuditEvents` hook.

src/components/AuditLogsTable/AuditLogsTable.test.tsx
```diff
@@ -132,21 +107,13 @@ describe("AuditLogsTable", () => {
     expect(cells[5]).toHaveTextContent("3");
   });
 
-  it("should show refresh button", async () => {
```
huwshimi, line 135: These tests have been moved to the `AuditLogsTableActions` tests.

src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -55,29 +53,15 @@ const COLUMN_DATA: Column[] = [
...
-    page: null,
-  });
-
-  const fetchAuditEvents = useCallback(() => {
-    if (!wsControllerURL || !hasControllerConnection) {
-      return;
-    }
-    dispatch(jujuActions.fetchAuditEvents({ wsControllerURL }));
-  }, [dispatch, hasControllerConnection, wsControllerURL]);
+  const fetchAuditEvents = useFetchAuditEvents();
```
huwshimi, line 64: I've moved all this to the `useFetchAuditEvents` so that it can also be called by the refresh button now that it's in the action bar.

src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -121,15 +105,8 @@ const AuditLogsTable = ({ showModel = false }: Props) => {
...
   return (
-    <div className="audit-logs__table">
-      <Button
-        onClick={() => {
-          fetchAuditEvents();
-          setQueryParams({ page: "1" });
-        }}
-      >
-        Refresh
-      </Button>
```
huwshimi, line 132: This button is now in the action bar.

src/components/AuditLogsTable/hooks.ts
```diff
@@ -0,0 +1,23 @@
...
+  const wsControllerURL = useSelector(getWSControllerURL);
+  const hasControllerConnection = useAppSelector((state) =>
+    getControllerConnection(state, wsControllerURL)
+  );
+  return useCallback(() => {
+    if (!wsControllerURL || !hasControllerConnection) {
+      return;
+    }
+    dispatch(jujuActions.fetchAuditEvents({ wsControllerURL }));
+  }, [dispatch, hasControllerConnection, wsControllerURL]);
```
huwshimi, line 22: This will get updated with the changes in the pagination PR once that lands.

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
```diff
@@ -0,0 +1,117 @@
...
+  AFTER = "After",
+  BEFORE = "Before",
+  FACADE = "Facade",
+  METHOD = "Method",
+  MODEL = "Model",
+  USER = "User",
+  VERSION = "Version",
+}
+
+const Fields = (): JSX.Element => {
```
huwshimi, line 27: These fields are in a separate component from the form so that we can use `useFormikContext`.

src/components/AuditLogsTable/AuditLogsTableFilters/AuditLogsTableFilters.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+    return chips;
+  }, []);
+
+const AuditLogsTableFilters = () => {
+  const [filters, setFilters] = useQueryParams<AuditLogFilters>({
+    // Spread so that the default doesn't get updated when the filters change.
+    ...DEFAULT_AUDIT_LOG_FILTERS,
+  });
+  const hasFilters = Object.values(filters).some((filter) => !!filter);
```
vladimir-cucu, line 53: I don't think this is an issue, but rather it is a small inconsistency. In case we remove all the filters manually, the `queryParams` would still leak into the url without any values. At the same time, if we remove the filters by pressing the `Clear` button, the `queryParams` would not leak into the url. I think that adding the following `useEffect` here would fix the aforementioned inconsistency:

```
 useEffect(() => {
    if (!hasFilters) {
      setFilters(DEFAULT_AUDIT_LOG_FILTERS);
    }
  }, [hasFilters, setFilters]);
```

vladimir-cucu (approved):
Looks good to me! I've added just a small optional remark.

src/components/AuditLogsTable/AuditLogsTableFilters/AuditLogsTableFilters.tsx
huwshimi: Thanks, it turns out that these blank query params were appearing because when I apply the filters I use the form values which are empty strings and `useQueryParams` will set them in the URL unless they are `null` or `undefined`.

I've updated it to pass `null` for blank filters and added a couple of tests.

PR #1562, opened 2023-08-11
WD-5221- Fetch when filters change

huwshimi:
## Done

- Refetch the audit logs when the filters change.

## QA

- Connect to staging.
- Go to the logs page.
- Open the filters panel and enter some filters (note: facade name and version aren't currently supported - I've added a Jira task for this - and the facade method does not appear to be changing anything in the response from the API).
- Click Filter and the logs should reload.

## Details

https://warthogs.atlassian.net/browse/WD-5221

### DIFF
--- a/src/components/AuditLogsTable/AuditLogsTableActions/AuditLogsTableActions.test.tsx
+++ b/src/components/AuditLogsTable/AuditLogsTableActions/AuditLogsTableActions.test.tsx
@@ -71,7 +71,7 @@ describe("AuditLogsTableActions", () => {
     await userEvent.click(screen.getByRole("button", { name: "Next page" }));
     expect(window.location.search).toEqual("?page=2");
     await userEvent.click(screen.getByRole("button", { name: "Refresh" }));
-    expect(window.location.search).toEqual("?page=1");
+    expect(window.location.search).toEqual("");
   });
 
   it("should navigate to next page and then to previous page", async () => {
@@ -101,6 +101,6 @@ describe("AuditLogsTableActions", () => {
     await userEvent.click(dropdownMenu);
     await userEvent.selectOptions(dropdownMenu, "100/page");
     expect(dropdownMenu).toHaveTextContent("100/page");
-    expect(window.location.search).toEqual("?limit=100&page=1");
+    expect(window.location.search).toEqual("?limit=100");
   });
 });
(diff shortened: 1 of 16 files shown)


src/components/AuditLogsTable/AuditLogsTableActions/AuditLogsTableActions.test.tsx
```diff
@@ -71,7 +71,7 @@ describe("AuditLogsTableActions", () => {
     await userEvent.click(screen.getByRole("button", { name: "Next page" }));
     expect(window.location.search).toEqual("?page=2");
     await userEvent.click(screen.getByRole("button", { name: "Refresh" }));
-    expect(window.location.search).toEqual("?page=1");
+    expect(window.location.search).toEqual("");
```
huwshimi, line 74: In a number of places I've changed this to remove the page param so that it uses the default provided from the hook rather than explicitly setting it back to page 1.

src/store/juju/selectors.ts
```diff
@@ -152,8 +152,20 @@ export const getModelList = createSelector(
...
+export const getFullModelNames = createSelector(
+  [getModelList, getControllerData],
+  (modelList, controllers) =>
+    controllers
+      ? Object.values(modelList)?.reduce<string[]>((modelNames, model) => {
+          const controller =
+            model.wsControllerURL in controllers &&
+            controllers[model.wsControllerURL][0];
+          if (controller && "name" in controller) {
+            modelNames.push(`${controller.name}/${model.name}`);
```
huwshimi, line 164: This is the model name format that the API expects.

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
```diff
@@ -24,13 +25,15 @@ export enum Label {
   VERSION = "Version",
 }
 
+export const DATETIME_LOCAL = "yyyy-MM-dd'T'hh:mm";
```
vladimir-cucu, line 28: I think we need `HH` instead of `hh` here. The error that we were facing on Friday, where we could not insert the current time, most likely came from here, as `hh` represents the hour in 12-hour clock format instead of 24-hour clock like `HH` does.

vladimir-cucu (approved):
Looks good to me! I left a small comment that could potentially fix the bug we were facing on Friday.

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
huwshimi: Ah thankyou!

PR #1564, opened 2023-08-13
WD-5745 - Remove facade name and version filters

huwshimi:
## Done

- Remove the facade name and version filters as they are not supported by the API.

## QA

- Go to the logs page.
- Open the filter panel.
- See that the facade name and version inputs are no longer there.

## Details

https://warthogs.atlassian.net/browse/WD-5745

### DIFF
--- a/src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
+++ b/src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
@@ -7,7 +7,6 @@ import { useParams } from "react-router-dom";
 import AutocompleteInput from "components/AutocompleteInput";
 import type { EntityDetailsRoute } from "components/Routes/Routes";
 import {
-  getAuditEventsFacades,
   getAuditEventsMethods,
   getAuditEventsUsers,
   getAuditEventsModels,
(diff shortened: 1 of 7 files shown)


src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
```diff
@@ -18,11 +17,9 @@ import type { FormFields } from "../types";
 export enum Label {
   AFTER = "After",
   BEFORE = "Before",
-  FACADE = "Facade",
   METHOD = "Method",
```
vladimir-cucu, line 20: The column in the `Audit logs` table is named `Facade Method`, so I was wondering if maybe, for consistency reasons, we can name the label here the same. Although, at the same time, we send this as `method` in the request, so I think it can be left as it is as well.

vladimir-cucu (approved):
Looks good to me! I've left an optional comment. 🙂

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
huwshimi: I think naming the field the same as the table makes sense. I'll update it.

PR #1565, opened 2023-08-14
WD-5780 - Add pagination to bottom of table

huwshimi:
## Done

- Add back to first page button.
- Add pagination to bottom of table.

## QA

- Go to the logs page.
- Scroll to the bottom of the page.
- You should see pagination buttons.
- Navigate forward.
- Click the back to start button and it should take you to the first page.

## Details

https://warthogs.atlassian.net/browse/WD-5780


src/components/AuditLogsTable/AuditLogsTablePagination/AuditLogsTablePagination.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+            }),
+          ],
+          loaded: true,
+        }),
+      }),
+    });
+  });
+
+  it("should navigate to next page and then to previous page", async () => {
+    state.juju.auditEvents.items = auditEventFactory.buildList(51);
```
huwshimi, line 48: I remembered that the factories have `buildList()` instead of having to create them in a loop.

src/components/AuditLogsTable/AuditLogsTableActions/AuditLogsTableActions.test.tsx
```diff
@@ -56,51 +56,4 @@ describe("AuditLogsTableActions", () => {
     await userEvent.click(screen.getByRole("button", { name: Label.FILTER }));
     expect(window.location.search).toEqual("?panel=audit-log-filters");
   });
-
```
huwshimi, line 59: Moved to the pagination component.

src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -127,10 +129,11 @@ const AuditLogsTable = ({ showModel = false }: Props) => {
         <ModularTable
           columns={columnData}
           // Table will display at most (limit) entries.
-          data={tableData.length <= limit ? tableData : tableData.slice(0, -1)}
+          data={hasNextPage ? tableData.slice(0, -1) : tableData}
           emptyMsg={emptyMsg}
         />
       )}
+      {hasNextPage && auditLogsLoaded ? <AuditLogsTablePagination /> : null}
```
huwshimi, line 136: I added the pagination to the bottom of the page as well as the top as it felt a bit unnatural to have to scroll back to the top of the page to navigation to the next page.

src/components/AuditLogsTable/AuditLogsTablePagination/AuditLogsTablePagination.tsx
```diff
@@ -0,0 +1,99 @@
...
+        hasIcon
+        onClick={() => {
+          setQueryParams({ page: null });
+          scrollToTop();
+        }}
+      >
+        <Icon className="p-icon--rotate-270" name="back-to-top">
+          {Label.FIRST_PAGE}
+        </Icon>
+      </Button>
```
huwshimi, line 81: I added this button to go back to the first page as it seemed like this was an action I was wanting to take a number of times, and the only way to do it was refresh or click the Logs link in the side nav.

src/components/AuditLogsTable/AuditLogsTable.tsx
vladimir-cucu: That's a really good addition from a user experience perspective! However, while this is highly subjective, I was wondering if maybe making `AuditLogsTablePagination` sit on the bottom right would be "better" from a consistency standpoint? This way, both instances of this component would appear on the right hand side.

src/components/AuditLogsTable/AuditLogsTablePagination/AuditLogsTablePagination.tsx
vladimir-cucu: There is a small issue with this button and the `Filter` button when the horizontal width of the page is small. Both of them will grow in height.

vladimir-cucu (approved):
Looks good to me! Added 2 optional comments. 🙂

src/components/AuditLogsTable/AuditLogsTable.tsx
huwshimi: I think this is a good suggestion, but as far as I'm aware all the pagination at the bottom of tables in our other products is either on the left or centred, so I might leave this where it is for consistency with our other products for now and maybe we can look at this again when this feature gets some input from design.

PR #1566, opened 2023-08-14
WD-5747 - Filter logs by the current model model

huwshimi:
## Done

- Filter by model and hide model filter when viewing model audit logs.

## QA

- Connect to JAAS (not staging - this is so you can get to a model page).
- Open your network tab in your dev tools and find the controller websocket connection (it should be the one with the JIMM requests).
- Go to the logs page.
- In your controller websocket you should see the `FindAuditEvents` with no `model` param.
- Click the filter button.
- In the filter panel you should see the model name input.
- Enter a model name and click the Filter button.
- In your controller websocket you should see the `FindAuditEvents` with the `model` param set to the value you entered.
- Go to the audit logs tab inside a model.
- In your controller websocket you should see the `FindAuditEvents` with the `model` param set to the current model.
- Open the filter panel and there shouldn't be a model input.

## Details

https://warthogs.atlassian.net/browse/WD-5747
https://warthogs.atlassian.net/browse/WD-5222


src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -54,8 +50,9 @@ const COLUMN_DATA: Column[] = [
   },
 ];
 
-const AuditLogsTable = ({ showModel = false }: Props) => {
+const AuditLogsTable = () => {
   const { modelName } = useParams<EntityDetailsRoute>();
+  const showModel = !modelName;
```
huwshimi, line 55: Instead of having `showModel` props I've changed these components to instead check if there is a model name in the current route.

huwshimi (commented):
@vladimir-cucu oops, I forgot you had the task for hiding the model filter assigned to you. I can remove my implementation if you've already done this.

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
```diff
@@ -77,16 +81,19 @@ const Fields = (): JSX.Element => {
...
-        id={Label.MODEL}
-        label={Label.MODEL}
-        name="model"
-        as={AutocompleteInput}
-        options={models}
-      />
+      {showModel ? (
+        <Field
+          type="text"
+          help='A model name in the format "controller-name/model-name"'
```
vladimir-cucu, line 87: Looked in other places and I think, for consistency reasons, it could be better to end the help message with a period.

vladimir-cucu (approved):
Looks good to me and works as expected! Left 2 optional comments here as well. 🙂

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
huwshimi: Done!

src/panels/AuditLogsFilterPanel/Fields/Fields.tsx
```diff
@@ -77,16 +81,19 @@ const Fields = (): JSX.Element => {
...
-        type="text"
-        // Have to manually set the id until this issue has been fixed:
-        // https://github.com/canonical/react-components/issues/957
-        id={Label.MODEL}
-        label={Label.MODEL}
-        name="model"
-        as={AutocompleteInput}
-        options={models}
-      />
+      {showModel ? (
```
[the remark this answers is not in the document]
huwshimi: We actually do the inverse: we only show this field when there is no model present (i.e. this field is only displayed on the top level Logs page). `jujuModels` will be an array of all models that the controller returns (it gets the names from `juju.models` in the Redux store).

Hopefully I've understood you correctly, let me know if not!

PR #1567, opened 2023-08-14
refactor: Audit logs limit redux

vladimir-cucu:
## Done

- Store limit of AuditLogs in Redux instead of query params.

## Details

https://warthogs.atlassian.net/browse/WD-5744


src/components/AuditLogsTable/AuditLogsTableActions/AuditLogsTableActions.test.tsx
```diff
@@ -101,6 +101,5 @@ describe("AuditLogsTableActions", () => {
     await userEvent.click(dropdownMenu);
     await userEvent.selectOptions(dropdownMenu, "100/page");
     expect(dropdownMenu).toHaveTextContent("100/page");
-    expect(window.location.search).toEqual("?limit=100");
```
huwshimi, line 104: As you mentioned in the PR description, this should check that the `updateAuditEventsLimit` action was dispatched. There's an example of how we check that actions are dispatched here: https://github.com/canonical/juju-dashboard/blob/audit-logs/src/components/AuditLogsTable/AuditLogsTable.test.tsx#L87-L97.

src/components/AuditLogsTable/hooks.ts
```diff
@@ -23,14 +24,12 @@ export const useFetchAuditEvents = () => {
...
       page: string;
-      limit: string;
     } & AuditLogFilters
   >({
     page: DEFAULT_PAGE,
-    limit: DEFAULT_LIMIT_VALUE.toString(),
     ...DEFAULT_AUDIT_LOG_FILTERS,
   });
-  const limit = Number(queryParams.limit);
+  const limit = Number(useSelector(getAuditEventsLimit));
```
huwshimi, line 32: It looks like`getAuditEventsLimit` will return a number, so no need to wrap this in `Number(...)`

src/store/juju/slice.ts
```diff
@@ -139,6 +142,9 @@ const slice = createSlice({
       state.auditEvents.loaded = false;
       state.auditEvents.loading = false;
     },
+    updateAuditEventsLimit: (state, { payload }: PayloadAction<number>) => {
```
huwshimi, line 145: This will need a test in https://github.com/canonical/juju-dashboard/blob/audit-logs/src/store/juju/actions.test.ts

src/store/middleware/check-auth.ts
```diff
@@ -14,17 +14,24 @@ import type { RootState, Store } from "store/store";
 import { actions as uiActions } from "store/ui";
 
 function error(name: string, wsControllerURL: string) {
-  console.log(
-    "unable to perform action:",
+  console.error(
```
huwshimi, line 17: Thanks!

src/components/AuditLogsTable/AuditLogsTable.tsx
```diff
@@ -60,6 +60,7 @@ const AuditLogsTable = ({ showModel = false }: Props) => {
   const auditLogs = useSelector(getAuditEvents);
   const auditLogsLoaded = useSelector(getAuditEventsLoaded);
   const auditLogsLoading = useSelector(getAuditEventsLoading);
+  const auditLogsLimit = Number(useSelector(getAuditEventsLimit));
```
huwshimi, line 63: I don't think this needs `Number(...)` does it? I'm pretty sure it should return a number according to the type.

huwshimi (approved):
This works well, just a few small improvements.

PR #1569, opened 2023-08-16
WD-5797 - feat: Add ​​CrossModelQuery to JIMM API

vladimir-cucu:
## Done

- Added CrossModelQuery to JIMM API.

## Details

https://warthogs.atlassian.net/browse/WD-5797


src/juju/jimm-facade.ts
```diff
@@ -3,10 +3,24 @@ import { autoBind } from "@canonical/jujulib/dist/api/utils";
 
 import type { Controller } from "store/juju/types";
 
+// As typed in JIMM:
+// https://github.com/canonical/jimm/blob/1da76ed2d8dba741f5880f32c073f85f7d518904/api/params/params.go#L344
```
vladimir-cucu, line 7: This link and the one on line 14 can only be accessed internally. Is this ok, considering the open source nature of this project and that users can't access it externally?

src/store/juju/reducers.test.ts
```diff
@@ -172,6 +173,71 @@ describe("reducers", () => {
...
+        errors,
+        loaded: true,
+        loading: false,
+      }),
+    });
+  });
+
+  it("clearCrossModelQuery", () => {
+    const results = { mockResultKey: "mockResultValue" };
+    const errors = { mockErrorKey: "mockErrorValue" };
```
vladimir-cucu, line 221: We repeat these in a few places. Should we add them to factory?

src/testing/factories/juju/juju.ts
```diff
@@ -152,7 +153,17 @@ export const modelDataFactory = Factory.define<ModelData>(() => ({
   "remote-applications": {},
 }));
 
+export const crossModelQueryStateFactory = Factory.define<CrossModelQueryState>(
+  () => ({
+    results: null,
+    errors: null,
```
vladimir-cucu, line 159: These are not necessary in every place they appear, but I thought that the tests become more descriptive if we include them.

src/store/juju/slice.ts
```diff
@@ -108,6 +122,27 @@ const slice = createSlice({
...
+      action: PayloadAction<CrossModelQueryRequest & WsControllerURLParam>
+    ) => {
+      state.crossModelQuery.loading = true;
+    },
+    updateCrossModelQuery: (
+      state,
+      { payload }: PayloadAction<CrossModelQueryResponse>
+    ) => {
+      state.crossModelQuery.results = payload.results;
+      state.crossModelQuery.errors = payload.errors;
```
huwshimi, line 136: In the case that the payload is a string it should be stored in the error.

src/juju/jimm-facade.ts
```diff
@@ -3,10 +3,24 @@ import { autoBind } from "@canonical/jujulib/dist/api/utils";
...
+  type: string;
+  query: string;
+};
+
+// As typed in JIMM:
+// https://github.com/canonical/jimm/blob/1da76ed2d8dba741f5880f32c073f85f7d518904/api/params/params.go#L353
+export type CrossModelQueryResponse = {
+  results: Record<string, any>;
+  errors: Record<string, string>;
+};
```
huwshimi, line 18: It appears that if you pass an invalid type or query the response is just a string. So that you can refer to the `errors` and `results` types in store/juju/types.ts it might be good to keep this object type separate under a different name e.g.:

```suggestion
export type CrossModelQueryFullResponse = {
  results: Record<string, unknown[]>;
  errors: Record<string, string[]>;
};

export type CrossModelQueryResponse = CrossModelQueryFullResponse | string;
```

src/store/juju/types.ts
```diff
@@ -53,7 +54,15 @@ export type ModelsList = {
   [uuid: string]: ModelListInfo;
 };
 
+export type CrossModelQueryState = {
+  results: CrossModelQueryResponse["results"] | null;
+  errors: CrossModelQueryResponse["errors"] | null;
```
huwshimi, line 59: These would need to refer to the object type using the name that you set in jimm-facade and the errors type would also need to support a string.

```suggestion
  results: CrossModelQueryFullResponse["results"] | null;
  errors: CrossModelQueryFullResponse["errors"] | string | null;
```

src/testing/factories/juju/juju.ts
huwshimi: Yep, this is good as it is.

huwshimi (approved):
This is looking good, just a few little improvements to make.

src/juju/api.ts
```diff
@@ -774,16 +774,17 @@ export async function getCharmsURLFromApplications(
   return charms.filter((charm) => !!charm).map((charm) => charm?.url);
 }
 
-export function findCrossModelQuery(
+export function crossModelQuery(
```
vladimir-cucu, line 777: I left the name as `crossModelQuery` as I couldn't really think of anything better. If you have some proposals, don't hesitate to tell them! 🙂

huwshimi: I think that's fine!

src/store/juju/slice.ts
```diff
@@ -108,6 +123,31 @@ const slice = createSlice({
     clearControllerData: (state) => {
       state.controllers = {};
     },
+    fetchCrossModelQuery: (
+      state,
+      action: PayloadAction<CrossModelQueryRequest & WsControllerURLParam>
```
huwshimi, line 128: I should have noticed this in my previous review, but I think this should just accept `query` as the `type` param is provided in the facade.

```suggestion
      action: PayloadAction<Pick<CrossModelQueryRequest["query"]> & WsControllerURLParam>
```

src/juju/jimm-facade.ts
```diff
@@ -3,10 +3,32 @@ import { autoBind } from "@canonical/jujulib/dist/api/utils";
...
+  errors: Record<string, string[]>;
+};
+
+// In the case an invalid type and/or query is passed, the response might
+// sometimes be a string.
+export type CrossModelQueryFullResponse = CrossModelQueryResponse | string;
+
+export const isCrossModelQueryResponse = (
+  response: CrossModelQueryFullResponse
+): response is CrossModelQueryResponse => typeof response === "object";
```
huwshimi, line 26: Nice work adding this. On little issue is that `typeof null === "object"` as well so this should probably also check that the value is truthy.

```suggestion
): response is CrossModelQueryResponse => response && typeof response === "object";
```

src/juju/jimm-facade.ts
```diff
@@ -16,20 +38,31 @@ class JIMMV2 {
...
+    });
+  }
+
+  crossModelQuery(query: string = "."): Promise<CrossModelQueryFullResponse> {
+    return new Promise((resolve, reject) => {
+      const req = {
+        type: "JIMM",
+        request: "CrossModelQuery",
+        version: 4,
+        params: { type: "jq", query },
```
huwshimi, line 65: It might be worth leaving a comment to say that the API only currently supports the "jq" type so that we remember why this is hardcoded.

src/store/juju/slice.ts
```diff
@@ -108,6 +123,31 @@ const slice = createSlice({
...
+    updateCrossModelQuery: (
+      state,
+      { payload }: PayloadAction<CrossModelQueryFullResponse>
+    ) => {
+      state.crossModelQuery.results = isCrossModelQueryResponse(payload)
+        ? payload.results
+        : null;
+      state.crossModelQuery.errors = isCrossModelQueryResponse(payload)
+        ? payload.errors
+        : payload;
```
huwshimi, line 141: It might be worth leaving a comment here to explain that a string response from the API is an error (similar to the comment you left on the type in jimm-facade.ts).

huwshimi (approved):
I've left some comments for a couple of things I missed in the original review

src/store/juju/slice.ts
vladimir-cucu: I think it should be `Pick<CrossModelQueryRequest, "query">` instead. Also, due to this change, I think we also have to change `const { wsControllerURL, ...params } = action.payload;` to `const { wsControllerURL, query } = action.payload;` in `model-poller.ts`.

src/juju/jimm-facade.ts
huwshimi: Ah yes of course, I think it's fine how it is then.

src/store/juju/slice.ts
huwshimi: Haha yes, that's how you use Pick, not how I wrote it :).

PR #1570, opened 2023-08-17
WD-5811 - feat: Add page component, route and side nav link

vladimir-cucu:
## Done

- Set up the page component with an “Advanced search“ header.
- Set up a “/search“ URL and route. Route is not accessible when using Juju controller.
- Added an “Advanced search“ link to the sidebar that is not shown when using Juju controller.

## Details

https://warthogs.atlassian.net/browse/WD-5811

## Screenshots

Screenshot 2023-08-18 at 00 09 42

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.tsx
+++ b/src/components/PrimaryNav/PrimaryNav.tsx
@@ -1,38 +1,29 @@
 import { dashboardUpdateAvailable } from "@canonical/jujulib/dist/api/versions";
 import { Icon, StatusLabel, Tooltip } from "@canonical/react-components";
-import classNames from "classnames";
 import { useEffect, useMemo, useRef, useState } from "react";
 import { useSelector } from "react-redux";
-import { NavLink } from "react-router-dom";
 
 import Logo from "components/Logo/Logo";
 import UserMenu from "components/UserMenu/UserMenu";
-import { getAppVersion } from "store/general/selectors";
+import { getAppVersion, getIsJuju } from "store/general/selectors";
 import {
   getControllerData,
   getGroupedModelStatusCounts,
 } from "store/juju/selectors";
 import type { Controllers } from "store/juju/types";
 import urls, { externalURLs } from "urls";
+
 import "./_primary-nav.scss";
+import PrimaryNavLink from "./PrimaryNavLink";
 
 const ModelsLink = () => {
   const { blocked: blockedModels } = useSelector(getGroupedModelStatusCounts);
   return (
-    <NavLink
-      className={({ isActive }) =>
-        classNames("p-list__link", {
-          "is-selected": isActive,
-        })
-      }
-      to={urls.models.index}
-    >
-      <i className={`p-icon--models is-light`}></i>
-      <span className="hide-collapsed">Models</span>
+    <PrimaryNavLink to={urls.models.index} iconName="models" title="Models">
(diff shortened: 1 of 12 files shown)


src/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -1,82 +1,79 @@
...
+        controllersCount +
+        controllerArray.reduce(
+          (count, controller) =>
+            "version" in controller && controller.updateAvailable
+              ? count + 1
+              : count,
+          0
+        ),
+      0
+    );
```
vladimir-cucu, line 48: Same functionality, but refactored code. Let me know if previous version is better.

src/components/PrimaryNav/PrimaryNavLink/PrimaryNavLink.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+type PrimaryNavLinkProps = {
+  children?: React.ReactNode;
+  to: string;
+  iconName: string;
+  title: string;
+};
+
+const PrimaryNavLink = ({
```
vladimir-cucu, line 16: This component is identical to the one we used in Audit logs feature. I thought that doing it this way would make it easier for us to merge this branch and the `audit-logs` branch in the future. Let me know if it would be better to remove it.

src/components/Routes/Routes.tsx
```diff
@@ -70,6 +75,16 @@ export function Routes() {
         }
       />
       <Route path="*" element={<PageNotFound />} />
+      {!isJuju && (
```
vladimir-cucu, line 78: As we are adding the Cross Model Search in JIMM, I thought that the user shouldn't be able to access the route when using a Juju controller.

src/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -103,6 +100,11 @@ const PrimaryNav = () => {
         <li className="p-list__item">
           <ControllersLink />
         </li>
+        {!isJuju && (
```
vladimir-cucu, line 103: As we are adding the Cross Model Search in JIMM, I thought that the user shouldn't be able to access this link on the sidebar when using a Juju controller. Let me know if this isn't the case.

huwshimi: Good catch!

huwshimi (approved):
This looks good, good catch that this should only appear for JIMM.

Do you think you could do a second PR that adds the `getIsJuju` selector and `PrimaryNavLink` component to main so that we don't have to have these same improvements in this branch and the audit-logs branch?

src/components/PrimaryNav/PrimaryNavLink/PrimaryNavLink.tsx
huwshimi: That makes sense. It would be great to make this change in main as well to make it easier to merge them both in the future.

huwshimi:
This all looks good, so I'm going to merge it so I can start on the next part.

huwshimi:
Oops! This merged into main, so I've created another PR to revert it: https://github.com/canonical/juju-dashboard/pull/1571 and another to merge these changes to the cross-model-queries branch: https://github.com/canonical/juju-dashboard/pull/1572.

PR #1575, opened 2023-08-21
WD-5812 - test: Add CrossModelQuery test factories

vladimir-cucu:
## Done

- Added CrossModelQuery test factories.
- *Note: I am not sure where `CrossModelQuery.types.ts` should be located and if it should be named that way.*

## Details

https://warthogs.atlassian.net/browse/WD-5812


src/testing/factories/juju/CrossModelQuery.types.ts
```diff
@@ -0,0 +1,108 @@
+export type CrossModelQuery = {
+  "application-endpoints"?: Record<string, ApplicationEndpoint>;
+  applications?: Record<string, Application>;
+  controller?: {
+    imestamp?: string;
```
huwshimi, line 5: Missing a "t"?
```suggestion
    timestamp?: string;
```

src/testing/factories/juju/CrossModelQuery.types.ts
```diff
@@ -0,0 +1,108 @@
...
+    [key: string]: unknown;
+  };
+  machines?: Record<string, Machine>;
+  model?: Model;
+  offers?: Record<string, Offer>;
+  storage?: Record<string, unknown>;
+  [key: string]: unknown;
```
huwshimi, line 12: I think this was a good idea.

src/testing/factories/juju/CrossModelQuery.types.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+type ApplicationEndpoint = {
```
huwshimi, line 15: Most of the types in this file are similar to types we have elsewhere and have the same name. I wonder if it would be helpful to have these all prefixed with `CrossModelQuery...`? It would make them quite long, but would make it clear that these types are only to be used with CrossModelQuery responses.

huwshimi (approved):
This is a great start. I think we should do a followup to add some extensions to the base query, like we do for the general config: https://github.com/canonical/juju-dashboard/blob/main/src/testing/factories/general.ts.

Because there will be a lot of permutations of the query response it would be nice to be able to do something like: crossModelQueryFactory.withRelations().withOffers().withMachines().build()

PR #1576, opened 2023-08-21
WD-5813 - Cross model search form

huwshimi:
## Done

- Add cross model search form

## QA

- Connect to staging.
- Go to the advanced search page.
- Type something into the search input and submit.
- You should see the query in the URL.
- Open the redux dev tools and you should see the dispatch to fetch the queries, and a response from the API.

## Details

https://warthogs.atlassian.net/browse/WD-5813

## Screenshots

Screenshot 2023-08-21 at 1 04 45 pm

### DIFF
--- a/src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
+++ b/src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
@@ -25,18 +25,23 @@ const SearchForm = (): JSX.Element => {
   const [queryParams, setQueryParams] = useQueryParams<{ q: string }>({
     q: "",
   });
-  const query = decodeURIComponent(queryParams.q);
+  const jqParam = decodeURIComponent(queryParams.q);
 
   useEffect(() => {
-    if (query && hasControllerConnection && wsControllerURL) {
-      dispatch(jujuActions.fetchCrossModelQuery({ query, wsControllerURL }));
+    if (jqParam && hasControllerConnection && wsControllerURL) {
+      dispatch(
+        jujuActions.fetchCrossModelQuery({
+          query: jqParam,
+          wsControllerURL,
+        })
+      );
     }
-  }, [dispatch, hasControllerConnection, query, wsControllerURL]);
+  }, [dispatch, hasControllerConnection, jqParam, wsControllerURL]);
 
   return (
     <Formik<Fields>
       initialValues={{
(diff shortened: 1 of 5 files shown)


src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <Formik<Fields>
+      initialValues={{
+        query,
+      }}
+      innerRef={formikRef}
+      onSubmit={(values) => {
+        const { query } = values;
+        if (query) {
+          setQueryParams({ q: encodeURIComponent(query) });
```
vladimir-cucu, line 45: This is just a general thought that crossed my mind now. If we set the query params like this throughout our application, aren't we open to cross-site scripting attacks? Or is this not really a concern for us? I think that this issue can be easily remedied by using something like `DOMPurify.sanitize()` in `useQueryParams`.

vladimir-cucu (approved):
Looks good to me! I've added a few optional comments. 🙂

src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
huwshimi: I don't think this has been a big issue until now, as I think in almost every other case the values have been internal (e.g. the panel) or from the API (e.g. model uuid) instead of user provided (except for the app search in model details).

I think you're right, we should sanitise the values in `useQueryParams`. I'll add a task to do that.

src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const formikRef = useRef<FormikProps<Fields>>(null);
+  const dispatch = useAppDispatch();
+  const wsControllerURL = useAppSelector(getWSControllerURL);
+  const hasControllerConnection = useAppSelector((state) =>
+    getControllerConnection(state, wsControllerURL)
+  );
+  const [queryParams, setQueryParams] = useQueryParams<{ q: string }>({
+    q: "",
+  });
+  const query = decodeURIComponent(queryParams.q);
```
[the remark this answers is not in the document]
huwshimi: Good idea.

PR #1581, opened 2023-08-22
WD-5815 - Add search history

huwshimi:
## Done

- Add a history of queries to the cross model search form.

## QA

- Connect to staging.
- Go to the advanced search page.
- See that there is a button to open the search history.
- Perform a search.
- Open the menu and you should see the search item.
- Do a second search and open the menu again and click on the first search you did. The URL and input should update.
- Refresh the page and open the menu again and your items should be there again.
- Click the clear button and they should get removed.

## Details

https://warthogs.atlassian.net/browse/WD-5815

## Screenshots

Screenshot 2023-08-22 at 2 59 52 pm

### DIFF
--- a/src/pages/AdvancedSearch/SearchForm/SearchHistoryMenu/SearchHistoryMenu.tsx
+++ b/src/pages/AdvancedSearch/SearchForm/SearchHistoryMenu/SearchHistoryMenu.tsx
@@ -2,6 +2,8 @@ import { ContextualMenu, Icon } from "@canonical/react-components";
 import { useFormikContext } from "formik";
 
 import { useQueryParams } from "hooks/useQueryParams";
+import { getCrossModelQueryLoading } from "store/juju/selectors";
+import { useAppSelector } from "store/store";
 
 import type { FormFields } from "../types";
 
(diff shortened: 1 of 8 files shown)


src/pages/AdvancedSearch/SearchForm/SearchHistoryMenu/SearchHistoryMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <>
+              <Icon name="delete" /> {Label.CLEAR}
+            </>
+          ),
+          onClick: () => setQueryHistory([]),
+        },
+      ]}
+      position="left"
+      toggleClassName="has-icon"
+      toggleDisabled={queryHistory.length === 0}
```
vladimir-cucu, line 48: I was wondering if maybe we could also disable it when `isCrossModelQueryLoading = true`. Could we have 2 requests being done almost instantly (by pressing 2 values from history one after another) and then having the first `updateCrossModelQuery` finish after the second one, thus getting the wrong result?

vladimir-cucu (approved):
Looks good to me! I've left a small comment regarding an issue that I'm not sure if we might face.

src/pages/AdvancedSearch/SearchForm/SearchHistoryMenu/SearchHistoryMenu.tsx
huwshimi: Yep, good thinking. I've added that.

PR #1582, opened 2023-08-22
WD-5819 - feat: Add copy JSON button

vladimir-cucu:
## Done

- Added copy JSON button

## Details

https://warthogs.atlassian.net/browse/WD-5819

## Screenshots

Screenshot 2023-08-22 at 17 33 04

### DIFF
--- a/src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
+++ b/src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
@@ -3,13 +3,24 @@ import type { FormikProps } from "formik";
 import { Field, Form, Formik } from "formik";
 import { useEffect, useRef } from "react";
 
+import { copyToClipboard } from "components/utils";
 import { useQueryParams } from "hooks/useQueryParams";
 import {
   getControllerConnection,
   getWSControllerURL,
 } from "store/general/selectors";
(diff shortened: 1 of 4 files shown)


src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
```diff
@@ -67,7 +85,14 @@ const SearchForm = (): JSX.Element => {
               name="query"
               rows={8}
             />
-            <Button type="submit">Search</Button>
+            <Button type="submit">{Label.SEARCH}</Button>
+            <Button
+              type="button"
+              onClick={() => copyToClipboard(crossModelQueryResultsJSON)}
+              disabled={!isCrossModelQueryLoaded || isCrossModelQueryLoading}
```
huwshimi, line 92: I think we should also disable this if the result is an error.

huwshimi (approved):
We don't display the errors yet (https://warthogs.atlassian.net/browse/WD-5869), but I think the button should be disabled if there is an error.

src/pages/AdvancedSearch/SearchForm/SearchForm.tsx
vladimir-cucu: Makes sense! Added in the latest commit.

PR #1587, opened 2023-08-23
WD-5868 - feat: Sanitise query params

vladimir-cucu:
## Done

- Sanitised query params inside `useQueryParams`.

## Details

https://warthogs.atlassian.net/browse/WD-5868

### DIFF
--- a/src/hooks/useQueryParams.ts
+++ b/src/hooks/useQueryParams.ts
@@ -37,14 +37,14 @@ export const useQueryParams = <P extends QueryParams>(
         // Sanitize all query params keys and values to prevent XSS attacks.
         searchParams.forEach((value, key) => {
           const sanitizedKey = DOMPurify.sanitize(key);
-          if (key !== DOMPurify.sanitize(key)) {
+          if (key !== sanitizedKey) {
             searchParams.delete(key);
             console.log(
-              `Key ${key} has been changed to ${sanitizedKey} in order to ` +
-                "prevent potential XSS Attacks."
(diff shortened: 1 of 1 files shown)


src/hooks/useQueryParams.ts
```diff
@@ -33,6 +34,13 @@ export const useQueryParams = <P extends QueryParams>(
             );
           }
         });
+        // Sanitize all query params keys and values to prevent XSS attacks.
+        searchParams.forEach((value, key) => {
```
vladimir-cucu, line 38: I thought that sanitising all query params keys and values would be a good extra protection for the instances where we have query params keys/values that require to be sanitised, but we don't change them directly with `useQueryParams`.

vladimir-cucu: Also, on lines 56 and 58, I don't think it makes sense to sanitise those values. If we do so, we would have a discrepancy between the results we get from `useQueryParams` and the actual ones, as those sanitised results won't be applied to the actual query params. Instead, I think it would make sense to sanitise when we are calling `setParam`.

huwshimi: I think we need to both, otherwise wouldn't this be vulnerable to someone modifying the query string before loading the page? I think it would be OK to have a discrepancy in the case that someone has entered something manually in the URL.

vladimir-cucu: Makes sense! Changed it in my latest commit.

huwshimi (approved):
This looks good. Thanks.

PR #1598, opened 2023-08-29
WD-6052 - Link charms to Charmhub

huwshimi:
## Done

- Link to Charmhub from charms.

## QA

- Go to the advanced search page and do a search.
- In the tree open an applications object and then open an app.
- The "charm" value should link to charmhub.io.

## Details

https://warthogs.atlassian.net/browse/WD-6052

## Screenshots

Screenshot 2023-08-29 at 1 44 43 pm

### DIFF
--- a/src/pages/AdvancedSearch/CodeSnippetBlock/CodeSnippetBlock.test.tsx
+++ b/src/pages/AdvancedSearch/CodeSnippetBlock/CodeSnippetBlock.test.tsx
@@ -4,11 +4,26 @@ import userEvent from "@testing-library/user-event";
 import type { RootState } from "store/store";
 import { jujuStateFactory, rootStateFactory } from "testing/factories";
 import { generalStateFactory, configFactory } from "testing/factories/general";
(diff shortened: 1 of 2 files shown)


src/pages/AdvancedSearch/ResultsBlock/ResultsBlock.tsx
```diff
@@ -70,6 +71,23 @@ const getTab = (key: string) => {
   }
 };
 
+const getCurrentObject = (
```
vladimir-cucu, line 74: It took me a bit of time to understand what this does, so maybe it might be a good idea to add a small comment here?

vladimir-cucu (approved):
Looks good to me! Added just a small optional sugestion.

src/pages/AdvancedSearch/ResultsBlock/ResultsBlock.tsx
huwshimi: Yes! Sorry, I definitely should have left some comments here. I've added a bunch to this function, so hopefully it makes more sense now.

PR #1600, opened 2023-08-30
WD-5822 - feat: Map unit-key tree data to links

vladimir-cucu:
## Done

- Added the remaining links to app, machine and unit as described in the bellow Jira ticket.

## Details

https://warthogs.atlassian.net/browse/WD-5822


src/components/hooks.ts
```diff
@@ -14,3 +22,16 @@ export const useEntityDetailsParams = () => {
     userName,
   };
 };
+
+export const useModelByUUIDDetails = ({
```
vladimir-cucu, line 26: I thought it might be a good idea to create this custom hook, as we would need to get the `modelName` and `userName` in quite a few places where we want to map tree data to links. I still have to modify `ModelDetailsLink` to use it, as I didn't know if creating this custom hook would be a good approach to begin with.

src/pages/AdvancedSearch/CodeSnippetBlock/CodeSnippetBlock.tsx
```diff
@@ -139,6 +150,8 @@ const CodeSnippetBlock = ({ className, title, code }: Props): JSX.Element => {
     (valueAsString, value, ...keyPath) => {
       const currentKey = keyPath[0];
       const parentKey = keyPath.length >= 2 ? keyPath[1] : null;
+      const grandparentKey = keyPath.length >= 3 ? keyPath[2] : null;
```
vladimir-cucu, line 153: I definitely think there should be a better name for this, but couldn't find one. Should we just remove this line and input the value directly in line 209?

src/pages/AdvancedSearch/UnitsLink/UnitsLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useModelByUUIDDetails } from "components/hooks";
+import urls from "urls";
+
+type Props = {
+  uuid: string;
+  appName: string | null;
+  unitId: string | null;
+} & PropsWithChildren;
+
+const UnitsLink = ({ uuid, appName, unitId, children }: Props): JSX.Element => {
```
vladimir-cucu, line 13: Both `UnitsLink` and `UnitsMachineLink` look similar, however, there are a few subtle differences between both of them and thus, I didn't know if we should combine both of them in a single component.

src/components/ModelDetailsLink/ModelDetailsLink.tsx
```diff
@@ -40,9 +42,11 @@ const ModelDetailsLink = ({
...
   ...props
 }: Props) => {
-  const modelDetails = useAppSelector((state) => getModelByUUID(state, uuid));
-  const owner = uuid ? modelDetails?.ownerTag : ownerTag;
-  const model = uuid ? modelDetails?.name : modelName;
+  const { modelName: model, userName: owner } = useModelByUUIDDetails({
+    uuid,
+    ownerTag,
+    modelName,
+  } as ModelByUUIDDetailsProps);
```
huwshimi, line 49: It could be a little clearer to pass `uuid
      ? { uuid }
      : ({
          ownerTag,
          modelName,
        } as NameProps)`

With possibly also a comment for why the `as NameProps` is necessary (maybe something like "This component's props require either `uuid` or `modelName` to be exist, but this gets lost when the props are destructured."

huwshimi (commented):
This is looking good!

src/components/AppLink/AppLink.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+  it("should render correctly when there is no view", () => {
+    renderComponent(
+      <AppLink uuid="abc123" appName="mockApp">
+        abc123
+      </AppLink>,
+      { state }
+    );
+    expect(screen.queryByRole("link", { name: "abc123" })).toHaveAttribute(
+      "href",
+      "/models/eggman@external/test-model/app/mockApp"
```
huwshimi, line 37: I think it would be nice if all these tests used the URL definitions, so that if the URLs ever change the tests will continue to work.

```suggestion
      urls.model.app.index({ userName: "eggman@external", modelName: "test-model", appName: "mockApp" })
```

huwshimi (approved):
Nicely done. One small suggestion about making the tests a bit easier to maintain.

src/components/AppLink/AppLink.test.tsx
vladimir-cucu: Thanks for the review! Changed this in my latest commit.

PR #1604, opened 2023-09-01
Versioned JIMM facades

huwshimi:
## Done

- Set up versioned JIMM facades.
- Add facade tests.

## QA

- Connect to production JIMM.
- Check that you can load the dashboard and your models appear.

### DIFF
--- a/src/juju/jimm/index.ts
+++ b/src/juju/jimm/index.ts
@@ -1,0 +1,10 @@
+import JIMMV3 from "./JIMMV3";
+
+export * as JIMMV3 from "./JIMMV3";
+
+const JIMM = {
+  name: "JIMM",
+  versions: [JIMMV3],
+};
+
+export default JIMM;
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -13,8 +13,8 @@ import type ModelManagerV9 from "@canonical/jujulib/dist/api/facades/model-manag
 import type { ModelSLAInfo } from "@canonical/jujulib/dist/api/facades/model-manager/ModelManagerV9";
 import type PingerV1 from "@canonical/jujulib/dist/api/facades/pinger/PingerV1";
 
-import type JIMMV2 from "juju/jimm-facade";
-// See https://github.com/juju/juju/blob/develop/apiserver/params/multiwatcher.go
+import type JIMMV3 from "./jimm/JIMMV3";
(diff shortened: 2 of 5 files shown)


src/juju/jimm/index.ts
```diff
@@ -0,0 +1,10 @@
+import JIMMV3 from "./JIMMV3";
+
+export * as JIMMV3 from "./JIMMV3";
```
vladimir-cucu, line 3: I am not sure if I understood the intention here correctly, but if we want to export the imported `JIMMV3`, shouldn't we have `export { default as JIMMV3 } from "./JIMMV3";` or `export { JIMMV3 };` here instead?

src/juju/types.ts
```diff
@@ -13,8 +13,8 @@ import type ModelManagerV9 from "@canonical/jujulib/dist/api/facades/model-manag
 import type { ModelSLAInfo } from "@canonical/jujulib/dist/api/facades/model-manager/ModelManagerV9";
 import type PingerV1 from "@canonical/jujulib/dist/api/facades/pinger/PingerV1";
 
-import type JIMMV2 from "juju/jimm-facade";
-// See https://github.com/juju/juju/blob/develop/apiserver/params/multiwatcher.go
+import type JIMMV3 from "./jimm/JIMMV3";
```
vladimir-cucu, line 16: With the above proposed change, I think we could replace this line with `import { type JIMMV3 } from "./jimm";`.

vladimir-cucu (approved):
Looks good to me! I proposed a potential change, however, I am unsure if I understood correctly what should be exported in `index.ts`.

src/juju/jimm/index.ts
huwshimi: I think normally that would be a better idea (or probably forgo the re-export entirely), but I'm inclined to keep it like this so that it's consistent with the other facades from jujulib, so they can all be consumed in the same way e.g. https://github.com/juju/js-libjuju/blob/master/api/facades/client/index.ts.

For the JIMM v3 facade there's only a single export that gets captured by this wild card export, but for JIMM v4 there are more: https://github.com/canonical/juju-dashboard/blob/63230bd08bc31c61ef74fa67ad10dccd7ddb2c16/src/juju/jimm/JIMMV4.ts.

PR #1611, opened 2023-09-04
Create JIMM v4 class

huwshimi:
## Done

- Add JIMM v4 class.

## QA

- Go to the audit logs page.
- Check that the audit logs load.

### DIFF
--- a/src/juju/jimm/JIMMV4.ts
+++ b/src/juju/jimm/JIMMV4.ts
@@ -1,0 +1,73 @@
+import type { ConnectionInfo, Transport } from "@canonical/jujulib";
+import { autoBind } from "@canonical/jujulib/dist/api/utils";
+
+import JIMMV3 from "./JIMMV3";
+
+// As typed in JIMM:
+// https://github.com/canonical/jimm/blob/1da76ed2d8dba741f5880f32c073f85f7d518904/api/params/params.go#L92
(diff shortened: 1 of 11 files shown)


src/juju/jimm/JIMMV4.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Automatically bind all methods to instances.
+    autoBind(this);
+  }
+
+  findAuditEvents(params: FindAuditEventsRequest = {}): Promise<AuditEvents> {
+    return new Promise((resolve, reject) => {
+      const req = {
+        type: "JIMM",
+        request: "FindAuditEvents",
+        version: 3,
```
vladimir-cucu, line 63: This might be a miss-understanding from my part, but considering that it uses version 3, shouldn't it be placed in `JIMMV3` instead?

vladimir-cucu (approved):
Looks good to me and `AuditLogs` works as expected!

src/juju/jimm/JIMMV4.ts
huwshimi: I think you're right. I don't think the feature is enabled with the JAAS deployment (it returns an error), but the method exists in v3 and using the v3 facade with staging works just fine, so this PR is not necessary.

PR #1612, opened 2023-09-04
WD-6219 - Enable cross model search with JIMM v4

huwshimi:
## Done

- Use the JIMM facade version to enable cross model search.

## QA

- Connect your dashboard to staging.
- Load the dashboard and check that the Advanced search link is in the sidenav.
- Connect your dashboard to JAAS.
- Check that the Advanced search link is NOT in the sidenav.

## Details

https://warthogs.atlassian.net/browse/WD-6219

### DIFF
--- a/src/components/Routes/Routes.tsx
+++ b/src/components/Routes/Routes.tsx
@@ -1,5 +1,4 @@
 import { useEffect } from "react";
-import { useSelector } from "react-redux";
 import {
   Navigate,
   Route,
@@ -15,11 +14,7 @@ import ModelDetails from "pages/ModelDetails/ModelDetails";
 import ModelsIndex from "pages/ModelsIndex/ModelsIndex";
 import PageNotFound from "pages/PageNotFound/PageNotFound";
 import Settings from "pages/Settings/Settings";
-import {
-  getIsJuju,
-  getControllerFeatureEnabled,
-  getWSControllerURL,
-} from "store/general/selectors";
+import { isCrossModelQueriesEnabled } from "store/general/selectors";
 import { useAppSelector } from "store/store";
 import urls from "urls";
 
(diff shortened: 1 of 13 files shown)


src/store/middleware/model-poller.ts
```diff
@@ -92,6 +92,16 @@ export const modelPollerMiddleware: Middleware<
...
+          const jimmVersion = conn.facades.jimM?.version ?? 0;
+          reduxStore.dispatch(
+            generalActions.updateControllerFeatures({
+              wsControllerURL,
+              features: {
+                auditLogs: jimmVersion >= 4,
+                crossModelQueries: jimmVersion >= 4,
+              },
+            })
+          );
```
huwshimi, line 104: This is where we flag whether a feature should be enabled.

src/components/Routes/Routes.tsx
```diff
@@ -31,6 +36,10 @@ export function Routes() {
   const location = useLocation();
 
   const isJuju = useSelector(getIsJuju);
+  const wsControllerURL = useSelector(getWSControllerURL);
+  const crossModelQueriesSupported = useAppSelector((state) =>
+    getControllerFeatureEnabled(state, wsControllerURL, "crossModelQueries")
+  );
```
vladimir-cucu, line 42: Was wondering if maybe it might make sense to make this its own selector, as we are using this exact selector in 2 places.

vladimir-cucu (approved):
Looks good to me! Tested on staging, JAAS and local controller, and all works as expected.

src/components/Routes/Routes.tsx
huwshimi: Good idea. I've added a selector that also retrieves the other data it needs, so this simplifies the components quite a bit.

PR #1613, opened 2023-09-05
WD-5786 - Sort by most recent first

huwshimi:
## Done

- Sort audit logs by most recent first.

## QA

- Go to the logs page.
- Check that the audit logs appear with most recent first.

## Details

https://warthogs.atlassian.net/browse/WD-5786

### DIFF
--- a/src/components/AuditLogsTable/hooks.ts
+++ b/src/components/AuditLogsTable/hooks.ts
@@ -62,6 +62,8 @@ export const useFetchAuditEvents = () => {
         "user-tag": queryParams.user ? `user-${queryParams.user}` : undefined,
         model: model ?? undefined,
         method: queryParams.method ?? undefined,
+        // Sort by most recent first.
+        sortTime: true,
       })
     );
   }, [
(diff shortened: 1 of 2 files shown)


vladimir-cucu (approved):
Looks good to me and works as expected! 

Note: One thing that I've observed is that, due to the fact that a lot of logs are generated, if you stay for a while on a page and move to the next one, you might actually end up retrieving logs that are newer than the ones on the page you were before. I am not sure how intuitive this would be for the end user.

huwshimi:
> Looks good to me and works as expected!
> 
> Note: One thing that I've observed is that, due to the fact that a lot of logs are generated, if you stay for a while on a page and move to the next one, you might actually end up retrieving logs that are newer than the ones on the page you were before. I am not sure how intuitive this would be for the end user.

That's a good point. I think we should see what what this looks like after the login bug that causes many retries is resolved as I think the quantity will go down considerably.

We could potentially set the "before" filter to the current time whenever you visit the first page which would mean you don't get any newer logs. The "before" time could get reset if you click the refresh button or if you navigate back to the first page.

Issue #1529, opened 2023-07-13
Get model info from watcher for Juju 3.2


huwshimi:
Juju 3.2 adds additional model data to the watcher response. This resolves this issue: https://bugs.launchpad.net/juju/+bug/1939341.

When connected to 3.2 the dashboard no longer needs to make the call here: https://github.com/canonical/juju-dashboard/blob/304e00ad1a3cda8a9cd72fbd8b1798c2b6ed7aea/src/pages/ModelDetails/ModelDetails.tsx#L39.

We'll also need to update the (or add a) selector to fetch that data from the correct spot depending on which version of Juju is being used.

PR #1617, opened 2023-09-07
WD-6200 - reduce full status calls for Juju 3.2

huwshimi:
## Done

 - Use data from the watcher instead of getting the full status when using Juju 3.2 and later. This reduces the number of full status calls we need to make, which is expensive for the server.

## QA

- Connect to a local Juju 3.2 controller.
- Go to the model list and click on a model.
- On the left side there should be data for controller, cloud/region and version.
- Open the redux dev tools and filter the dispatched actions by `populateMissingAllWatcherData`. There should be no entries.
- Connect to a pre Juju 3.2 controller (e.g. JAAS which is 2.9).
- Go to the model list and click on a model.
- On the left side there should be data for controller, cloud/region and version.
- Open the redux dev tools and filter the dispatched actions by `populateMissingAllWatcherData`. There should be at least one entry.

## Details

https://warthogs.atlassian.net/browse/WD-6200
Fixes: https://github.com/canonical/juju-dashboard/issues/1529.

### DIFF
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -106,7 +106,7 @@ export interface WatcherModelData {
 // This type is used for Juju versions before 3.2.
 export interface Pre32AnnotatedWatcherModelInfo extends Pre32ModelChangeDelta {
   cloud: string;
-  region?: string;
+  "cloud-region"?: string;
   type: string;
   version: string;
 }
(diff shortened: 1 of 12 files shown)


vladimir-cucu (approved):

vladimir-cucu (approved):
Looks good to me! Tested on local controller with version 3.2.3 and with JAAS, and get the expected result.

Issue #1532, opened 2023-07-17
Use upstream ConfirmationModal


huwshimi:
We should replace the custom ConfirmationModal with the upstream one: https://canonical.github.io/react-components/?path=/docs/confirmationmodal--default-story.

We need to display the confirmation when clicking buttons and submitting forms so we might not be able to use the upstream ConfirmationButton component.

PR #1618, opened 2023-09-07
WD-6216 - refactor: Use ConfirmationModal from react-components

vladimir-cucu:
## Done

- Replaced custom made `ConfirmationModal` with the respective component from @canonical/react-components.

## Details

- Jira: https://warthogs.atlassian.net/browse/WD-6216
- Fixes: #1532


src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -215,29 +219,56 @@ export default function ConfigPanel(): JSX.Element {
...
-            setConfirmType(null);
-            // Clear the form errors if there were any from a previous submit.
-            setFormErrors(null);
-            _submitToJuju();
-          },
-          () => setConfirmType(null)
+        return (
+          // Render the submit confirmation modal.
+          <ConfirmationModal
+            className="p-confirmation-modal"
```
vladimir-cucu, line 226: I left the `className` to be `p-confirmation-modal` here, as I thought that by doing so, the `className` would be consistent with the other `ConfirmationModel`s used in `ActionsPanel` and `CharmActionsPanel`. Please let me know if adding an extra class here or removing the current class and adding another one would make more sense.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -72,6 +71,8 @@ type ActionsQueryParams = {
   units: string[];
 };
 
+type ConfirmTypes = "submit" | null;
```
vladimir-cucu, line 74: I was thinking that maybe we could create a `types.ts` file in `panels` folder and move this one there, as this exact type is utilized in `CharmActionsPanel`. Also, we utilize `type ConfirmTypes = "apply" | "cancel" | null` in `ConfigPanel.tsx` and I was thinking if maybe it might make sense to make a common one for all of them that looks something like `type ConfirmTypes = "submit" | "cancel" | null`?

huwshimi: I think that would be fine to do.

huwshimi (commented):
This is looking good. I've made a few suggestions and it also appears that these confirmations are being rendered inside the panel which makes the background only cover the panel:

Screenshot 2023-09-08 at 10 14 32 am

I think these probably need to be updated to use the `usePortal` hook to put them at the top of the DOM.

src/panels/ConfigPanel/ConfigPanel.tsx
vladimir-cucu: Just read the suggestions and I agree with them. Thank you for the review!

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -215,29 +220,62 @@ export default function ConfigPanel(): JSX.Element {
...
+              <p>
+                You have edited the following values to the {appName}{" "}
+                configuration:
+              </p>
+              {changedConfigList}
+              <div className="config-panel__modal-button-row-hint">
+                You can revert back to the applications default settings by
+                clicking the “Reset all values” button; or reset each edited
+                field by clicking “Use default”.
+              </div>
```
huwshimi, line 252: To make this look a little like the old confirmation, you could pass this as a prop to `<ConfirmationModal>`:

```

              confirmExtra={
                <p className="u-text--muted u-align--left">
                  You can revert back to the applications default settings by
                  clicking the “Reset all values” button; or reset each edited
                  field by clicking “Use default”.
                </p>
              }
```

The `&__modal-button-row-hint` class can then be removed from `_config-panel.scss`.

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -215,29 +220,62 @@ export default function ConfigPanel(): JSX.Element {
...
-            setConfirmType(null);
-            // Clear the form errors if there were any from a previous submit.
-            setFormErrors(null);
-            _submitToJuju();
-          },
-          () => setConfirmType(null)
+      if (confirmType === "submit") {
+        // Render the submit confirmation modal.
+        return (
+          <Portal>
```
huwshimi, line 226: This same change also needs to happen in `ActionsPanel`.

huwshimi (approved):
One thing I noticed on the config panel is that the modal appears below the panel, but it seems OK for the actions panel.

Screenshot 2023-09-11 at 11 24 20 am

src/panels/ConfigPanel/ConfigPanel.tsx
vladimir-cucu: Thanks for pointing this out! I think it needs to be added to `CharmActionsPanel` as well.

src/scss/_breakpoints.scss
```diff
@@ -2,7 +2,7 @@
 
 // 0px -> mobile
 @mixin mobile {
-  @media screen and (max-width: $breakpoint-small) {
+  @media screen and (max-width: $breakpoint-small - 1px) {
```
vladimir-cucu, line 5: Forgot to add this change in the previous PR that fixed sidebar scroll when `BaseLayout` height exceeds viewport height.

src/components/PortalConfirmationModal/PortalConfirmationModal.tsx
```diff
@@ -0,0 +1,17 @@
+import { ConfirmationModal } from "@canonical/react-components";
+import type { ConfirmationModalProps } from "@canonical/react-components/dist/components/ConfirmationModal";
+import usePortal from "react-useportal";
+
+import "./_portal-confirmation-modal.scss";
+
+const PortalConfirmationModal = (props: ConfirmationModalProps) => {
```
vladimir-cucu, line 7: We use this in 4 different places and it might get used in other places in the future, so I thought it might make sense to make it its own component.

vladimir-cucu: Also, we change the `z-index` of `.p-modal`, thus, by making this its own component, the change to `.p-modal` would not influence other elements that might have this class.

vladimir-cucu: Thank you for this explanation!

Issue #1616, opened 2023-09-07
Sidebar scrolls in model list


huwshimi:
**Describe the bug**
The sidebar scrolls when there are a lot of models

**To Reproduce**
1. Add a lot of models.
2. Go to the model list.
3. Scroll the list of the models.
4. See that the sidebar scrolls.

**Expected behavior**
The sidebar should be sticky.

**Screenshots**
<img width="841" alt="Screenshot 2023-09-07 at 11 03 58 am" src="https://github.com/canonical/juju-dashboard/assets/361637/3bf91ef1-d9b2-476c-82d1-a5384801ec65">

**Additional context**
This has been fixed in the audit logs and cross model query branches, but not in main.

https://github.com/canonical/juju-dashboard/commit/a7f42ab45b389d5a0c0f779007519f99973b0f1c#diff-fd8132c704f92ac7d42cd11b60ec0393b0d08e6655e2e72c8d24c99d21909572

PR #1622, opened 2023-09-10
fix: Fix sidebar scroll when BaseLayout height exceeds viewport height

vladimir-cucu:
## Done

- Fixed sidebar scroll when BaseLayout height exceeds viewport height.

## Details

- Fix was previously implemented in a7f42ab for Audit Logs feature.
- Fixes: #1616 

## Screenshots

Without fix:
Screenshot from 2023-09-11 00-23-27

With fix:
Screenshot from 2023-09-11 00-23-17


huwshimi (approved):
Thanks!

Issue #1620, opened 2023-09-07
Don't show configure for read-only models


huwshimi:
When viewing an application in a model that you don't have write or admin permissions for, it should not show the Configure button.

PR #1629, opened 2023-09-12
WD-6258 - Hide controls on read only models

huwshimi:
## Done

- Don't display configure and action controls if the user has read only permissions.

## QA

- Using a local controller add a new user with `juju add-user [username]`.
- Grant them read only permissions for a model with at least one app: `juju grant [username] read [model-name]`.
- In the dashboard log in as the new user.
- Go to the model with read-only permissions.
- Search for an app. There shouldn't be any checkboxes or a "Run action" button.
- Click on an app. There shouldn't be a "Configure" button on the left or a "Run action" button and there should be no checkboxes.

## Details

https://warthogs.atlassian.net/browse/WD-6258
Fixes: https://github.com/canonical/juju-dashboard/issues/1620

### DIFF
--- a/src/components/ModelTableList/CloudGroup.tsx
+++ b/src/components/ModelTableList/CloudGroup.tsx
@@ -12,7 +12,7 @@ import {
 } from "store/juju/selectors";
 import type { Filters } from "store/juju/utils/models";
 import {
-  canAdministerModelAccess,
+  canAdministerModel,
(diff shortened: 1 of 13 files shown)


vladimir-cucu (approved):
Looks good to me! Tested it with a local controller and a user that has only read access to my model. Works as expected!

Issue #1534, opened 2023-07-17
Display names of apps when displaying charms in actions panel


huwshimi:
In the applications tab when displaying the list of charms to choose from to perform an action, it would be helpful to list the names of the applications with each radio button option so you can tell which apps the action will be performed on.

<img width="785" alt="Screenshot 2023-07-17 at 5 12 22 pm" src="https://github.com/canonical/juju-dashboard/assets/361637/47be9842-2658-4519-a6ec-01b372dbd8a7">

PR #1631, opened 2023-09-12
WD-6271 - feat: Display names of apps when displaying charms in actions panel

vladimir-cucu:
## Done

- Display names of apps when displaying charms in actions panel

## Details

- Jira: https://warthogs.atlassian.net/browse/WD-6271
- Fixes: #1534 

## Screenshots

Screenshot from 2023-09-13 12-54-53


src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const CharmApplicationsDetails = ({ charmURL }: Props): JSX.Element => {
+  const selectedApplications = useAppSelector(
+    getSelectedApplications(charmURL)
+  );
+
+  return (
+    <p className="p-form-help-text is-tick-element">
+      Apps:{" "}
+      {selectedApplications
```
huwshimi, line 16: What do you think about truncating this list if there are more than say five apps? I think I suggested a tooltip, but it could also just have a "+3 more" which when clicked could append them to the existing list, and then a "show less" link could appear which would return the list to the previous state.

huwshimi (approved):

huwshimi:
It looks like this might need a max-width and/or setting the text to wrap so that it handles long lists:

Screenshot 2023-09-14 at 9 23 11 am

src/scss/_utils.scss
```diff
@@ -51,8 +51,17 @@
...
 .p-tooltip--constrain-width .p-tooltip__message {
-  max-width: 20rem;
+  @media (width <= 25rem) {
+    width: 80vw;
+  }
+
+  @media (width > 25rem) {
+    width: 20rem;
+  }
+
```
vladimir-cucu, line 64: Tested it for a bit and couldn't realize why the tooltip always seemed to have 0 width. This fix works for the tooltips with 0 width as well.

src/scss/_utils.scss
```diff
@@ -51,8 +51,17 @@
...
   white-space: pre-wrap;
```
huwshimi, line 65: I'm not sure why you were getting tooltips with 0 width (maybe we can talk about this later), but I think we can improve this slightly by using the predefined breakpoints. In this case we probably want to change the tooltip at the same breakpoint that the panel goes full width.

I also think we probably want to handle this differently to the `p-tooltip--constrain-width` tooltip so that those tooltips don't get a fixed width.

The issue with using a fixed with is that the tooltips then don't reduce the width to smaller content. In MAAS I think we made [this util](https://github.com/canonical/maas-ui/blob/main/src/app/utils/breakLines.ts) to set a max width on the lines before they were inserted into the tooltip.

Here's how we could implement this using the existing breakpoints:

```suggestion
.p-tooltip--constrain-width .p-tooltip__message {
  max-width: 20rem;
  white-space: pre-wrap;

  @media (max-width: $breakpoint-x-small) {
    max-width: 80vw;
  }
}

.p-tooltip--fixed-width .p-tooltip__message {
  white-space: pre-wrap;
  width: 20rem;

  @media (max-width: $breakpoint-x-small) {
    width: 80vw;
  }
}
```

src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+const CharmApplicationsDetails = ({ charmURL }: Props): JSX.Element => {
+  const selectedApplications = useAppSelector(
+    getSelectedApplications(charmURL)
+  );
+  const displayedApplications = selectedApplications.slice(0, 5);
+  const hiddenApplications = selectedApplications.slice(5);
+
+  return (
+    <p className="p-form-help-text is-tick-element">
+      <Tooltip
```
huwshimi, line 26: I think you forgot to add the class: `tooltipClassName="p-tooltip--fixed-width"` (or `tooltipClassName="p-tooltip--constrain-width"` depending on which you want to use).

src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+        message={
+          hiddenApplications.length ? (
+            <span>..., {formatApplicationsDetails(hiddenApplications)}</span>
```
huwshimi, line 29: I wonder if this ellipsis makes it look something is cut off? It's probably the correct grammar, but if we move the tooltip to "+5 more" (see my suggestion below) then it'll be more obvious that the tooltip is only showing the extra items.

src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -0,0 +1,43 @@
...
+          ) : null
+        }
+        position="btm-left"
+      >
+        {formatApplicationsDetails(displayedApplications)}
+        {hiddenApplications.length ? (
+          <span>, +{hiddenApplications.length} more...</span>
```
huwshimi, line 36: I wonder if the tooltip should only appear when hovering `+5 more` and in which case we could wrap it in `<u>` it to signify there is an interaction. I also don't think we need the ellipsis here.

src/scss/_utils.scss
```diff
@@ -51,18 +51,22 @@
...
+    max-width: 80vw;
   }
+}
 
-  white-space: pre-wrap;
+// Specifies horizontal width of tooltips with 0 width on small screens.
+// Combine with utils/breakLines for tooltips with 0 width on bigger width screens.
+.p-tooltip--fixed-x-small-width .p-tooltip__message {
+  @media (max-width: $breakpoint-x-small) {
+    white-space: pre-wrap;
```
huwshimi, line 67: Because this has a fixed width we can just use `wrap` here to avoid having to conditionally breaking the lines in the component.

```suggestion
    white-space: wrap;
```

src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -23,19 +27,23 @@ const CharmApplicationsDetails = ({ charmURL }: Props): JSX.Element => {
...
+      {hiddenApplications.length ? (
+        <>
+          {" + "}
+          <Tooltip
+            tooltipClassName="p-tooltip--fixed-x-small-width"
+            message={
+              isSmallScreen
+                ? formatApplicationsDetails(hiddenApplications)
+                : breakLines(formatApplicationsDetails(hiddenApplications))
+            }
```
huwshimi, line 40: I've suggested a small change to the CSS that means we don't have to check for the screen size in this component and conditionally break the lines.

```suggestion
            message={breakLines(formatApplicationsDetails(hiddenApplications))}
```

src/panels/CharmsPanel/CharmApplicationsDetails/CharmApplicationsDetails.tsx
```diff
@@ -23,19 +27,23 @@ const CharmApplicationsDetails = ({ charmURL }: Props): JSX.Element => {
...
+      {formatApplicationsDetails(displayedApplications)}
+      {hiddenApplications.length ? (
+        <>
+          {" + "}
+          <Tooltip
+            tooltipClassName="p-tooltip--fixed-x-small-width"
+            message={
+              isSmallScreen
+                ? formatApplicationsDetails(hiddenApplications)
+                : breakLines(formatApplicationsDetails(hiddenApplications))
```
huwshimi, line 39: it might be helpful to leave comment here for why we need to break the lines, maybe something along the lines of: `Insert line breaks into the tooltip so that the content wraps instead of appearing as one long line.`

huwshimi (approved):
I've made a suggestion so that we don't have to monitor the screen size in JS.

PR #1644, opened 2023-09-29
WD-6565 - feat: Handle Promise errors

vladimir-cucu:
## Done

- Set-up `eslint-plugin-promise` and handle Promise errors in all files.

## TODO

- Improve code coverage by adding additional tests.

## Details

- https://warthogs.atlassian.net/browse/WD-6565

### DIFF
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -201,13 +201,16 @@ export async function connectAndLoginWithTimeout(
     CLIENT_VERSION
   );
   return new Promise((resolve, reject) => {
-    Promise.race([timeout, juju]).then((resp) => {
-      if (typeof resp === "string") {
-        reject("timeout");
+    Promise.race([timeout, juju])
+      .then((resp) => {
+        if (typeof resp === "string") {
+          reject(new Error("timeout"));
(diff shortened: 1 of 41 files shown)


src/juju/api.ts
```diff
@@ -457,14 +469,12 @@ export async function fetchControllerList(
...
-      try {
-        await conn.facades.jimM.disableControllerUUIDMasking();
-        resolve();
-      } catch (e) {
-        reject();
-      }
+      conn.facades.jimM
+        .disableControllerUUIDMasking()
+        .then(() => resolve())
+        .catch((e) => reject(new Error()));
```
huwshimi, line 477: I think the error that is thrown here should have a message, maybe something like:

```suggestion
        .catch((e) => reject(new Error("Unable to disabled controller UUID masking.", e)));
```

huwshimi (approved):
Just one small comment. Thanks for all these fixes!

src/juju/api.ts
vladimir-cucu: Good suggestion! Changed it in my latest commit.

PR #1651, opened 2023-10-31
Include ReBAC Admin package

huwshimi:
## Done

- Add the rebac-admin package.
- Add a users page and display the ReBACAdmin component.

## QA

- Click on Users in the sidebar.
- It should show the ReBAC Admin component.

## Details

- https://warthogs.atlassian.net/browse/WD-6757

## Screenshots

Screenshot 2023-10-31 at 9 23 45 am

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -142,7 +142,7 @@ developing Juju Dashboard.
 
 ### React
 
-Juju Dashboard uses [React](https://react.dev/) for it's component based UI.
+Juju Dashboard uses [React](https://react.dev/) for its component based UI.
 
 #### Components
 
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "react-hot-toast": "2.4.1",
     "react-json-tree": "0.18.0",
     "react-redux": "8.1.0",
-    "react-router-dom": "6.12.1",
+    "react-router-dom": "6.18.0",
     "react-scripts": "5.0.1",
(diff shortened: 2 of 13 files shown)


vladimir-cucu (approved):
It works as expected once we do the following changes in `yarn.lock` file:

Change line 1719 to: `resolution: "@canonical/rebac-admin@git+ssh://git@github.com:canonical/rebac-admin.git#commit=e5d00eeb809d4b13d120128b03c086f88440da20"`

Change line 1732 to: `checksum: 4448322b7a0c25a444e11ba6253d8875ba59eaeb35fca52a87bd2117c039b165d3d8286438c1c049d1cfa08930ce6a5c436311c20a215ec26576cf53c7d623d7`

If we don't do so, we will get in `src/components/PrimaryNav` the error: `Module '"@canonical/rebac-admin"' has no exported member 'AuthenticationLink'.` It won't see `AuthenticationLink`, `EntitlementsLink`, `GroupsLink`, `ResourcesLink`, `RolesLink` and `UsersLink` as well. I assume this is due to the fact that `yarn install` tries to build `@canonical/rebac-admin` from an older version of `main` that doesn't have those components.

huwshimi:
@vladimir-cucu ah right, I wrongly assumed it would use the latest commit from the branch. I've updated the yarn.lock. I needed to run `yarn clean` for it to update the package so you might find you need to do the same thing.

Issue #1652, opened 2023-11-07
UI only displays error if there is no controller


vladimir-cucu:
**Describe the bug**
Once the controllers are fetched, we get following screen in case there is no controller available:

![Screenshot from 2023-11-07 12-18-26](https://github.com/canonical/juju-dashboard/assets/108150922/635c68a2-5db5-4932-82a8-43a18459ec21)

**To Reproduce**
Login with a user that has 0 controllers.

**Expected behavior**
The dashboard should load as usual and show that there are 0 controllers.

PR #1653, opened 2023-11-07
WD-7282 - fix: Display dashboard with 0 controllers

vladimir-cucu:
## Done

- Display dashboard as usual instead of error message when there are 0 controllers.

## QA

- Login with a user that has 0 controllers.
- Check that the dashboard loads as usual instead of a `Notification` component with an error message being displayed on the whole page.

## Details

- https://warthogs.atlassian.net/browse/WD-7282
- Fixes: https://github.com/canonical/juju-dashboard/issues/1652

### DIFF
--- a/src/juju/jimm/JIMMV3.ts
+++ b/src/juju/jimm/JIMMV3.ts
@@ -93,7 +93,7 @@ class JIMMV3 {
     });
   }
 
-  listControllers(): Promise<{ controllers: ControllerInfo[] }> {
+  listControllers(): Promise<{ controllers: ControllerInfo[] | null }> {
     return new Promise((resolve, reject) => {
       const req = {
         type: "JIMM",
--- a/src/store/juju/selectors.ts
+++ b/src/store/juju/selectors.ts
@@ -187,8 +187,8 @@ export const getControllerData = createSelector(
 export const getControllersCount = createSelector([slice], (sliceState) => {
   const controllerData = sliceState.controllers;
   return controllerData
-    ? Object.entries(controllerData).reduce(
-        (count, [, controllers]) => count + controllers.length,
+    ? Object.values(controllerData).reduce(
+        (count, controllers) => count + controllers.length,
         0
       )
     : 0;
--- a/src/store/juju/selectors.ts
+++ b/src/store/juju/selectors.ts
@@ -184,6 +184,16 @@ export const getControllerData = createSelector(
   (sliceState) => sliceState.controllers
 );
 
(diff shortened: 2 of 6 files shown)


src/juju/jimm/JIMMV3.ts
```diff
@@ -93,7 +93,7 @@ class JIMMV3 {
     });
   }
 
-  listControllers(): Promise<{ controllers: ControllerInfo[] }> {
+  listControllers(): Promise<{ controllers: ControllerInfo[] | null }> {
```
vladimir-cucu, line 96: When calling `fetchControllerList` from `src/juju/api.ts`, the `response.controllers` can be `null` in case there are 0 controllers. This can be seen in the screenshot bellow when `listControllers` is called on line 412 in `src/juju/api.ts`.

Screenshot from 2023-11-07 12-29-00

This would make it so there is an error on line 413 when .map` is called on `null`. This error will be catched and then displayed in a `Notification` component, because of the wrapper `ConnectionError`. I am not sure if this is the desired behaviour.

huwshimi:
I was able to reproduce this and your fix works!

One thing I spotted was that the controllers page still lists "1 controller" even though there are none in the table.

Screenshot 2023-11-09 at 11 32 31 am

vladimir-cucu:
> I was able to reproduce this and your fix works!
> 
> One thing I spotted was that the controllers page still lists "1 controller" even though there are none in the table.
> 
> Screenshot 2023-11-09 at 11 32 31 am

Didn't spot this issue. I'll look into it. Thanks for the review!

src/store/juju/selectors.ts
```diff
@@ -184,6 +184,16 @@ export const getControllerData = createSelector(
   (sliceState) => sliceState.controllers
 );
 
+export const getControllersCount = createSelector([slice], (sliceState) => {
+  const controllerData = sliceState.controllers;
+  return controllerData
+    ? Object.entries(controllerData).reduce(
+        (count, [, controllers]) => count + controllers.length,
```
huwshimi, line 191: You could use `Object.values` here so that you don't have to destructure the array.

```suggestion
    ? Object.values(controllerData).reduce(
        (count, controllers) => count + controllers.length,
```

huwshimi (approved):
Thanks for tracking down that bug. This is much better.

src/store/juju/selectors.ts
vladimir-cucu: Good catch! Changed it in my latest commit.

PR #1654, opened 2023-11-08
WD-7354 - fix: Fix Audit Logs filter

vladimir-cucu:
## Done

- Fix `Cannot destructure property 'value' of 'option' as it is undefined` when accessing `Audit Logs` filter.

## QA

- Without the changes, access Audit Logs on JIMM and try to open the filter. Error `Cannot destructure property 'value' of 'option' as it is undefined` should be displayed on screen.
- With the changes, the error should disapear and filter should work as expected.

## Details
- https://warthogs.atlassian.net/browse/WD-7354

## Screenshots

Screenshot from 2023-11-08 14-57-26

### DIFF
--- a/src/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/src/components/AutocompleteInput/AutocompleteInput.tsx
@@ -14,30 +14,25 @@ type Option = PropsWithSpread<
   HTMLProps<HTMLOptionElement>
 >;
 
-const isOption = (option: any): option is Option =>
-  option instanceof Object &&
-  option?.label &&
-  typeof option?.value === "string";
-
-type Props = PropsWithSpread<{ options: unknown[] }, InputProps>;
+type Props = PropsWithSpread<
+  {
+    options?: (string | Option)[] | null;
+  },
+  InputProps
+>;
 
 const generateOptions = (options: Props["options"]) =>
-  options
-    .filter((option) => typeof option === "string" || isOption(option))
-    .map((option) => {
-      if (typeof option === "string") {
-        return <option value={option} key={option} />;
-      }
-      if (isOption(option)) {
-        const { value, label, ...props } = option;
-        return (
-          <option value={value} key={value} {...props}>
-            {label}
-          </option>
-        );
-      }
-      return null;
-    });
+  options?.map((option) => {
+    if (typeof option === "string") {
+      return <option value={option} key={option} />;
+    }
+    const { value, label, ...props } = option;
+    return (
+      <option value={value} key={value} {...props}>
+        {label}
+      </option>
+    );
+  });
 
 const AutocompleteInput = ({ options, ...props }: Props) => {
   const id = useId();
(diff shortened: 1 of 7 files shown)


src/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -14,25 +14,30 @@ type Option = PropsWithSpread<
...
-    options?: (string | Option)[] | null;
-  },
-  InputProps
->;
+const isOption = (option: any): option is Option =>
+  option instanceof Object &&
+  option?.label &&
+  typeof option?.value === "string";
+
+type Props = PropsWithSpread<{ options: unknown[] }, InputProps>;
```
vladimir-cucu, line 22: This proposed change is due to the fact that this component is only used in `src/panels/AuditLogsFilterPanel/Fields/Fields.tsx`, where all the `options` property are of type `unknown[]`.

src/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -14,25 +14,30 @@ type Option = PropsWithSpread<
...
+      }
+      if (isOption(option)) {
+        const { value, label, ...props } = option;
+        return (
+          <option value={value} key={value} {...props}>
+            {label}
+          </option>
+        );
+      }
+      return null;
```
vladimir-cucu, line 39: This is theoretically redundant (considering we are using the `filter` method before it), but I think `map` method expects to get a return for all potential paths.

vladimir-cucu: Codecov also fails, cause I can't test this line as the code will never arrive here.

huwshimi:
Thanks for this fix @vladimir-cucu. This works, but I think it shows we have an issue with our types; the selectors are losing their types and returning `unknown` which is masking the problem that we're passing values that are undefined.

I think we should fix those types and filter the undefined out before passing to the autocomplete field.

Let's have a chat next week about this.

vladimir-cucu:
> Thanks for this fix @vladimir-cucu. This works, but I think it shows we have an issue with our types; the selectors are losing their types and returning `unknown` which is masking the problem that we're passing values that are undefined.
> 
> I think we should fix those types and filter the undefined out before passing to the autocomplete field.
> 
> Let's have a chat next week about this.

That makes sense. I think that would definitely be a much cleaner fix! Thanks for the review.

src/testing/factories/juju/jimm.ts
```diff
@@ -15,17 +15,21 @@ import {
   type CrossModelQueryOffer,
 } from "./jimm-types";
 
-export const auditEventFactory = Factory.define<AuditEvent>(() => ({
+export const auditEventBaseFactory = Factory.define<AuditEvent>(() => ({
```
huwshimi, line 18: Instead of adding this second factory, I would just have one that only has the required fields (the same as this base factory but named `auditEventFactory`) and then the tests can provide any additional optional attributes.

huwshimi (approved):
This looks good now. Thanks for the improvements.

PR #1656, opened 2023-11-14
Use application layout components

huwshimi:
## Done

- Copy over application layout components.
- Replace custom elements with application layout components.

## QA

- Click through as much of the UI as possible and make sure nothing looks broken (some things will be visually different but that's OK).

## Screenshots

Screenshot 2023-11-14 at 2 13 45 pm


vladimir-cucu (approved):
Tested it on JIMM and local controller. Haven't observed any broken element. Looks good to me! :rocket:

PR #1660, opened 2023-11-28
WD-7607 - fix: Specify cloud info for JIMM controllers

vladimir-cucu:
## Done

- Specify cloud info for JIMM controllers

## QA

- Verify that, in the controllers table, JIMM enabled controllers have actual `CLOUD/REGION` info, instead of `unknown/unknown`.

## Details

- https://warthogs.atlassian.net/browse/WD-7607


huwshimi (approved):
This looks good, it seems we're missing a test for the case where the data comes from `location` so that might be worth adding as well.

PR #1658, opened 2023-11-27
WD-7495 - Make model tables responsive

huwshimi:
## Done

- Update the model status, owner and cloud tables to be responsive.

## QA

- Go to models.
- Switch between the Status, Cloud and Owner tables and make sure that the tables look OK at each 

## Details

- https://warthogs.atlassian.net/browse/WD-7495
- https://warthogs.atlassian.net/browse/WD-7496
- https://warthogs.atlassian.net/browse/WD-7497

## Screenshots

Screenshot 2023-11-27 at 3 16 21 pm
Screenshot 2023-11-27 at 3 16 44 pm
Screenshot 2023-11-27 at 3 16 57 pm

PR #1657, opened 2023-11-23
WD-7498 - Remove horizontal scrolling from controllers

huwshimi:
## Done

- Remove min widths from tables.
- Remove horizontal scrolling from controller tables.

## QA

- Go to the controllers page.
- Check that the controllers table looks good at all screen sizes

## Details

https://warthogs.atlassian.net/browse/WD-7498

## Screenshots

Screen Recording 2023-11-23 at 12 35 47 pm

PR #1661, opened 2023-11-28
WD-7494 - Make action logs table responsive

huwshimi:
## Done

- Make action logs table responsive.

## QA

- Go to the 'Action logs' tab in a model.
- If you don't have a model with action logs you might need to run some actions on an app.
- Check that the action logs table appears OK at all screen sizes.

## Details

- https://warthogs.atlassian.net/browse/WD-7494

## Screenshots

Screenshot 2023-11-28 at 1 24 43 pm
Screenshot 2023-11-28 at 1 24 56 pm
Screenshot 2023-11-28 at 1 25 11 pm

### DIFF
--- a/src/pages/EntityDetails/Model/Logs/ActionLogs/_action-logs.scss
+++ b/src/pages/EntityDetails/Model/Logs/ActionLogs/_action-logs.scss
@@ -6,13 +6,13 @@
   .entity-details__action-logs {
     td,
     th {
-      // Completion time column.
+      // Completion time.
       &:nth-child(5) {
         width: 15%;
       }
 
       @include mobile-and-medium {
-        // Completion time column.
+        // Completion time.
         &:nth-child(5) {
           display: none;
         }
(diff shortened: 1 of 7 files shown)


src/pages/EntityDetails/Model/Logs/ActionLogs/_action-logs.scss
```diff
@@ -1,21 +1,33 @@
...
+@import "../../../../../scss/breakpoints";
 
 @mixin action-logs {
   .entity-details__action-logs {
-    table {
-      // Operation ID column
-      th:nth-child(2) {
+    td,
+    th {
+      // Completion time column.
```
vladimir-cucu, line 9: Small nitpicky remark: Was wondering if maybe we can standardize the way we write these comments and specify only the name of the column here instead, as was done in #1657 and #1658?

vladimir-cucu (approved):
Looks good and works as expected locally! :rocket: There are 2 places that were not covered with tests. Was wondering if it is worth writing tests for those instances?

PR #1669, opened 2023-11-30
WD-7504 - Make local apps table responsive

huwshimi:
## Done

- Make local apps table responsive.

## QA

- Go to the applications tab in a model.
- Check that the local apps table appears OK at all screen sizes.
- Use the input to filter the applications.
- The table should now have checkboxes. Check that it looks good at all screen sizes.

## Details

https://warthogs.atlassian.net/browse/WD-7504

## Screenshots

Screenshot 2023-11-30 at 12 45 02 pm
Screenshot 2023-11-30 at 12 45 16 pm
Screenshot 2023-11-30 at 12 45 40 pm

### DIFF
--- a/src/pages/EntityDetails/_entity-details.scss
+++ b/src/pages/EntityDetails/_entity-details.scss
@@ -233,9 +233,9 @@
 
           // Version.
           &:nth-child(3),
-          // Revision.
-          &:nth-child(5),
           // Store.
+          &:nth-child(5),
+          // Revision.
           &:nth-child(6) {
             display: none;
           }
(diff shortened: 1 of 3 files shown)


src/pages/EntityDetails/_entity-details.scss
```diff
@@ -212,79 +212,84 @@
...
+          // Message.
+          &:nth-child(7) {
+            display: none;
+          }
+        }
       }
     }
   }
 
   .entity-details__apps.selectable {
```
vladimir-cucu, line 254: Just a small thought, but was wondering if it might make sense to move this after line 251 as `&.selectable`?

src/pages/EntityDetails/_entity-details.scss
```diff
@@ -212,79 +212,84 @@
...
+          }
 
-      &:nth-child(7) {
-        width: 10%;
-      }
+          // Version.
+          &:nth-child(3),
+          // Revision.
+          &:nth-child(5),
+          // Store.
```
vladimir-cucu, line 238: I think the comments should be the other way around, as Store is 6th and Revision is 7th.

src/pages/EntityDetails/_entity-details.scss
```diff
@@ -212,79 +212,84 @@
...
+          width: 15%;
+        }
 
-      &:nth-child(8) {
-        width: 10%;
+        // Version.
+        &:nth-child(4),
+        // Revision.
+        &:nth-child(6),
+        // Store.
```
vladimir-cucu, line 282: I think here is the same issue, as the comments should be the other way around.

vladimir-cucu (approved):
Looks good to me and works as expected with a local controller! :rocket: Just a small change needed regarding changing the order of some comments.

src/pages/EntityDetails/_entity-details.scss
huwshimi: Done.

PR #1674, opened 2023-12-04
WD-7720 - feat: Update Juju Dashboard webteam demos to use JIMM demo controller

vladimir-cucu:
## Done

- Updated Juju Dashboard webteam demos to use JIMM demo controller.

## Details

- https://warthogs.atlassian.net/browse/WD-7720

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,12 @@ RUN yarn config set httpProxy $HTTP_PROXY
 RUN yarn config set httpsProxy $HTTPS_PROXY
 RUN yarn install
 
+COPY public/config.demo.js public/config.local.js
+
 # Build stage: Run "yarn run build-js"
 # ===
 FROM yarn-dependencies AS build-js
 ADD . .
-COPY config.demo.js config.local.js
 RUN yarn run build
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,11 +9,9 @@ WORKDIR /srv
 COPY .yar[n] ./.yarn
 COPY package.json yarn.lock .yarnrc.yml ./
 RUN if [ -n "$HTTP_PROXY" ]; then \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -8,14 +8,22 @@ WORKDIR /srv
...
+
 RUN yarn install
 
 # Build stage: Run "yarn run build-js"
 # ===
 FROM yarn-dependencies AS build-js
 ADD . .
+RUN if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ]; then \
+      cp public/config.demo.js public/config.js; \
+    fi
```
vladimir-cucu, line 26: I wasn't able to access the Jenkins environment variables from here, so, checking if either of `HTTP_PROXY` or `HTTPS_PROXY` is defined in order to see if it is running the demo, is the best I could do.

Also, I couldn't copy to `public/config.local.js`, as the file is not being parsed. I think it might be due to the fact that `process.env.NODE_ENV` is set to `production` when the image is built, thus parsing `public/config.local.js` in `index.html` is skipped.

public/config.demo.js
```diff
@@ -0,0 +1,23 @@
...
+  // by the controller when logging in so it's not currently used directly.
+  // But it is available in the charm so putting it here in the event that we
+  // would like to use it in the future.
+  identityProviderURL: "",
+  // Is this application being rendered in Juju and not JAAS. This flag should
+  // only be used for superficial updates like logos. Use feature detection
+  // for other environment features.
+  isJuju: false,
+  // If true, then Google Analytics and Sentry data will be sent.
+  analyticsEnabled: false,
```
vladimir-cucu, line 22: Should this be `true` or `false`?

huwshimi (approved):
Thanks the demo service works again, and building this locally without the proxy and build args also works so this should fix the issue for the charms.

PR #1687, opened 2024-01-24
WD-8431 - Fetch and store secrets

huwshimi:
## Done

- Fetch and store secrets.

## QA

- Set up a controller with Juju 3.3.1.
- Add some secrets to a model with: `juju add-secret my-apitoken token=34ae35facd4`
- Open the dashboard and navigate to the model.
- Click on the Secrets tab.
- The secrets should get fetch and displayed in a list.

## Details

https://warthogs.atlassian.net/browse/WD-8431

### DIFF
--- a/src/juju/apiHooks.ts
+++ b/src/juju/apiHooks.ts
@@ -44,8 +44,7 @@ export const useModelConnection = (
     )
       .then(response)
       .catch((error) => {
-        const message = error instanceof Error ? error.message : error;
-        response(null, typeof message === "string" ? message : "Unknown error");
+        response(null, error instanceof Error ? error.message : error);
       });
   }, [
     response,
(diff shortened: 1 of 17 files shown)


src/juju/apiHooks.ts
```diff
@@ -0,0 +1,114 @@
...
+import {
+  getConfig,
+  getUserPass,
+  getWSControllerURL,
+} from "store/general/selectors";
+import { actions as jujuActions } from "store/juju";
+import { getModelByUUID, getModelUUIDFromList } from "store/juju/selectors";
+import { useAppSelector, useAppDispatch } from "store/store";
+
+export const useModelConnection = (
```
vladimir-cucu, line 19: Nice addition! I see that we are only using this hook here for the moment. Are we exporting it because we might use it in the future, or only doing so for testing purposes?

vladimir-cucu (approved):
Looks good to me! Tested it yesterday as described in the QA section and works as expected with local juju v3.3.1.

src/juju/apiHooks.ts
huwshimi: I'm hoping we can use this for all API calls that currently have the Redux store state or the dispatch method passed as arguments.

PR #1688, opened 2024-01-26
WD-8308 - feat: Add comments to errors that should continue to only be logged to the console

vladimir-cucu:
## Done

- Add comments to errors that should continue to only be logged to the console

## Details

- https://warthogs.atlassian.net/browse/WD-8308


src/serviceWorker.js
```diff
@@ -94,6 +94,8 @@ function registerValidSW(swUrl, config) {
       };
     })
     .catch((error) => {
+      // Entire code is autogenerated from CRA. The dashboard will continue to
+      // work if we get an error here. Logged for debugging purposses.
```
huwshimi, line 98: I don't think we should modify this file as it is autogenerated.

src/setupTests.js
```diff
@@ -12,6 +12,7 @@ const animateMock = () => {
 if (!window.HTMLDivElement.prototype.animate) {
   window.HTMLDivElement.prototype.animate = animateMock;
 } else {
+  // Error is related to tests. Not shown in UI. Logged for debugging purposses.
```
huwshimi, line 15: I don't think we need this comment as it's not part of the shipped client.

src/store/store.ts
```diff
@@ -15,6 +15,8 @@ if (process.env.NODE_ENV !== "production" && process.env.REACT_APP_MOCK_STORE) {
   try {
     preloadedState = JSON.parse(process.env.REACT_APP_MOCK_STORE);
   } catch (error) {
+    // Error is related to tests and is not accessible in production.
+    // Not shown in UI. Logged for debugging purposses.
```
huwshimi, line 19: I don't think we need this comment as it's excluded from production, but fine to leave it here I guess.

src/hooks/useLocalStorage.ts
```diff
@@ -11,6 +11,7 @@ function useLocalStorage<V>(
       const item = window.localStorage.getItem(key);
       return item ? JSON.parse(item) : initialValue;
     } catch (error) {
+      // Not shown in UI. Logged for debugging purposses.
```
huwshimi, line 14: A spelling fix here for all these comments.
```suggestion
      // Not shown in UI. Logged for debugging purposes.
```

huwshimi (approved):
A few small comments.

src/hooks/useLocalStorage.ts
vladimir-cucu: I've fixed this and also fixed the same issue on line 28. Thanks for the review!

src/store/store.ts
vladimir-cucu: I've removed it.

PR #1690, opened 2024-01-29
WD-8541 - fix: Scroll to the top of Config Panel if error occurs

vladimir-cucu:
## Done

- Scroll to the top of Config Panel if error occurs.

#QA

- A possible way to QA the changes is to go to the config panel of an app that has a field which is required to be a number. Inspect the page, change the input type to `text`, write invalid data and then press "Save and apply". The panel should scroll to the top and the error should be displayed inline.

## Details

- https://warthogs.atlassian.net/browse/WD-8541


src/components/ScrollOnRender/ScrollOnRender.tsx
```diff
@@ -1,14 +1,17 @@
...
 type Props = {
   scrollArea?: HTMLElement | null;
 } & PropsWithChildren;
 
 const ScrollOnRender = ({ children, scrollArea }: Props) => {
-  const onRenderRef = useScrollOnRender(scrollArea);
-  return <div ref={onRenderRef}>{children}</div>;
+  useEffect(() => {
+    if (scrollArea) {
+      scrollArea.scrollIntoView();
```
huwshimi, line 10: It might be nice to set `{ behavior: "smooth" }`.

src/components/ScrollOnRender/ScrollOnRender.tsx
```diff
@@ -1,14 +1,17 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
huwshimi, line 13: Is there a reason we don't want to have `scrollArea` in the dependencies? If `scrollArea` is initially undefined and then gets defined wouldn't we want the `useEffect` to get called again?

huwshimi (commented):
This is much cleaner! I've just got a question about the useEffect deps.

huwshimi (approved):
Just needs the useEffect dependency and move the scrollIntoView mock to the setupTests.js file.

src/components/ScrollOnRender/ScrollOnRender.tsx
vladimir-cucu: Done! Although it is strange that, on my laptop running Ubuntu 22.04, the scrolling animation is not smooth in Chrome.

PR #1691, opened 2024-01-29
WD-8528 - feat: Display errors in Actions Panel

vladimir-cucu:
## Done

- Display inline error when trying to get actions for application.
- Display error in toast when trying to execute action on units.

## TODO
- Write new tests and fix old tests.
- Remove unused `EXECUTE_ACTION_ERROR` label.

## Details

- https://warthogs.atlassian.net/browse/WD-8528

### DIFF
--- a/src/panels/ActionsPanel/ActionsPanel.tsx
+++ b/src/panels/ActionsPanel/ActionsPanel.tsx
@@ -1,7 +1,12 @@
 import type { ActionSpec } from "@canonical/jujulib/dist/api/facades/action/ActionV7";
-import { Button, ConfirmationModal } from "@canonical/react-components";
+import {
+  Button,
+  ConfirmationModal,
+  Notification,
+} from "@canonical/react-components";
 import type { MutableRefObject } from "react";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import reactHotToast from "react-hot-toast";
 import { useSelector } from "react-redux";
 import { useParams } from "react-router-dom";
 import usePortal from "react-useportal";
@@ -11,6 +16,7 @@ import LoadingHandler from "components/LoadingHandler/LoadingHandler";
 import Panel from "components/Panel";
 import RadioInputBox from "components/RadioInputBox/RadioInputBox";
 import type { EntityDetailsRoute } from "components/Routes/Routes";
+import ToastCard from "components/ToastCard/ToastCard";
 import { executeActionOnUnits, getActionsForApplication } from "juju/api";
 import { usePanelQueryParams } from "panels/hooks";
 import type { ConfirmTypes } from "panels/types";
@@ -26,7 +32,7 @@ export enum Label {
   CONFIRM_BUTTON = "Confirm",
   NO_UNITS_SELECTED = "0 units selected",
   NO_ACTIONS_PROVIDED = "This charm has not provided any actions.",
-  GET_ACTIONS_ERROR = "Error while trying to get actions for application.",
+  GET_ACTIONS_ERROR = "Unable to get actions for application.",
   EXECUTE_ACTION_ERROR = "Error while trying to execute action on units.",
 }
 
@@ -101,6 +107,7 @@ export default function ActionsPanel(): JSX.Element {
     string | undefined,
     SetSelectedAction,
   ] = useState<string>();
+  const [inlineError, setInlineError] = useState<string | null>(null);
   const { Portal } = usePortal();
 
   const actionOptionsValues = useRef<ActionOptionValues>({});
@@ -119,9 +126,13 @@ export default function ActionsPanel(): JSX.Element {
             setActionData(actions.results[0].actions);
           }
           setFetchingActionData(false);
+          setInlineError(null);
           return;
         })
-        .catch((error) => console.error(Label.GET_ACTIONS_ERROR, error));
+        .catch((error) => {
+          setInlineError(Label.GET_ACTIONS_ERROR);
+          console.error(Label.GET_ACTIONS_ERROR, error);
+        });
     }
   }, [appName, appStore, modelUUID]);
 
@@ -130,14 +141,14 @@ export default function ActionsPanel(): JSX.Element {
       ? appState.juju?.modelData?.[modelUUID]?.applications?.[appName]?.charm
       : null;
 
-  const generateSelectedUnitList = () => {
+  const generateSelectedUnitList = useCallback(() => {
     if (!selectedUnits.length) {
       return Label.NO_UNITS_SELECTED;
     }
     return selectedUnits.reduce((acc, unitName) => {
       return `${acc}, ${unitName.split("/")[1]}`;
     });
-  };
+  }, [selectedUnits]);
 
   const generateTitle = () => {
     const unitLength = selectedUnits.length;
@@ -213,9 +224,15 @@ export default function ActionsPanel(): JSX.Element {
               confirmButtonAppearance="positive"
               onConfirm={() => {
                 setConfirmType(null);
-                executeAction().catch((error) =>
-                  console.error(Label.EXECUTE_ACTION_ERROR, error),
-                );
+                executeAction().catch((error) => {
+                  const errorMessage = `Unable to execute ${selectedAction} action on units: ${generateSelectedUnitList()}.`;
+                  reactHotToast.custom((t) => (
+                    <ToastCard toastInstance={t} type="negative">
+                      {errorMessage}
+                    </ToastCard>
+                  ));
+                  console.error(errorMessage, error);
+                });
                 handleRemovePanelQueryParams();
               }}
               close={() => setConfirmType(null)}
@@ -256,6 +273,9 @@ export default function ActionsPanel(): JSX.Element {
       onRemovePanelQueryParams={handleRemovePanelQueryParams}
     >
       <p data-testid="actions-panel-unit-list">
+        {inlineError ? (
+          <Notification severity="negative">{inlineError}</Notification>
+        ) : null}
         Run action on: {generateSelectedUnitList()}
       </p>
       <LoadingHandler


src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -213,9 +224,15 @@ export default function ActionsPanel(): JSX.Element {
...
                 setConfirmType(null);
-                executeAction().catch((error) =>
-                  console.error(Label.EXECUTE_ACTION_ERROR, error),
-                );
+                executeAction().catch((error) => {
+                  const errorMessage = `Unable to execute ${selectedAction} action on units: ${generateSelectedUnitList()}.`;
+                  reactHotToast.custom((t) => (
+                    <ToastCard toastInstance={t} type="negative">
+                      {errorMessage}
+                    </ToastCard>
```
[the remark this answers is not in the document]
vladimir-cucu: I've changed it to inline error as discussed in the morning.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -211,12 +244,21 @@ export default function ActionsPanel(): JSX.Element {
               cancelButtonLabel={Label.CANCEL_BUTTON}
               confirmButtonLabel={Label.CONFIRM_BUTTON}
               confirmButtonAppearance="positive"
-              onConfirm={() => {
+              onConfirm={(event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+                event.nativeEvent.stopImmediatePropagation();
```
vladimir-cucu, line 248: This line of code (and maybe similar lines in other places) might become redundant in the case [this gets implemented](https://github.com/canonical/react-components/issues/1032).

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -119,9 +145,16 @@ export default function ActionsPanel(): JSX.Element {
             setActionData(actions.results[0].actions);
           }
           setFetchingActionData(false);
+          setInlineErrors((prevInlineErrors) => [null, prevInlineErrors[1]]);
```
huwshimi, line 148: This case (and the other setInlineErrors() calls below) would lead to bugs if the total number of errors in the tuple changed (e.g. if a third error was added it would get lost during this call). Instead we should only update the slot that matches the error we're concerned with.

If you're concerned about mutating state then you could do something like:
```
errors = [... prevInlineErrors];
errors[0] = null;
```

src/panels/ActionsPanel/ActionsPanel.tsx
huwshimi: Thanks for filing that issue. I think it would be good to leave a comment here about why this `stopImmediatePropagation` is necessary and probably link to the issue you filed as well.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -211,12 +244,21 @@ export default function ActionsPanel(): JSX.Element {
...
               confirmButtonAppearance="positive"
-              onConfirm={() => {
+              onConfirm={(event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+                event.nativeEvent.stopImmediatePropagation();
                 setConfirmType(null);
-                executeAction().catch((error) =>
-                  console.error(Label.EXECUTE_ACTION_ERROR, error),
-                );
-                handleRemovePanelQueryParams();
+                executeAction()
```
huwshimi, line 250: I think this should clear the errors here in case the user submits the form again. Also, depending on how long this action takes there could be a pause before the panel closes. You could show a loading state in the button. You'd need a new state (e.g. "const [saving, setSaving] = useState(false);".

```suggestion
                setSaving(true);
                executeAction()
```

You'll also need to update the button on line 287 to be `<ActionButton` and then give it the props `disabled={saving} loading={saving}`.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -211,12 +244,21 @@ export default function ActionsPanel(): JSX.Element {
...
+                  .then(() => {
+                    handleRemovePanelQueryParams();
+                    return;
+                  })
+                  .catch((error) => {
+                    setInlineErrors((prevInlineError) => [
```
huwshimi, line 256: If you set the saving state when the action starts then you'll need to update it again here:

```suggestion
                  .catch((error) => {
                    setSaving(false);
                    setInlineErrors((prevInlineError) => [
```

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -119,9 +145,16 @@ export default function ActionsPanel(): JSX.Element {
             setActionData(actions.results[0].actions);
           }
           setFetchingActionData(false);
+          setInlineErrors((prevInlineErrors) => [null, prevInlineErrors[1]]);
           return;
         })
-        .catch((error) => console.error(Label.GET_ACTIONS_ERROR, error));
+        .catch((error) => {
+          setInlineErrors((prevInlineErrors) => [
```
huwshimi, line 152: Here also we need to only update a single error slot so we don't accidentally remove errors if the number of errors in the tuple changes.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -211,12 +244,21 @@ export default function ActionsPanel(): JSX.Element {
...
-                handleRemovePanelQueryParams();
+                executeAction()
+                  .then(() => {
+                    handleRemovePanelQueryParams();
+                    return;
+                  })
+                  .catch((error) => {
+                    setInlineErrors((prevInlineError) => [
+                      prevInlineError[0],
+                      Label.EXECUTE_ACTION_ERROR,
```
huwshimi, line 258: Also need to only update a single slot in the tuple.

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -392,7 +395,7 @@ export default function ConfigPanel(): JSX.Element {
 
             <div className="config-panel__list">
               {formErrors
```
huwshimi, line 397: I think this should also have `.some(...` or at the very least check that this isn't an empty array.

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -57,12 +57,15 @@ type ConfigQueryParams = {
...
-  ));
+const generateErrors = (errors: string[], scrollArea: HTMLElement | null) => (
+  <ScrollOnRender scrollArea={scrollArea}>
+    {errors.map((error) => (
+      <Notification key={error} severity="negative">
+        {error}
+      </Notification>
+    ))}
+  </ScrollOnRender>
+);
```
huwshimi, line 68: Possibly in a follow-up branch this could become a shared component, something like `PanelErrors` and include the `.some(...)` like you do in the other panel, as we're doing the same thing in a number of places.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -254,8 +296,12 @@ export default function ActionsPanel(): JSX.Element {
       data-testid={TestId.PANEL}
       title={generateTitle()}
       onRemovePanelQueryParams={handleRemovePanelQueryParams}
+      ref={scrollArea}
     >
       <p data-testid="actions-panel-unit-list">
+        {inlineErrors.some((error) => error)
+          ? generateInlineErrors(inlineErrors, scrollArea.current)
+          : null}
```
huwshimi, line 304: This needs to be outside of the wrapping `<p>` as this is currently giving errors about nested `<p>` tags.

huwshimi (commented):
I've ended up making quite a few small comments so I'll do another review of this once those changes have been made.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -119,9 +128,21 @@ export default function ActionsPanel(): JSX.Element {
...
           setFetchingActionData(false);
+          setInlineErrors((prevInlineErrors) => {
+            const newInlineErrors = [...prevInlineErrors];
+            newInlineErrors[0] = null;
+            return newInlineErrors;
+          });
           return;
         })
-        .catch((error) => console.error(Label.GET_ACTIONS_ERROR, error));
+        .catch((error) => {
```
huwshimi, line 138: I think this also needs to call `setFetchingActionData(false);` so that the spinner hides when the error appears.

src/panels/PanelInlineErrors/PanelInlineErrors.tsx
```diff
@@ -0,0 +1,26 @@
+import { Notification } from "@canonical/react-components";
+
+import ScrollOnRender from "components/ScrollOnRender";
+
+type Props = {
+  inlineErrors: (string | null)[] | null;
+  scrollArea?: HTMLElement | null;
+};
+
+const PanelInlineErrors = ({
```
huwshimi, line 10: Nice, thanks for adding this.

src/panels/PanelInlineErrors/PanelInlineErrors.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { render, screen } from "@testing-library/react";
+
+import PanelInlineErrors from "./PanelInlineErrors";
+
+describe("PanelInlineErrors", () => {
```
huwshimi, line 5: I think it would be good to also test that the `ScrollOnRender` component does not appear if `inlineErrors` is `null`, an empty array or only contains falsey items.

huwshimi (approved):
Thanks for all the changes. I've made a couple of small comments but this can land once those have been fixed.

src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -242,26 +290,51 @@ export default function ActionsPanel(): JSX.Element {
...
       data-testid={TestId.PANEL}
       title={generateTitle()}
       onRemovePanelQueryParams={handleRemovePanelQueryParams}
+      ref={scrollArea}
     >
+      <PanelInlineErrors
+        inlineErrors={
+          inlineErrors[0]
+            ? inlineErrors.map((error, index) =>
+                index === 0 ? (
```
huwshimi, line 312: I think it might be nice to leave a comment here to explain why this is necessary.

huwshimi (approved):
Just one tiny comment.

PR #1694, opened 2024-02-01
WD-8593 - handle optional juju types

huwshimi:
## Done

- Update Jujulib to 5.0.0
- Handle new optional types from Jujulib.

## Details

https://warthogs.atlassian.net/browse/WD-8593

### DIFF
--- a/src/pages/EntityDetails/Model/Logs/ActionLogs/ActionLogs.tsx
+++ b/src/pages/EntityDetails/Model/Logs/ActionLogs/ActionLogs.tsx
@@ -154,16 +154,16 @@ const generateApplicationRow = (
   // performed across multiple units of the same application. I expect
   // that the CLI may gain this functionality in the future and we'll have
   // to update this code to display the correct action name.
-  const actionName = actionData.action.name;
+  const actionName = actionData.action?.name;
   const operationId = operationData.operation.split("-")[1];
   // The receiver is in the format "unit-ceph-mon-0" to "ceph-mon"
-  const parts = actionData.action.receiver.match(/unit-(.+)-\d+/);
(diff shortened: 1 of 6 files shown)


src/pages/EntityDetails/Model/Logs/ActionLogs/ActionLogs.tsx
```diff
@@ -281,15 +284,17 @@ export default function ActionLogs() {
         const hasSterr =
           actionFullDetails.status === "failed" && !!actionFullDetails.message;
         const stdout = hasStdout
-          ? actionFullDetails.log.map((m, i) => (
+          ? actionFullDetails.log?.map((m, i) => (
```
vladimir-cucu, line 287: Just a small observation. TypeScript will state that the type of `stdout` is `JSX.Element[] | undefined`, while I think that, from the definition of `hasStdout`, it can't be `undefined`. Was thinking if adding the redundant `&& actionFullDetails.log` on line 286 might make sense?

vladimir-cucu (approved):
Looks good to me! :+1:

src/pages/EntityDetails/Model/Logs/ActionLogs/ActionLogs.tsx
huwshimi: I've improved this by assigning `actionFullDetails.log` to a variable and then using that variable when creating `hasStdout` and also inside the ternary. This way typescript knows that when `hasStdout` is `true` then the variable must not be `undefined`.

```typescript
const log = actionFullDetails.log;
const hasStdout = log && log.length > 0;
...
const stdout = hasStdout
    ? log.map((m, i) => (
```

PR #1697, opened 2024-02-02
WD-8617 - fix: Fix displaying the first colorless part of the message in WebCLI

vladimir-cucu:
## Done

- Issue resolved: Until now, if the content in the `WebCLI` started straight away with a color, we displayed twice the beginning of the content. The second time it was displayed without color and with the color encoding appended to it (see screenshots).

## Details

- https://warthogs.atlassian.net/browse/WD-8617

## Screenshots
Before:
Screenshot from 2024-02-02 12-48-11
After:
Screenshot from 2024-02-02 12-48-33

### DIFF
--- a/src/components/WebCLI/Output.test.tsx
+++ b/src/components/WebCLI/Output.test.tsx
@@ -1,0 +1,52 @@
+import { screen } from "@testing-library/react";
+
+import { renderComponent } from "testing/utils";
+
+import Output from "./Output";
+
+describe("Output", () => {
+  it("should display content and not display help message", () => {
+    renderComponent(
+      <Output
+        content="Output"
+        helpMessage="Help message"
+        showHelp={false}
+        setShouldShowHelp={jest.fn()}
+      />,
+    );
+    expect(screen.getByText("Output")).toBeInTheDocument();
+    expect(screen.queryByText("Help message")).not.toBeInTheDocument();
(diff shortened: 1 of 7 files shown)


huwshimi (approved):
Seems to work now.

src/components/WebCLI/Output.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+        helpMessage="Help message"
+        showHelp={true}
+        setShouldShowHelp={jest.fn()}
+      />,
+    );
+    expect(screen.queryByRole("Output")).not.toBeInTheDocument();
+    expect(screen.getByText("Help message")).toBeInTheDocument();
+  });
+
+  it("should display the content with correct color", () => {
```
huwshimi, line 34: I think it would be good to add a test for strings starting with a colour.

vladimir-cucu: I've changed the code to use `ansi-to-html` library instead of the previous custom code. Should we still add this test, considering that it is handled internally by the library? I've updated the previous test to check for bold and colored text.

huwshimi: No need to add that test now.

huwshimi (approved):
Thanks, this is much cleaner now.

PR #1698, opened 2024-02-02
WD-8625 - feat: Display errors in WebCLI

vladimir-cucu:
## Done

- Display connection and authentication errors in WebCLI

## Details

- https://warthogs.atlassian.net/browse/WD-8625

### DIFF
--- a/src/components/WebCLI/WebCLI.tsx
+++ b/src/components/WebCLI/WebCLI.tsx
@@ -17,7 +17,7 @@ import Connection from "./connection";
 
 import "./_webcli.scss";
 
-enum Label {
+export enum Label {
   CONNECTION_ERROR = "Unable to connect to the model.",
   AUTHENTICATION_ERROR = "Unable to authenticate.",
 }
@@ -188,7 +188,7 @@ const WebCLI = ({
       }
       setInlineErrors((prevInlineErrors) => {
         const newInlineErrors = [...prevInlineErrors];
-        newInlineErrors[1] = macaroons ? null : Label.CONNECTION_ERROR;
+        newInlineErrors[1] = macaroons ? null : Label.AUTHENTICATION_ERROR;
         return newInlineErrors;
       });
     }
@@ -225,40 +225,34 @@ const WebCLI = ({
     setShouldShowHelp(!shouldShowHelp);
   };
 
-  // If we do not have an address then do not try and render the UI.
-  if (!wsAddress) {
-    return null;
-  }
-
   return (
     <div className="webcli">
       <WebCLIOutput
-        content={
-          // If inline errors exist, display the errors instead of the output.
-          inlineErrors.some((error) => error)
-            ? (inlineErrors.reduce(
-                (errorsOutput, error) =>
-                  errorsOutput + (error ? `ERROR: ${error}\n` : ""),
-                // Use red color for displaying the errors.
-                "[31m",
-              ) as string)
-            : output
-        }
+        content={output}
         showHelp={shouldShowHelp}
         setShouldShowHelp={setShouldShowHelp}
         helpMessage={
-          <>
-            Welcome to the Juju Web CLI - see the{" "}
-            <a
-              href={externalURLs.cliHelp}
-              className="p-link--inverted"
-              rel="noreferrer"
-              target="_blank"
-            >
-              full documentation here
-            </a>
-            .
-          </>
+          // If errors exist, display them instead of the default help message.
+          inlineErrors.some((error) => error) ? (
+            <>
+              {inlineErrors.map((error) =>
+                error ? <div key={error}>ERROR: {error}</div> : null,
+              )}
+            </>
+          ) : (
+            <>
+              Welcome to the Juju Web CLI - see the{" "}
+              <a
+                href={externalURLs.cliHelp}
+                className="p-link--inverted"
+                rel="noreferrer"
+                target="_blank"
+              >
+                full documentation here
+              </a>
+              .
+            </>
+          )
         }
       />
       <div className="webcli__input">
@@ -274,6 +268,8 @@ const WebCLI = ({
             name="command"
             ref={inputRef}
             placeholder="enter command"
+            // Disable input if there is a connection error.
+            disabled={!!inlineErrors[0]}
           />
         </form>
         <div className="webcli__input-help">
--- a/src/components/WebCLI/WebCLI.tsx
+++ b/src/components/WebCLI/WebCLI.tsx
@@ -215,15 +215,15 @@ const WebCLI = ({
     <div className="webcli">
       <WebCLIOutput
         content={output}
-        showHelp={shouldShowHelp}
+        showHelp={shouldShowHelp || hasInlineError(InlineErrors.AUTHENTICATION)}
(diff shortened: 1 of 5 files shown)


src/components/WebCLI/WebCLI.tsx
```diff
@@ -41,6 +46,13 @@ const WebCLI = ({
   const [connection, setConnection] = useState<Connection | null>(null);
   const [shouldShowHelp, setShouldShowHelp] = useState(false);
   const [historyPosition, setHistoryPosition] = useState(0);
+  // First element in inLineErrors array corresponds to the error when no
+  // websocket address is provided. Second element corresponds to the error
+  // when no authentication information is available for Web CLI.
+  const [inlineErrors, setInlineErrors] = useState<(string | null)[]>([
+    null,
+    null,
+  ]);
```
vladimir-cucu, line 55: I will change this part of the code to use the hook in your PR once it lands.

src/hooks/useInlineErrors.ts
```diff
@@ -21,7 +21,8 @@ function useInlineErrors(
   const [inlineErrors, setInlineErrors] = useState<InlineError[]>([]);
   const setError = useCallback(
     (key: InlineError["key"], error: InlineError["error"]) => {
-      setInlineErrors((inlineErrors) => {
+      setInlineErrors((prevInlineErrors) => {
+        const inlineErrors = [...prevInlineErrors];
```
vladimir-cucu, line 25: I think, without this fix, the Connection error won't appear straight away in the `WebCLI`, as we were mutating the state here directly.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -41,6 +46,13 @@ const WebCLI = ({
   const [connection, setConnection] = useState<Connection | null>(null);
   const [shouldShowHelp, setShouldShowHelp] = useState(false);
   const [historyPosition, setHistoryPosition] = useState(0);
+  // First element in inLineErrors array corresponds to the error when no
+  // websocket address is provided. Second element corresponds to the error
+  // when no authentication information is available for Web CLI.
+  const [inlineErrors, setInlineErrors] = useState<(string | null)[]>([
```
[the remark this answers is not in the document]
vladimir-cucu: Done

src/hooks/useInlineErrors.ts
huwshimi: I'd always understood updater functions to be safe as they do not mutate the current state, but rather the functions get queued to be reduced on the next render. However, I couldn't see anywhere in the documentation where it mentions whether it is safe to mutate the previous state inside the function.

I dug into the source and it looks like the updater function is just a reducer function passed to useReducer and in the docs for useReducer it says to use a new object so it doesn't have the same reference.

Saying all that, was there a specific bug that you were running into? I tried both update methods and the render order appeared to be working correctly for me.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -197,30 +211,34 @@ const WebCLI = ({
...
-              className="p-link--inverted"
-              rel="noreferrer"
-              target="_blank"
-            >
-              full documentation here
-            </a>
-            .
-          </>
+          // If errors exist, display them instead of the default help message.
+          Object.values(InlineErrors).some((error) => hasInlineError(error)) ? (
```
huwshimi, line 222: I think we should update `hasInlineError` to check all errors if you don't pass a key.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -197,30 +211,34 @@ const WebCLI = ({
...
+            <>
+              {inlineErrors.map((error) =>
+                error ? <div key={error as string}>ERROR: {error}</div> : null,
```
huwshimi, line 225: `inlineErrors` are reduced to truthy values before being returned by `useInlineErrors` so this ternary isn't necessary.

src/hooks/useInlineErrors.test.ts
```diff
@@ -1,12 +1,12 @@
...
 import useInlineErrors from "./useInlineErrors";
 
 describe("useInlineErrors", () => {
   it("can add an error", () => {
     const { result, rerender } = renderHook(() => useInlineErrors());
     let [errors, setError] = result.current;
-    setError("Error1", "Uh oh!");
+    act(() => setError("Error1", "Uh oh!"));
     rerender();
     [errors] = result.current;
```
huwshimi, line 11: Looks like I somehow doubled up the rerender and destructure in this test so you could remove these two lines.

src/hooks/useInlineErrors.test.ts
```diff
@@ -17,11 +17,11 @@ describe("useInlineErrors", () => {
   it("can update an error", () => {
     const { result, rerender } = renderHook(() => useInlineErrors());
     let [errors, setError] = result.current;
-    setError("Error1", "Uh oh!");
+    act(() => setError("Error1", "Uh oh!"));
     rerender();
```
huwshimi, line 21: Now that these are using `act()` I don't think they need to also have `rerender()`.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -159,10 +172,11 @@ const WebCLI = ({
         );
         authentication.user = activeUser ? getUserName(activeUser) : undefined;
         authentication.macaroons = [deserialized];
-      } else {
-        // XXX Surface error to the user.
-        console.error("No authentication information available");
       }
+      setInlineError(
+        InlineErrors.AUTHENTICATION,
+        macaroons ? null : Label.AUTHENTICATION_ERROR,
```
huwshimi, line 178: While I don't think we should pop open the console if there is a connection error, I think we should open it when someone tries to send a command and then get this error.

src/hooks/useInlineErrors.ts
vladimir-cucu: Thanks for the detailed explanation! After today's discussion, I looked over the code a bit and I think I'll use `immer` here. This is mainly cause, in some cases, we might make a shallow copy here (e.g. when the error has type `Iterable<ReactNode>`), which might prove to have the same issue we are trying to fix right now.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Done.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Nice catch! Changed it.

src/hooks/useInlineErrors.test.ts
vladimir-cucu: Removed this part.

src/hooks/useInlineErrors.test.ts
vladimir-cucu: Removed all the `rerender()` bits.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Makes sense, added this in my latest commit.

huwshimi (approved):

PR #1699, opened 2024-02-05
Store inline errors by key

huwshimi:
## Done

- Add a hook to store inline errors by key.
- Update ActionPanel to use the hook.

## QA

- Open the action panel.
- Check that inline errors appear like they did before (you might need to throw some errors in the component).

### DIFF
--- a/src/panels/ActionsPanel/ActionsPanel.tsx
+++ b/src/panels/ActionsPanel/ActionsPanel.tsx
@@ -15,6 +15,7 @@ import LoadingHandler from "components/LoadingHandler/LoadingHandler";
 import Panel from "components/Panel";
 import RadioInputBox from "components/RadioInputBox/RadioInputBox";
 import type { EntityDetailsRoute } from "components/Routes/Routes";
+import useInlineErrors from "hooks/useInlineErrors";
 import { executeActionOnUnits, getActionsForApplication } from "juju/api";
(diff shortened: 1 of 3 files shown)


src/panels/ActionsPanel/ActionsPanel.tsx
```diff
@@ -106,12 +112,22 @@ export default function ActionsPanel(): JSX.Element {
     string | undefined,
     SetSelectedAction,
   ] = useState<string>();
-  // First element in inLineErrors array corresponds to the get actions for
-  // application error. Second element corresponds to execute action error.
-  const [inlineErrors, setInlineErrors] = useState<(string | null)[]>([
-    null,
-    null,
-  ]);
+  const [inlineErrors, setInlineErrors, hasError] = useInlineErrors({
```
vladimir-cucu, line 115: For consistency reasons, maybe we could name `hasError` as `hasInlineErrors` instead?

vladimir-cucu (approved):
Thank you! Using the hook will make the code a lot more readable and maintainable. Tested it locally and works as expected. :rocket:

src/panels/ActionsPanel/ActionsPanel.tsx
huwshimi: Done.

PR #1703, opened 2024-02-06
WD-8729 - feat: Display errors in Config Panel

vladimir-cucu:
## Done

- Displayed get config and submit to juju errors in Config Panel.

## Details

https://warthogs.atlassian.net/browse/WD-8729

### DIFF
--- a/src/panels/ConfigPanel/ConfigPanel.tsx
+++ b/src/panels/ConfigPanel/ConfigPanel.tsx
@@ -52,6 +52,7 @@ export enum TestId {
 }
 
 enum InlineErrors {
+  FORM = "form",
   GET_CONFIG = "get-config",
   SUBMIT_TO_JUJU = "submit-to-juju",
 }
@@ -74,7 +75,6 @@ export default function ConfigPanel(): JSX.Element {
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [savingConfig, setSavingConfig] = useState<boolean>(false);
   const [confirmType, setConfirmType] = useState<ConfirmTypes>(null);
-  const [formErrors, setFormErrors] = useState<string[] | null>(null);
   const [inlineErrors, setInlineError] = useInlineErrors({
     [InlineErrors.GET_CONFIG]: (error) => (
       <>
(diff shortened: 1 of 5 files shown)


src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -69,6 +75,23 @@ export default function ConfigPanel(): JSX.Element {
   const [savingConfig, setSavingConfig] = useState<boolean>(false);
   const [confirmType, setConfirmType] = useState<ConfirmTypes>(null);
   const [formErrors, setFormErrors] = useState<string[] | null>(null);
+  const [inlineErrors, setInlineError] = useInlineErrors({
```
huwshimi, line 78: I think we should merge these two error states. To do this we would need to update `useInlineErrors` to output an error for each item if the error is an array, something like this:

```
const errors = inlineErrors.reduce<ReactNode[]>((nodes, { key, error }) => {
    const errorItems = mapping && key in mapping ? mapping[key](error) : error;
    const filteredItems = (Array.isArray(errorItems) ? errorItems : [errorItems]).filter(Boolean);
    return nodes.concat(filteredItems);
  }, []);
```

vladimir-cucu: Done!

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -194,7 +222,7 @@ export default function ConfigPanel(): JSX.Element {
     setEnableSave(false);
     setConfirmType(null);
     if (errors?.length) {
-      setFormErrors(errors);
+      setInlineError(InlineErrors.FORM, errors);
       return;
     }
     await getConfig(
```
huwshimi, line 228: I might be missing something, but I can't see why it fetches the config once it saves successfully as all it does after that is close the panel.

If we remove this call then `getConfig` no longer needs to return a promise, so we can avoid the `void` inside `getConfigCallback`.

huwshimi (approved):
This looks good, I've just made one suggestion that's not something you've changed but should improve performance and avoid a `void`.

PR #1711, opened 2024-02-12
WD-8829 - feat: display errors in Charms and Actions Panel

vladimir-cucu:
## Done

- Display get charm application(s) error (triggered in `CharmsAndActionsPanel`) in `CharmsPanel`.

## Details

- https://warthogs.atlassian.net/browse/WD-8829

## Screenshots

Screenshot from 2024-02-12 17-42-51

### DIFF
--- a/src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
+++ b/src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
@@ -1,4 +1,5 @@
 import { act, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 
 import * as juju from "juju/api";
 import { getCharmsURLFromApplications } from "juju/api";
@@ -148,7 +149,8 @@ describe("CharmsAndActionsPanel", () => {
     expect(screen.getByRole("button", { name: "Next" })).toBeEnabled();
   });
 
-  it("should show console error when calling getCharmsURLFromApplications", async () => {
+  it("should show should show error in CharmsPanel", async () => {
+    state.juju.charms = [];
     jest
       .spyOn(juju, "getCharmsURLFromApplications")
       .mockImplementation(
@@ -158,7 +160,9 @@ describe("CharmsAndActionsPanel", () => {
             new Error("Error while calling getCharmsURLFromApplications"),
           ),
       );
-    renderComponent(<CharmsAndActionsPanel />, { path, url, state });
+    const {
+      result: { container },
+    } = renderComponent(<CharmsAndActionsPanel />, { path, url, state });
     expect(juju.getCharmsURLFromApplications).toHaveBeenCalledTimes(1);
     await waitFor(() =>
       expect(console.error).toHaveBeenCalledWith(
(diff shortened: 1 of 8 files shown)


src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
```diff
@@ -158,13 +160,32 @@ describe("CharmsAndActionsPanel", () => {
...
       ),
     );
+    expect(container.querySelector(".p-panel__title")).toContainHTML(
+      CharmsPanelLabel.PANEL_TITLE,
+    );
+    const getCharmsURLErrorNotification = screen.getByText(
+      new RegExp(
+        CharmsAndActionsPanelLabel.GET_URL_ERROR.replace(
+          /[.*+?^${}()|[\]\\]/g,
+          "\\$&",
```
vladimir-cucu, line 180: This is needed in order to escape the special characters of regular expressions, in our case "(" and ")". Was wondering if it might make sense to make a utility function out of this and use it instead of `RegExp`?

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.tsx
```diff
@@ -62,10 +77,32 @@ const CharmsAndActionsPanel = () => {
...
+        setInlineErrors(InlineErrors.GET_URL, null);
         return;
       })
-      .catch((error) => console.error(Label.GET_URL_ERROR, error));
-  }, [appState, dispatch, isPanelLoading, modelUUID, selectedApplications]);
+      .catch((error) => {
+        setInlineErrors(InlineErrors.GET_URL, Label.GET_URL_ERROR);
+        console.error(Label.GET_URL_ERROR, error);
+      });
+  }, [appState, dispatch, modelUUID, selectedApplications, setInlineErrors]);
```
huwshimi, line 87: This useCallback is passed `appState` which means that this callback gets recalculated every time the redux store changes. If there's an error during the `getCharmsURLFromApplications` it means that this will enter an infinite loop while the store gets updated, this callback gets recalculated and then the useEffect below gets run (to see this add `throw new Error();` before the return in `getCharmsURLFromApplications` in api.ts).

To prevent this we can change `const appState = useAppStore().getState();` to instead be a reference to the `getState` function `const getState = useAppStore().getState;` which won't change and then call that inside the `useCallback`.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
huwshimi: In this particular case I think RTL provides a couple of other ways to check for the substring that means we don't have to escape the error message:

With the `exact` option:
```typescript
    const getCharmsURLErrorNotification = screen.getByText(
      CharmsAndActionsPanelLabel.GET_URL_ERROR,
      { exact: false },
    );
```

With a custom matcher:
```typescript
    const getCharmsURLErrorNotification = screen.getByText((content) =>
      content.startsWith(CharmsAndActionsPanelLabel.GET_URL_ERROR),
    );
```

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.tsx
vladimir-cucu: Thanks for the explanation! Fixed.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
vladimir-cucu: Done. Used the first proposed fix and also used this fix in other places to combat future potential bugs.

huwshimi (approved):
This works well now.

PR #1712, opened 2024-02-13
WD-8402 - grant/revoke secrets

huwshimi:
## Done

- Add a panel for granting/revoking secrets.

## QA

- Go to the secrets tab for a model with more than one application.
- Open the action menu for a secret.
- Click "Grant"
- Choose an application and save.
- The panel should close and the number of granted counts should update.
- Open the grant panel again and deselect the selected app and select another and save.
- Reopen the panel and the checkboxes should reflect the changes you just made.

## Details

https://warthogs.atlassian.net/browse/WD-8403
https://warthogs.atlassian.net/browse/WD-8402

### DIFF
--- a/src/juju/apiHooks.test.tsx
+++ b/src/juju/apiHooks.test.tsx
@@ -1312,11 +1312,6 @@ describe("useGrantSecret", () => {
   it("can grant apps", async () => {
     const store = mockStore(state);
     changeURL("/models/eggman@external/group-test/app/etcd");
-    const secrets = [
-      {
-        uri: "secret:aabbccdd",
-      },
-    ];
     const results = { results: [{ result: "secret:def456" }] };
     const grantSecret = jest
       .fn()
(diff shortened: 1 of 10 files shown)


vladimir-cucu (commented):
I might be wrong, but it seems like we can't grant access to the owner application. We receive the error as in the attached screenshot, where I tried to grant access to the owner application of the first secret in the list. If we select multiple applications (including the owner application), we get the same error, however the rest of the applications are granted access. Was wondering if we might be able to disable and tick the box with the owner application?

Screenshot from 2024-02-13 18-11-46

vladimir-cucu (approved):

huwshimi:
> I might be wrong, but it seems like we can't grant access to the owner application. We receive the error as in the attached screenshot, where I tried to grant access to the owner application of the first secret in the list. If we select multiple applications (including the owner application), we get the same error, however the rest of the applications are granted access. Was wondering if we might be able to disable and tick the box with the owner application?
> 

Thanks for finding that, I've disabled and ticked the box as you've suggested.

PR #1714, opened 2024-02-13
WD-8855 - feat: display model-poller errors

vladimir-cucu:
## Done

- Displayed the Audit Logs user permission check error.

## QA

- Run either demo or staging of JIMM.
- Throw an error on line 155 in `store/middleware/model-poller.ts`.
- A toast identical to the one in the attached screenshot should appear.

## Details

- https://warthogs.atlassian.net/browse/WD-8855

## Screenshots

Screenshot from 2024-02-13 16-38-42

### DIFF
--- a/src/components/ConnectionError/ConnectionError.test.tsx
+++ b/src/components/ConnectionError/ConnectionError.test.tsx
@@ -45,10 +45,13 @@ describe("ConnectionError", () => {
     const state = rootStateFactory.withGeneralConfig().build({
       juju: { auditEvents: { errors: "Oops!" } },
     });
-    const {
-      result: { rerender },
-    } = renderComponent(<ConnectionError />, { state });
-    rerender(<Toaster />);
+    renderComponent(
+      <>
+        <Toaster />
+        <ConnectionError />
+      </>,
+      { state },
+    );
     const auditLogsErrorNotification = screen.getByText(/Oops!/);
     expect(auditLogsErrorNotification.childElementCount).toBe(1);
     const refreshButton = auditLogsErrorNotification.children[0];
--- a/src/components/ConnectionError/ConnectionError.test.tsx
+++ b/src/components/ConnectionError/ConnectionError.test.tsx
@@ -1,5 +1,6 @@
-import { screen } from "@testing-library/react";
+import { act, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
+import reactHotToast, { Toaster } from "react-hot-toast";
 
 import { rootStateFactory } from "testing/factories/root";
 import { renderComponent } from "testing/utils";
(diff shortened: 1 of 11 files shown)


src/components/ConnectionError/ConnectionError.tsx
```diff
@@ -1,11 +1,28 @@
...
+import ToastCard from "components/ToastCard/ToastCard";
 import { getConnectionError } from "store/general/selectors";
+import { getAuditEventsErrors } from "store/juju/selectors";
 import { useAppSelector } from "store/store";
 
 const ConnectionError = ({ children }: PropsWithChildren) => {
   const error = useAppSelector(getConnectionError);
+  const auditLogsErrors = useAppSelector(getAuditEventsErrors);
+
+  if (auditLogsErrors) {
```
huwshimi, line 14: I think this should be in a `useEffect` in case this component rerenders while there are audit log errors as this would cause a new toast to appear on every render (currently there shouldn't be many cases where it will rerender but this could change).

huwshimi (commented):
I think this is the right approach, just one improvement noted.

src/components/ConnectionError/ConnectionError.tsx
vladimir-cucu: Done!

src/components/ConnectionError/ConnectionError.test.tsx
```diff
@@ -38,4 +40,20 @@ describe("ConnectionError", () => {
...
   });
+
+  it("displays Audit Logs errors", async () => {
+    const state = rootStateFactory.withGeneralConfig().build({
+      juju: { auditEvents: { errors: "Oops!" } },
+    });
+    const {
+      result: { rerender },
+    } = renderComponent(<ConnectionError />, { state });
+    rerender(<Toaster />);
```
vladimir-cucu, line 51: I'm not sure how we can test this bit without a forced rerender, considering that the `Toast` is triggered inside a `useEffect`. Is there a better way to do this part?

src/components/ConnectionError/ConnectionError.test.tsx
```diff
@@ -38,4 +40,20 @@ describe("ConnectionError", () => {
...
     expect(window.location.reload).toHaveBeenCalled();
   });
+
+  it("displays Audit Logs errors", async () => {
+    const state = rootStateFactory.withGeneralConfig().build({
+      juju: { auditEvents: { errors: "Oops!" } },
+    });
+    const {
+      result: { rerender },
+    } = renderComponent(<ConnectionError />, { state });
```
huwshimi, line 50: You can avoid the rerender by rendering the Toaster initially. This is how we handle this case in other tests.

```suggestion
    renderComponent(<>
        <Toaster />
        <ConnectionError />
      </>, { state });
```

huwshimi (approved):
This looks good, just one suggestion on how to remove the rerender in the test.

src/components/ConnectionError/ConnectionError.test.tsx
vladimir-cucu: Done!

PR #1715, opened 2024-02-14
WD-8411 - Limit write actions for secrets

huwshimi:
## Done

- Limit write actions for secrets to users that are have write or admin permission and are using a version of Juju with the Secrets v2 facade.

## QA

- Log in as a model admin (your normal user probably is otherwise log in as the admin user).
- Go to the secrets tab and check that you can open the panels for adding/updating secrets etc.
- Create a new user that has write permissions to a model (you will also need to make them a controller superuser due to this bug: https://bugs.launchpad.net/juju/+bug/2053102):
```
juju add-user writeuser
juju grant writeuser superuser
juju grant writeuser write [model-name]
```
- Log in as that user.
- Go to the secrets tab and check that you can open the panels for adding/updating secrets etc.
- Create a new user that has read permissions for a model:
```
juju add-user readuser
juju grant readuser superuser
juju grant readuser read [model-name]
```
- Log in as that user.
- Go to the secrets tab and check that there is no "Add secret" button and the action column doesn't appear in the table.

## Details

- https://warthogs.atlassian.net/browse/WD-8411
- https://warthogs.atlassian.net/browse/WD-8410

### DIFF
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -63,6 +63,11 @@ import type {
   FullStatusWithAnnotations,
 } from "./types";
 
+export enum Label {
+  START_MODEL_WATCHER_NO_CONNECTION_ERROR = "Could not connect to model",
+  START_MODEL_WATCHER_NO_ID_ERROR = "Could not watch model for changes",
+}
+
 export const PING_TIME = 20000;
 export const LOGIN_TIMEOUT = 5000;
 // Juju supports a client one major version away from the controller's version,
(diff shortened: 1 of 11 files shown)


src/juju/api.ts
huwshimi, line 1: The changes in this file are to display the error that gets returned for non admin users.

src/pages/ModelDetails/ModelDetails.tsx
```diff
@@ -38,9 +38,9 @@ export default function ModelDetails() {
     async function loadFullData() {
       try {
         const response = await startModelWatcher(modelUUID, appState, dispatch);
-        conn = response?.conn ?? null;
-        watcherHandle = response?.watcherHandle ?? null;
-        pingerIntervalId = response?.pingerIntervalId ?? null;
+        conn = response.conn;
+        watcherHandle = response.watcherHandle;
+        pingerIntervalId = response.pingerIntervalId;
```
huwshimi, line 43: These are no longer optional as startModelWatcher throws in an error instead of returning null if there is an error.

vladimir-cucu (commented):
EDIT x2: I was on the wrong branch. Ooops! :) Ignore everything I wrote bellow. ~I'll delete the comment tomorrow.~ Can't delete the comment.

For a user that has write access to a model, the secrets actions are displayed as specified. However, for a user that has read access to a model, all the secrets actions are still displayed, which I assume is not the expected behavior. Maybe I've done something wrong though. I've attached the setup for the read user and the view of the secrets tab in the attached screenshot.

Screenshot from 2024-02-14 23-44-43

Edit: If I try to perform any action, I get "permission denied" error as in the attached screenshot.

Screenshot from 2024-02-14 23-49-21

vladimir-cucu (approved):
Looks good and works as expected! :rocket:

PR #1717, opened 2024-02-15
WD-8929 - owner in access count

huwshimi:
## Done

- Include the secret owner in the access count

## QA

- Go to the secrets tab.
- Check that secrets that are owned by an application include the owner in the count of applications that have been granted access.

## Details

https://warthogs.atlassian.net/browse/WD-8929


vladimir-cucu (approved):
Looks good! Thanks for the change. :+1:

PR #1718, opened 2024-02-15
WD-8405 - Config panel secret picker

huwshimi:
## Done

- Add a secret picker to the config panel.

## QA

- Go to a model and click on an application.
- Click the Configure button to open the config panel.
- Next to the string inputs you should see a little shield icon.
- Click the shield icon and pick a secret.
- The text input should get the full secret URI.
Note: at the moment the list is empty if there are no secrets in the model, but this will get updated to have an "Add secret..." button.

## Details

- https://warthogs.atlassian.net/browse/WD-8405

## Screenshots

Screenshot 2024-02-15 at 4 44 07 pm


vladimir-cucu (approved):
Looks good and works as expected! :rocket:

PR #1720, opened 2024-02-18
WD-9004 - feat: handle connection errors in WebCLI

vladimir-cucu:
## Done

- Displayed the connection errors in `WebCLI`.

## Details

- https://warthogs.atlassian.net/browse/WD-9004


src/components/WebCLI/WebCLI.tsx
```diff
@@ -132,16 +135,38 @@ const WebCLI = ({
...
-    setConnection(conn);
+    setConnection((prevConnection) => {
+      if (!prevConnection || prevConnection.isOpen()) {
+        prevConnection?.disconnect();
+      } else {
+        setPendingConnections((prevPendingConnections) => [
+          ...prevPendingConnections,
+          prevConnection,
+        ]);
+      }
```
vladimir-cucu, line 157: Another option would be to just close the previous connection without checking if it is opened. This way, we won't need the extra `pendingConnections` state and all the extra logic surrounding it. The downside is that sometimes we might close the connection before it is opened and we'll receive a warning regarding this issue in the console.

src/components/WebCLI/connection.ts
```diff
@@ -3,84 +3,119 @@ type Options = {
...
-    this._wsOnOpen = options.onopen;
-    this._wsOnClose = options.onclose;
+    this.#messageCallback = options.messageCallback;
+    this.#address = options.address;
+    this.#wsOnOpen = options.onopen;
+    this.#wsOnClose = options.onclose;
+    this.#wsOnError = options.onerror;
+  }
+
+  getAddress() {
```
vladimir-cucu, line 27: Not sure how useful this getter is, considering that we are returning the address itself without any modifications. Should we make `#address` public instead and directly access it in `components/WebCLI/WebCLI.tsx`?

vladimir-cucu: Thanks for the suggestion!

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Removed `pendingConnections` state.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -127,22 +127,53 @@ const WebCLI = ({
     const conn = new Connection({
```
huwshimi, line 127: We shouldn't create a new connection if there is an active one. This shouldn't happen, but it's probably worth adding the check just in case.

```suggestion
    // If we have an active WebSocket connection then don't create a new one.
    if (connection.current?.isActive()) {
      return;
    }
    const conn = new Connection({
```

src/components/WebCLI/WebCLI.tsx
```diff
@@ -127,22 +127,53 @@ const WebCLI = ({
...
       address: wsAddress,
       onopen: () => {
-        // Unused handler.
+        // Close this websocket connection if it was in CONNECTING state
+        // while a new websocket connection was established.
+        if (conn.websocket !== connection.current?.websocket) {
+          conn.disconnect();
+        }
```
huwshimi, line 134: This can be removed as we're not creating a new connection if there is an active one, and in strict dev mode the socket will get closed by the cleanup useEffect.

```suggestion
        // Unused handler.
```

src/components/WebCLI/WebCLI.tsx
```diff
@@ -127,22 +127,53 @@ const WebCLI = ({
...
       },
       onclose: () => {
         // Unused handler.
       },
+      onerror: (error) => {
+        setInlineError(
+          InlineErrors.CONNECTION,
+          typeof error === "string" ? error : "Unknown error.",
+        );
```
huwshimi, line 143: To prevent the error from the old websocket closing while connecting we can check that the error is related to the current connection.

```suggestion
        // Only display errors if they're related to the current websocket connection.
        if (conn.websocket === connection.current?.websocket) {
          setInlineError(
            InlineErrors.CONNECTION,
            typeof error === "string" ? error : "Unknown error.",
          );
        }
```

src/components/WebCLI/WebCLI.tsx
```diff
@@ -127,22 +127,53 @@ const WebCLI = ({
...
+      // issue, in development environment we only close the WebSocket
+      // connection if it is in OPEN state. In production environment, as the
+      // cleanup runs only once, we just close the last connection, even if it
+      // is in OPEN state.
+      if (
+        process.env.NODE_ENV !== "development" ||
+        connection.current?.isOpen()
+      ) {
+        connection.current?.disconnect();
+      }
```
huwshimi, line 172: Close the connection even if it is pending on unmount (this will also close the first websocket in strict dev mode, but that's OK as a new one will get created).

```suggestion
      if (connection.current?.isActive()) {
        connection.current?.disconnect();
      }
```

src/components/WebCLI/connection.ts
```diff
@@ -3,84 +3,128 @@ type Options = {
...
+    }
+    if (this.#timeout) {
+      clearTimeout(this.#timeout);
     }
   }
 
-  _handleMessage(e: MessageEvent) {
+  isOpen() {
+    return this.#ws?.readyState === WebSocket.OPEN;
+  }
```
huwshimi, line 63: Also add the active check.

```suggestion
  isOpen() {
    return this.#ws?.readyState === WebSocket.OPEN;
  }
  
  isActive() {
    return this.isOpen() || this.#ws?.readyState === WebSocket.CONNECTING;
  }
```

huwshimi (commented):
It looks like this still has the same issue where if the WebSocket is still connecting when you navigate away from the model details then the connection doesn't close (you can reproduce this by going to the model list, clicking on a model than using your browser to go back to the list then you can use your keyboard shortcuts to go forward and back very quickly which will exit the model details while it is making the connection).

As I understand it, calling `ws.close()` is the right way to close a pending connection (from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close): "The WebSocket.close() method closes the WebSocket connection or **connection attempt**"). This results in the warning "WebSocket is closed before the connection is established." which I think is safe to ignore.

The problem is that it calls `onerror` when it closes while pending (this was the issue I had in my previous attempt), however I think we can get around this by only handling errors if they're from the current connection.

I've left comments about where I think it needs to change, but you can also see my complete version [here](https://github.com/canonical/juju-dashboard/compare/main...huwshimi:webcli-onerror-wip2).

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Thanks for the review! This works. I've changed the code slightly so that I don't expose the `wesocket` entirely, but just compare if 2 `Connections` are equal with a method within the class.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: I moved the check for `isActive` inside the `disconnect` method of `Connection` class. I was thinking that this approach would be less error prone.

huwshimi (commented):
This is working nicely now.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -20,6 +20,8 @@ import "./_webcli.scss";
 export enum Label {
   CONNECTION_ERROR = "Unable to connect to the model.",
   AUTHENTICATION_ERROR = "Unable to authenticate.",
+  NOT_OPEN_ERROR = "WebSocket connection is not open.",
```
huwshimi, line 23: Maybe this should say something like "Lost connection to model" or something similar.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -132,17 +138,32 @@ const WebCLI = ({
...
       messageCallback: (message: string) => {
         wsMessageStore.current = wsMessageStore.current + message;
         setOutput(wsMessageStore.current);
       },
     }).connect();
-    setConnection(conn);
-    return () => {
-      conn.disconnect();
-    };
+    connection.current?.disconnect();
```
huwshimi, line 156: Do we need this disconnect here? It checks that there isn't an active connection at the top of this useEffect.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -192,12 +213,16 @@ const WebCLI = ({
...
+      connection.current?.send(
+        JSON.stringify({
+          ...authentication,
+          commands: [command],
+        }),
+      );
+    } else {
+      setInlineError(InlineErrors.CONNECTION, Label.NOT_OPEN_ERROR);
+    }
     sendAnalytics({
```
huwshimi, line 226: Maybe this should be inside the `if (connection.current?.isOpen()) {` above?

huwshimi (approved):
Just a couple of small notes.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: You're right! Thanks for spotting this. Removed it.

src/components/WebCLI/WebCLI.tsx
vladimir-cucu: Oops :) Moved it inside the `if`. Thanks!

PR #1721, opened 2024-02-19
WD-8407 - config grant access

huwshimi:
## Done

- Display a confirmation to grant app access to secrets.

## QA

- Open the config panel for an app.
- Choose a secret for a text field.
- Save the form.
- After the confirmation you should be shown a confirmation for whether to grant access.
- Click Yes and the app should get granted.

## Details

https://warthogs.atlassian.net/browse/WD-8407

### DIFF
--- a/src/panels/ConfigPanel/ConfigPanel.tsx
+++ b/src/panels/ConfigPanel/ConfigPanel.tsx
@@ -1,3 +1,4 @@
+import type { ListSecretResult } from "@canonical/jujulib/dist/api/facades/secrets/SecretsV2";
 import {
   Button,
   ConfirmationModal,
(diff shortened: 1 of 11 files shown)


src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -57,14 +70,49 @@ enum InlineErrors {
...
+        (entry.type === "string" && value.startsWith("secret:")))
+    ) {
+      uris.push(value);
+    }
+    return uris;
+  }, []);
+  return secrets
+    ? secretURIs?.filter((secretURI) => {
+        const secret = secrets.find(({ uri }) => uri === secretURI);
+        const access = secret?.access?.find(
```
vladimir-cucu, line 106: I've observed that a secret that has an owner app doesn't have the `access` key (see attached screenshot).

Screenshot from 2024-02-19 21-15-39

Because of this, we will add this secret to the `getRequiredGrants(...)` list. Also, we might add the same secret multiple times, as the same secret might come from different config options (see attached screenshot).

Screenshot from 2024-02-19 21-18-51

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -277,6 +350,67 @@ export default function ConfigPanel(): JSX.Element {
...
+              confirmButtonLabel={Label.GRANT_CONFIRM_BUTTON}
+              confirmButtonAppearance="positive"
+              onConfirm={() => {
+                setConfirmType(null);
+                // Clear the form errors if there were any from a previous submit.
+                setInlineError(InlineErrors.FORM, null);
+                if (!appName || !requiredGrants) {
+                  // It is not possible to get to this point if these
+                  // variables aren't set.
+                  return;
```
vladimir-cucu, line 373: This part was not tested, but I am not sure if it is worth adding a test here anyway. What do you think?

huwshimi: There's no way to actually get to this code (there's no way to display this confirmation if there isn't an app name or required grants), it just needs this return to satisfy typescript.

src/panels/ConfigPanel/ConfigPanel.tsx
huwshimi: > I've observed that a secret that has an owner app doesn't have the `access` key (see attached screenshot).
> Because of this, we will add this secret to the `getRequiredGrants(...)` list, even if it has been granted permission to beforehand.

We shouldn't grant to app-owned secrets (this will be fixed in the API) so I've filtered them out when checking for required grants. I'll do another PR to remove them from the picker.


> Also, we might add the same secret multiple times to this list, as the same secret might come from different config options (see attached screenshot). I think it might make sense to remove duplicates here.

Thanks, I've deduped the apps when granting.

src/panels/ConfigPanel/ConfigPanel.tsx
vladimir-cucu: Oh, sorry, didn't read the comment that was literally there. :) Thanks for the explanation!

vladimir-cucu (approved):

PR #1726, opened 2024-02-21
WD-9028 - Dedupe API hooks

huwshimi:
## Done

- Combine common api hook handlers.

## QA

- Go to the secrets tab and check the secrets load.
- Check that you can perform write actions on the secrets.

## Details

- https://warthogs.atlassian.net/browse/WD-9028

### DIFF
--- a/src/utils/toErrorString.ts
+++ b/src/utils/toErrorString.ts
@@ -1,3 +1,7 @@
+export enum Label {
+  UNKNOWN_ERROR = "Unknown error",
+}
+
 const toErrorString = (error: unknown) => {
   if (error instanceof Error) {
     return error.message;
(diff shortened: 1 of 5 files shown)


src/utils/toErrorString.ts
```diff
@@ -0,0 +1,11 @@
+const toErrorString = (error: unknown) => {
+  if (error instanceof Error) {
+    return error.message;
+  }
+  if (typeof error === "string") {
+    return error;
+  }
+  return "Unknown error";
```
vladimir-cucu, line 8: Was wondering if it might be a slight improvement if we make an `enum` out of this? This way, if we change this error naming in the future, we won't need to manually change the error naming in all the respective test files where it is used, as we will just use the exported `enum` instead.

vladimir-cucu (approved):
Looks good to me! :rocket:

src/utils/toErrorString.ts
huwshimi: Yes, good idea. Done!

PR #1729, opened 2024-02-22
Improve timeout promise

huwshimi:
## Done

- Reject on timeout instead of returning special string.

## QA

- Go to the model list and check that the models appear.
- Open api.ts and change `LOGIN_TIMEOUT` to a very low number.
- Reload the model list and check that the console has timeout errors.


vladimir-cucu (approved):
LGTM! :rocket:

PR #1730, opened 2024-02-23
WD-9100 - action API hooks

huwshimi:
## Done

- Turn action calls into hooks.

## QA

- Go to the action log tab for a model and check that the logs load.
- Open the action panels and check that you can submit actions.

## Details

https://warthogs.atlassian.net/browse/WD-9100


vladimir-cucu (approved):
LGTM! :+1:

PR #1731, opened 2024-02-23
WD-9101 - application API hooks

huwshimi:
## Done

- Move application API calls to hooks.

## QA

- Open the config panel.
- Make a change and save.
- Open the panel again and check that the value you changed was saved.

## Details

- https://warthogs.atlassian.net/browse/WD-9101


vladimir-cucu (approved):
LGTM! :rocket:

PR #1738, opened 2024-03-05
WD-9417 - feat: update error handling after jujulib update to v6

vladimir-cucu:
## Done

- Update error handling after `jujulib` update to v6.

## Details

- https://warthogs.atlassian.net/browse/WD-9417

### DIFF
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -252,8 +252,7 @@ export async function fetchModelStatus(
           status =
             (await conn?.facades.client?.fullStatus({ patterns: [] })) ?? null;
           if (!status) {
-            // Placeholder error to be caught in the subsequent catch.
-            throw new Error();
+            throw new Error("Status not returned.");
           }
         } catch (error) {
           // XXX If there is an error fetching the full status it's likely that
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -477,34 +477,26 @@ export async function fetchControllerList(
   }
 
   if (controllers) {
-    try {
-      // check for updates
-      await Promise.all(
-        controllers.map(async (controller) => {
-          let updateAvailable = false;
-          try {
-            updateAvailable =
-              (await jujuUpdateAvailable(
-                "agent-version" in controller
-                  ? controller["agent-version"]
-                  : controller.version || "",
-              )) ?? false;
-          } catch (error) {
-            updateAvailable = false;
-          }
-          controller.updateAvailable = updateAvailable;
-        }),
-      );
-      dispatch(
-        jujuActions.updateControllerList({ wsControllerURL, controllers }),
-      );
-    } catch (error) {
-      dispatch(
-        generalActions.storeConnectionError(
-          "Unable to update the list of controllers.",
-        ),
-      );
-    }
+    // check for updates
+    await Promise.all(
+      controllers.map(async (controller) => {
+        let updateAvailable = false;
+        try {
+          updateAvailable =
+            (await jujuUpdateAvailable(
+              "agent-version" in controller
+                ? controller["agent-version"]
+                : controller.version || "",
+            )) ?? false;
+        } catch (error) {
+          updateAvailable = false;
+        }
+        controller.updateAvailable = updateAvailable;
+      }),
+    );
+    dispatch(
+      jujuActions.updateControllerList({ wsControllerURL, controllers }),
+    );
   }
 }
 
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -255,11 +255,6 @@ export async function fetchModelStatus(
             throw new Error("Status not returned.");
           }
         } catch (error) {
-          // XXX If there is an error fetching the full status it's likely that
-          // Juju can no longer access this model. At this moment we don't have
-          // a location to notify the user. In the new watcher model that's
-          // being implemented we will be able to surface this error in the
-          // model details page.
           throw new Error(
             `Unable to fetch the status. ${toErrorString(error)}`,
           );
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -42,11 +42,12 @@ import type {
 } from "store/juju/types";
 import { ModelsError } from "store/middleware/model-poller";
 import type { RootState, Store } from "store/store";
+import { toErrorString } from "utils";
 
 import { getModelByUUID } from "../store/juju/selectors";
 
 import type { AuditEvents, FindAuditEventsRequest } from "./jimm/JIMMV3";
-import type { CrossModelQueryFullResponse } from "./jimm/JIMMV4";
+import type { CrossModelQueryResponse } from "./jimm/JIMMV4";
 import type {
   AllWatcherDelta,
   ApplicationInfo,
@@ -233,7 +234,7 @@ export async function fetchModelStatus(
     useIdentityProvider = config?.identityProviderAvailable ?? false;
   }
   const modelURL = wsControllerURL.replace("/api", `/model/${modelUUID}/api`);
-  let status: FullStatusWithAnnotations | string | null = null;
+  let status: FullStatusWithAnnotations | null = null;
   let features: ModelFeatures | null = null;
   // Logged in state is checked multiple times as the user may have logged out
   // between requests.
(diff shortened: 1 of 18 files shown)


src/juju/api.ts
```diff
@@ -247,15 +248,22 @@ export async function fetchModelStatus(
...
       if (isLoggedIn(getState(), wsControllerURL)) {
-        status =
-          (await conn?.facades.client?.fullStatus({ patterns: [] })) ?? null;
-        if (!status || typeof status === "string") {
+        try {
+          status =
+            (await conn?.facades.client?.fullStatus({ patterns: [] })) ?? null;
+          if (!status) {
+            // Placeholder error to be caught in the subsequent catch.
+            throw new Error();
```
huwshimi, line 256: This could be:

```suggestion
            throw new Error("Status not returned.");
```

vladimir-cucu: Makes sense to add this. Done! Also removed the comment on line 255 as it becomes redundant.

src/panels/RemoveSecretPanel/RemoveSecretPanel.tsx
```diff
@@ -64,13 +65,8 @@ const RemoveSecretPanel = () => {
...
       ])
-        .then((response: ErrorResults | string) => {
-          let error;
-          if (typeof response === "string") {
-            error = response;
-          } else if (Array.isArray(response.results)) {
-            error = response.results[0]?.error?.message;
-          }
+        .then((response: ErrorResults) => {
+          const error = response.results[0]?.error?.message;
```
vladimir-cucu, line 69: I was wondering why here we check the first element in the array, however, in `GrantSecretPanel` we check the first instance of an element in the array that has `error` intead?

huwshimi: Ah that's a bug in `GrantSecretPanel`. Here we only care about the error in the first element of the returned array, as we're only passing one secret to be removed. In `GrantSecretPanel` we should display all the errors returned by the API as each application that is being granted can have an error. I'll fix that.

src/store/middleware/model-poller.ts
```diff
@@ -285,17 +286,30 @@ export const modelPollerMiddleware: Middleware<
...
+          reduxStore.dispatch,
+        );
+        return response;
+      } catch (error) {
+        console.error(
+          `Could not grant sharing permissions to model ${payload.modelUUID}`,
+          error,
+        );
+        reduxStore.dispatch(
+          jujuActions.updateModelsError({
```
huwshimi, line 307: Instead of handling the error inside the middleware and storing it in the model errors I think we should allow this to throw which will return the error to `ShareModel` which has as try catch around the `updatePermissions` dispatch and displays the error in a toast. I think you can remove your changes and it will handle the error correctly.

src/juju/api.ts
```diff
@@ -247,15 +248,21 @@ export async function fetchModelStatus(
...
+            (await conn?.facades.client?.fullStatus({ patterns: [] })) ?? null;
+          if (!status) {
+            throw new Error("Status not returned.");
+          }
+        } catch (error) {
           // XXX If there is an error fetching the full status it's likely that
           // Juju can no longer access this model. At this moment we don't have
           // a location to notify the user. In the new watcher model that's
           // being implemented we will be able to surface this error in the
           // model details page.
```
huwshimi, line 262: We can remove this comment now.

src/juju/jimm/JIMMV4.ts
```diff
@@ -36,14 +32,12 @@ export type RelationshipTuple = {
...
-export type CheckRelationResponse =
-  | {
-      allowed: boolean;
-    }
-  | string;
+export type CheckRelationResponse = {
+  allowed: boolean;
+};
 
 export const isCrossModelQueryResponse = (
```
huwshimi, line 39: Now that we don't get a string response we can remove this narrowing function (I'll leave a comment in the slice for what we can do instead).

src/panels/GrantSecretPanel/GrantSecretPanel.tsx
```diff
@@ -37,14 +38,10 @@ export type FormFields = {
...
-  } else if (Array.isArray(response.results)) {
-    const error = response.results.find(({ error }) => !!error);
-    if (error?.error?.message) {
-      throw new Error(error.error?.message);
-    }
+const handleErrors = (response: ErrorResults) => {
+  const error = response.results.find(({ error }) => !!error);
+  if (error?.error?.message) {
+    throw new Error(error.error?.message);
   }
```
huwshimi, line 45: Instead of me doing another PR, maybe you could fix this here with: 

```suggestion
  const errors = response.results.reduce((errorString, { error }) => {
    if (error?.message) {
      errorString = [errorString, error.message].filter(Boolean).join(". ");
    }
    return errorString;
  }, "");
  if (errors) {
    throw new Error(errors);
  }
```

src/panels/RemoveSecretPanel/RemoveSecretPanel.tsx
huwshimi: I've left a fix comment above to handle this case instead of me doing another PR.

src/store/juju/slice.ts
```diff
@@ -243,18 +243,16 @@ const slice = createSlice({
     },
     updateCrossModelQuery: (
```
huwshimi, line 244: We should remove this error handling from this action/reducer and instead add a `storeCrossModelQueryError` like we do for other fetch/store/error handlers.

src/store/middleware/model-poller.ts
```diff
@@ -335,13 +356,15 @@ export const modelPollerMiddleware: Middleware<
       if (!conn) {
         return;
       }
-      let crossModelQueryResponse: CrossModelQueryFullResponse;
+      let crossModelQueryResponse: CrossModelQueryResponse | Error;
       try {
         crossModelQueryResponse = await crossModelQuery(conn, query);
```
huwshimi, line 361: If successful this should store the response.

src/store/middleware/model-poller.ts
```diff
@@ -335,13 +356,15 @@ export const modelPollerMiddleware: Middleware<
...
       } catch (error) {
         console.error("Could not perform cross model query:", error);
         crossModelQueryResponse =
```
huwshimi, line 364: Instead of passing this to `updateCrossModelQuery` we should instead add a `storeCrossModelQueryError` action/reducer and dispatch that here.

src/store/juju/slice.ts
```diff
@@ -243,18 +243,16 @@ const slice = createSlice({
...
       state.crossModelQuery.results = isCrossModelQueryResponse(payload)
         ? payload.results
         : null;
       state.crossModelQuery.errors =
         isCrossModelQueryResponse(payload) && Object.keys(payload.errors).length
           ? payload.errors
-          : typeof payload === "string"
-            ? payload
+          : payload instanceof Error
+            ? payload.message
```
huwshimi, line 255:
```suggestion
          : payload instanceof Error
            ? payload.message
```
      const hasErrors = Object.keys(payload.errors).length > 0;
      state.crossModelQuery.results = hasErrors ? null : payload.results;
      state.crossModelQuery.errors = hasErrors ? payload.errors : null;

src/panels/RemoveSecretPanel/RemoveSecretPanel.tsx
vladimir-cucu: Thanks for the review! Fixed it.

src/store/middleware/model-poller.ts
```diff
@@ -335,17 +342,21 @@ export const modelPollerMiddleware: Middleware<
       if (!conn) {
         return;
       }
-      let crossModelQueryResponse: CrossModelQueryFullResponse;
       try {
-        crossModelQueryResponse = await crossModelQuery(conn, query);
+        const crossModelQueryResponse = await crossModelQuery(conn, query);
+        reduxStore.dispatch(
+          jujuActions.updateCrossModelQuery(crossModelQueryResponse),
+        );
```
huwshimi, line 349: It looks like my comment from yesterday was:

> If it has crossModelQueryResponse.errors then dispatch storeCrossModelQueryError with the array of errors.

So we should add that here and then we can remove the `hasErrors` check inside `updateCrossModelQuery` in the slice file.

src/store/juju/slice.ts
```diff
@@ -243,19 +242,20 @@ const slice = createSlice({
...
-        : null;
-      state.crossModelQuery.errors =
-        isCrossModelQueryResponse(payload) && Object.keys(payload.errors).length
-          ? payload.errors
-          : typeof payload === "string"
-            ? payload
-            : null;
+      const hasErrors = !!Object.keys(payload.errors).length;
+      state.crossModelQuery.results = hasErrors ? null : payload.results;
+      state.crossModelQuery.errors = hasErrors ? payload.errors : null;
```
huwshimi, line 249:
```suggestion
      state.crossModelQuery.results = payload.results;
      state.crossModelQuery.errors = null;
```

See my other comment in model-poller.ts.

huwshimi (approved):
Just one small fix! Feel free to land this after that change.

src/store/middleware/model-poller.ts
vladimir-cucu: Done!

vladimir-cucu: I changed the action name of `updateCrossModelQuery` to `updateCrossModelQueryResults` in order to be more consistent with the naming of `updateCrossModelQueryErrors`. I think this would make it less error prone as well, considering that we can't be in a situation where we dispatch `updateCrossModelQuery` that contains `errors`, as we dispatch the `errors` and `results` separately now.

PR #1756, opened 2024-05-21
WD-11355 - fix: remove overrides that have been fixed in Vanilla

huwshimi:
## Done

- remove overrides that have been fixed in Vanilla

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Check nothing looks broken (particularly the tabs in model details).

## Details

https://warthogs.atlassian.net/browse/WD-11355

### DIFF
--- a/src/components/ModelTableList/ModelSummary/ModelSummary.tsx
+++ b/src/components/ModelTableList/ModelSummary/ModelSummary.tsx
@@ -28,7 +28,7 @@ const ModelSummary = ({ modelData, ownerTag }: Props) => {
       <Tooltip
         message="See applications"
         position="top-center"
-        className="u-flex--block has-icon"
+        className="u-flex--block u-has-icon"
       >
         <ModelDetailsLink
           modelName={modelData.model.name}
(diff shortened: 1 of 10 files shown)


src/components/ModelTableList/ModelSummary/ModelSummary.tsx
```diff
@@ -28,7 +28,7 @@ const ModelSummary = ({ modelData, ownerTag }: Props) => {
       <Tooltip
         message="See applications"
         position="top-center"
-        className="u-flex--block has-icon"
+        className="u-flex--block u-has-icon"
```
vladimir-cucu, line 31: I think there might be some other places where the same change should be performed e.g. `SearchHistoryMenu`, `SecretsTable`, `SecrtetsPicker`, `ShareModel`.

vladimir-cucu (approved):
Looks good to me locally! :+1:

src/components/ModelTableList/ModelSummary/ModelSummary.tsx
huwshimi: In those other places the `has-icon` is applied to a button, where using `has-icon` is still the right class name (https://vanillaframework.io/docs/patterns/buttons#icon) whereas `u-has-icon` is for non-button elements (https://github.com/canonical/vanilla-framework/issues/3102).

PR #1757, opened 2024-05-27
WD-11520 - refactor: migrate to Vite and Vitest

huwshimi:
## Done

- Migrate from CRA to Vite.
  - CRA is not currently receiving updates which is causing issues with some updates e.g. doesn't support new versions of eslint.
  - Vite gives us a proxy which we require for OIDC.
- Migrate from Jest to Vitest.
- Update eslint config to support Vite (and fixed a bunch of issues that were revealed).

## QA

- Check that the UI works as before.

## Details

https://warthogs.atlassian.net/browse/WD-11520


vladimir-cucu (approved):
All tests pass locally. Also, it looks like all instances using `jest` have been modified to use `vi`. :+1:

PR #1761, opened 2024-05-29
WD-11676 - refactor: split exports to enable fast refresh

huwshimi:
## Done

- move exports to separate files to enable fast refresh of components
- add index.ts files where required

## Details

https://warthogs.atlassian.net/browse/WD-11676


src/components/Banner/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from "./Banner";
+export { Label as BannerLabel } from "./types";
```
vladimir-cucu, line 2: Was wondering if maybe excluding all `index.ts` files from test coverage would make sense in order to not get the Codecov false positive?

vladimir-cucu (approved):
Looks good! Tested it locally to check for weird behaviors and it works as expected. :rocket:

src/components/Banner/index.ts
huwshimi: The report actually just revealed a bunch of places I still need to move exports (not sure why the linter didn't catch them). I'll push a change shortly.

vladimir-cucu: The changes look good! Thanks.

PR #1762, opened 2024-05-29
WD-11588 - refactor: Remove nested components in components/Login

vladimir-cucu:
## Done

- Removed nested components in `components/Login`.

## Details

- https://warthogs.atlassian.net/browse/WD-11588

### DIFF
--- a/src/components/LogIn/LogIn.test.tsx
+++ b/src/components/LogIn/LogIn.test.tsx
@@ -2,9 +2,6 @@ import { screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { vi } from "vitest";
 
-import { thunks as appThunks } from "store/app";
-import { actions as generalActions } from "store/general";
-import * as dashboardStore from "store/store";
 import { configFactory, generalStateFactory } from "testing/factories/general";
 import { rootStateFactory } from "testing/factories/root";
 import { renderComponent } from "testing/utils";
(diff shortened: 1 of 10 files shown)


src/components/LogIn/LogIn.test.tsx
```diff
@@ -58,38 +31,6 @@ describe("LogIn", () => {
...
-      general: generalStateFactory.withConfig().build({
-        config: configFactory.build({
-          identityProviderAvailable: false,
-        }),
-      }),
-    });
-    renderComponent(<LogIn>App content</LogIn>, { state });
-    expect(screen.getByRole("button")).toHaveTextContent(
-      "Log in to the dashboard",
-    );
```
huwshimi, line 92: We still need tests in this file about which form is displayed as the logic for choosing the form based on the state of `identityProviderAvailable` is in the Login component.

src/components/LogIn/UserPassForm/UserPassForm.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const mockUseAppDispatch = vi.fn().mockReturnValue({
+      then: vi.fn().mockReturnValue({ catch: vi.fn() }),
+    });
+    vi.spyOn(dashboardStore, "useAppDispatch").mockReturnValue(
+      mockUseAppDispatch,
+    );
+    const state = rootStateFactory.build({
+      general: generalStateFactory.withConfig().build({
+        config: configFactory.build({
+          identityProviderAvailable: false,
```
huwshimi, line 41: I don't think this is required in this test either.

src/components/LogIn/index.ts
```diff
@@ -1,2 +1 @@
 export { default } from "./LogIn";
-export { Label as LogInLabel } from "./types";
```
huwshimi, line 2: I added these exports in my last PR so that we can always import the label by the full name to avoid conflicts instead of having to do `import { Label as LogInLabel } ...`. Was there a reason you removed this?

src/components/LogIn/UserPassForm/UserPassForm.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  useEffect(() => {
+    focus.current?.focus();
+  }, []);
+
+  return (
+    <form onSubmit={handleSubmit}>
+      <label htmlFor="username">Username</label>
+      <input type="text" name="username" id="username" ref={focus} />
```
huwshimi, line 51: There's currently a bug where the username doesn't always get focus, so as a tiny fix you could replace `ref={focus}` with `autoFocus` and remove the useRef and useEffect above.

huwshimi (commented):
I've left a few comments here.

huwshimi (approved):
Thanks for those fixes!

PR #1765, opened 2024-06-03
WD-11589 - refactor: remove nested components in StatusGroup

vladimir-cucu:
## Done

- Remove nested components in `StatusGroup`.

## Details

- https://warthogs.atlassian.net/browse/WD-11589

### DIFF
--- a/src/components/ModelTableList/StatusGroup/StatusGroup.test.tsx
+++ b/src/components/ModelTableList/StatusGroup/StatusGroup.test.tsx
@@ -1,7 +1,4 @@
-import { act, screen, within } from "@testing-library/react";
-import type { UserEvent } from "@testing-library/user-event";
-import userEvent from "@testing-library/user-event";
-import { vi } from "vitest";
+import { screen, within } from "@testing-library/react";
 
 import type { RootState } from "store/store";
 import { rootStateFactory } from "testing/factories";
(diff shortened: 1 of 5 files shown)


src/components/ModelTableList/StatusGroup/StatusGroup.test.tsx
```diff
@@ -177,50 +164,19 @@ describe("StatusGroup", () => {
     );
   });
 
-  it("displays links to blocked apps and units", async () => {
-    vi.useFakeTimers();
+  it("displays links to blocked apps", async () => {
```
vladimir-cucu, line 167: I think we still need to test that these links get rendered, however, I've removed the bit regarding the Tooltip testing, as the tests for it are present in `WarningMessage.test.tsx`.

src/components/ModelTableList/StatusGroup/WarningMessage/WarningMessage.tsx
```diff
@@ -0,0 +1,69 @@
...
+import urls from "urls";
+import { getUserName } from "utils";
+
+type Props = {
+  model: ModelData;
+};
+
+/**
+  Warning message for the model name cell.
+  @param model The full model data.
```
huwshimi, line 17: This component doesn't have a `model` param anymore (the `model` arg is part of the props object). If you would like this comment to appear as a hint when using the component you could move this comment to the props type.

```typescript
type Props = {
  /**
   * The full model data.
   */
  model: ModelData;
};
```

huwshimi (approved):
Just one tiny suggestion about the doc string.

src/components/ModelTableList/StatusGroup/WarningMessage/WarningMessage.tsx
vladimir-cucu: Fixed! Thanks for the review.

PR #1766, opened 2024-06-04
WD-11590 - refactor: remove nested components in CodeSnippetBlock

vladimir-cucu:
## Done

- Remove nested components in `pages/CodeSnippetBlock

## Details

- https://warthogs.atlassian.net/browse/WD-11590

### DIFF
--- a/src/pages/AdvancedSearch/CodeSnippetBlock/CodeSnippetBlock.tsx
+++ b/src/pages/AdvancedSearch/CodeSnippetBlock/CodeSnippetBlock.tsx
@@ -7,24 +7,19 @@ import { Highlight } from "prism-react-renderer";
 import Prism from "prismjs/components/prism-core";
 import { useCallback, useState } from "react";
 import type { KeyPath } from "react-json-tree";
-import {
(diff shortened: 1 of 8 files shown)


huwshimi:
@vladimir-cucu (not sure if this PR is ready for a review yet, if not you can ignore this comment). It looks like most of the logic in Label is only tested by the parent component. I think either the tests should be moved over, or if you think some of them are necessary then it'd be worth making sure the logic is also tested in the Label so that if the parent changes we don't lose those tests.

vladimir-cucu:
> @vladimir-cucu (not sure if this PR is ready for a review yet, if not you can ignore this comment). It looks like most of the logic in Label is only tested by the parent component. I think either the tests should be moved over, or if you think some of them are necessary then it'd be worth making sure the logic is also tested in the Label so that if the parent changes we don't lose those tests.

I thought of leaving the tests as they are in the parent component and add additional tests for `Label`. The PR is not yet ready for review, as most of the tests are missing. Sorry for not specifying that.

huwshimi (approved):

PR #1769, opened 2024-06-07
WD-11666 - refactor: remove nested components in ConfigPanel

vladimir-cucu:
## Done

- Removed nested components in `panels/ConfigPanel`.

## Details

- https://warthogs.atlassian.net/browse/WD-11666

### DIFF
--- a/src/panels/ConfigPanel/ConfirmationDialog/ConfirmationDialog.tsx
+++ b/src/panels/ConfigPanel/ConfirmationDialog/ConfirmationDialog.tsx
@@ -1,0 +1,208 @@
+import { ConfirmationModal } from "@canonical/react-components";
+import { useParams } from "react-router-dom";
+import usePortal from "react-useportal";
+
+import type { EntityDetailsRoute } from "components/Routes";
+import SecretLabel from "components/secrets/SecretLabel";
+import useAnalytics from "hooks/useAnalytics";
+import useCanManageSecrets from "hooks/useCanManageSecrets";
+import { type SetError } from "hooks/useInlineErrors";
+import { useGrantSecret, useSetApplicationConfig } from "juju/api-hooks";
+import type { usePanelQueryParams } from "panels/hooks";
+import { ConfirmType as DefaultConfirmType } from "panels/types";
+import { getModelSecrets } from "store/juju/selectors";
+import { useAppSelector } from "store/store";
+
+import ChangedKeyValues from "../ChangedKeyValues";
+import type { Config, ConfigQueryParams, ConfirmTypes } from "../types";
+import { ConfigConfirmType, InlineErrors, Label } from "../types";
+import { getRequiredGrants } from "../utils";
+
+type Props = {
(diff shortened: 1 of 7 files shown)


src/panels/ConfigPanel/types.ts
```diff
@@ -44,3 +46,22 @@ export enum Label {
 export enum TestId {
   PANEL = "config-panel",
 }
+
+export enum InlineErrors {
+  FORM = "form",
+  GET_CONFIG = "get-config",
+  SUBMIT_TO_JUJU = "submit-to-juju",
```
huwshimi, line 53: `FORM` and `SUBMIT_TO_JUJU` are only used inside ConfirmationDialog so they should be moved to a types file for that component.

It looks like a lot of of the values in the `Label` can also be moved to the new component.

huwshimi (commented):
This adds new tests for the confirmation dialog component, but does not remove/modify any from ConfigPanel even though the logic has moved into the new component.

I think it would be worth taking a look at what is occurring in each component and making sure the tests are relevant to the logic in that "unit".

For example: in the config panel we might want to make sure that a particular confirmation is displayed under certain conditions, and in the confirmation dialogue we'd check the API calls that it makes when confirmed.

src/panels/ConfigPanel/types.ts
vladimir-cucu: Nice catch! Everything that was no longer used in `ConfigPanel` was moved to `ConfirmationDialog`. Thanks for the review.

src/panels/ConfigPanel/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const response = await setApplicationConfig(appName, config);
+    const errors = response?.results?.reduce<string[]>((collection, result) => {
+      if (result.error) {
+        collection.push(result.error.message);
+      }
+      return collection;
+    }, []);
+    setSavingConfig(false);
+    setEnableSave(false);
+    setConfirmType(null);
```
vladimir-cucu, line 67: This call might be redundant, as we already set the config type to `null` on line 106.

src/panels/ConfigPanel/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,203 @@
...
+              // It is not possible to get to this point if these
+              // variables aren't set.
+              return;
+            }
+            void (async () => {
+              try {
+                for (const secretURI of requiredGrants) {
+                  await grantSecret(secretURI, [appName]);
+                }
+                setConfirmType(null);
```
vladimir-cucu, line 150: This call might be redundant as well, as we set the confirm type to `null` on line 137.

huwshimi (approved):
Just a couple of small changes

huwshimi (approved):
That fix looks good.

PR #1773, opened 2024-06-12
WD-12072 - docs: candid docs

huwshimi:
## Done

- add instructions for setting up Juju with Candid

## QA

- Follow the instructions in the doc and check that you can log in to the dashboard.

## Details

https://warthogs.atlassian.net/browse/WD-12072

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -34,6 +34,7 @@ contribute and what kinds of contributions are welcome.
     - [Vanilla React Components](#vanilla-react-components)
 - [Juju controllers in Multipass](#juju-controllers-in-multipass)
   - [Juju controller](#juju-controller)
+  - [Juju controller with Candid](#juju-controller-with-candid)
   - [JIMM controller](#jimm-controller)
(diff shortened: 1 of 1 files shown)


Lukewh (approved):
:+1: LGTM

HACKING.md
```diff
@@ -407,6 +408,90 @@ multipass delete juju
...
+`config.local.js`:
+
+```shell
+controllerAPIEndpoint: "wss://[container.ip]:17070/api",
+```
+
+You may also need to configure your dashboard to work with a local controller:
+
+```shell
+isJuju: true,
```
vladimir-cucu, line 490: Maybe we could also add the fact that the following should be set as well `identityProviderAvailable: true`.

vladimir-cucu (approved):
LGTM! QA'ed all the steps and it works as expected. :rocket:

HACKING.md
huwshimi: No need anymore! I removed that config option in https://github.com/canonical/juju-dashboard/pull/1768.

PR #1774, opened 2024-06-12
fix: dotrun workflow task

Lukewh:
## Done

- Updated the dotrun action to use the python version and prevent it always failing

## QA

- The dotrun job should work for this PR


vladimir-cucu (approved):
LGTM! :rocket:

PR #1789, opened 2024-07-16
WD-12713 - feat: handle oidc refresh and revoke

huwshimi:
## Done

- Poll the whoami endpoint to handle refreshing and revoking tokens.
- Log out if the user is no longer authenticated.

## QA

- Set up an OIDC enabled JIMM with the session max age to something short e.g. 30 seconds by changing `JIMM_SESSION_COOKIE_MAX_AGE: 30` in `docker-compose.yaml` before you set up the env.
- In the dashboard open `src/consts.ts`.
- Set the poll interval to less than the max age above e.g. `export const OIDC_POLL_INTERVAL = 20 * 1000;`.
- Run the dashboard.
- Log in.
- Using the network tab you should see that the whoami endpoint is pinged at the interval you set above.
- Inside the container running JIMM run the following commands to revoke the token (but don't stop the running environment or your dashboard you want this to keep pinging the API):
```
docker exec -it postgres bash
psql -U jimm -d jimm
DELETE FROM http_sessions;
```
- Next time the dashboard polls it should return you to the login screen with a message about being logged out.

## Details

https://warthogs.atlassian.net/browse/WD-12713

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -63,6 +63,7 @@
     "default-case": 0,
     "no-param-reassign": 0,
     "no-case-declarations": 0,
+    "no-constant-condition": ["error", { "checkLoops": false }],
     "prefer-destructuring": 0,
     "react/no-unescaped-entities": 0,
     "react/display-name": 0,
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -2,3 +2,8 @@ export const DARK_THEME = true;
 
 // The date format used in datetime-local fields.
 export const DATETIME_LOCAL = "yyyy-MM-dd'T'HH:mm";
+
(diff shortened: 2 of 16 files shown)


.eslintrc
```diff
@@ -63,6 +63,7 @@
     "default-case": 0,
     "no-param-reassign": 0,
     "no-case-declarations": 0,
+    "no-constant-condition": ["error", { "checkLoops": false }],
```
huwshimi, line 66: This is so we can use `while(true)` to keep the poll running

vladimir-cucu (approved):
All QA steps passed successfully using latest changes from jimm :rocket:

PR #1807, opened 2024-10-11
WD-15784 - fix: fix issues observed in k8s charm

vladimir-cucu:
## Done

- Fixed the second JQ query example which was wrong beforehand.
- Fixed the model info that showed the model owner's name in the left hand side. It got wrapped a fair bit as the owner was a service account.

## Details

- https://warthogs.atlassian.net/browse/WD-15784

## Screenshots

- Screen width < 1580px:
Screenshot from 2024-10-11 06-16-29

- Screen width >= 1580px:
Screenshot from 2024-10-11 06-16-13

### DIFF
--- a/src/components/EntityInfo/EntityInfo.tsx
+++ b/src/components/EntityInfo/EntityInfo.tsx
@@ -15,7 +15,13 @@ export default function EntityInfo({ data }: Props): JSX.Element {
         return (
           <div className="entity-info__grid-item" key={label}>
             <h4 className="p-muted-heading">{label}</h4>
-            <TruncatedTooltip message={value}>{value}</TruncatedTooltip>
+            {typeof value === "string" || typeof value === "number" ? (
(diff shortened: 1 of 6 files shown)


src/components/EntityInfo/EntityInfo.tsx
```diff
@@ -13,7 +15,7 @@ export default function EntityInfo({ data }: Props): JSX.Element {
         return (
           <div className="entity-info__grid-item" key={label}>
             <h4 className="p-muted-heading">{label}</h4>
-            <p data-name={label}>{value}</p>
+            <TruncatedTooltip message={value}>{value}</TruncatedTooltip>
```
vladimir-cucu, line 18: I thought it might be better to use `TruncatedTooltip` for all fields, in case there would be other fields populated with a long string. I checked and all `value` are string, so it works well with message prop.

huwshimi (approved):
Thanks for these fixes!

PR #1810, opened 2024-10-18
WD-16020 - refactor: replace BrowserRouter with createBrowserRouter

huwshimi:
## Done

- replace BrowserRouter with createBrowserRouter to support the routing in ReBACAdmin as they require the data APIs that are not supported by BrowserRouter.

## QA

- Check that the routing works as before.

## Details

https://warthogs.atlassian.net/browse/WD-16020

### DIFF
--- a/src/components/CaptureRoutes/CaptureRoutes.tsx
+++ b/src/components/CaptureRoutes/CaptureRoutes.tsx
@@ -1,0 +1,20 @@
+import { useEffect } from "react";
+import { Outlet, useLocation } from "react-router-dom";
+
+import useAnalytics from "hooks/useAnalytics";
+
+const CaptureRoutes = (): JSX.Element => {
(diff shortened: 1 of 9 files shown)


src/components/CaptureRoutes/CaptureRoutes.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import useAnalytics from "hooks/useAnalytics";
+
+const CaptureRoutes = (): JSX.Element => {
+  const sendAnalytics = useAnalytics();
+  const location = useLocation();
+
+  useEffect(() => {
+    // Send an analytics event when the URL changes.
+    sendAnalytics({
```
huwshimi, line 12: This is now in its own components as it needs to be rendered inside a route as it uses `useLocation`,

vladimir-cucu (approved):

PR #1842, opened 2025-01-31
fix: don't refetch nav permissions

huwshimi:
## Done

- Check if the primary nav permissions have already been fetch and if so don't fetch them again. This is required as the nav is part of the layout that wraps each component and gets recreated on each top level route change.

## QA

- Connect to JIMM.
- Check that the logs and permissions nav items are visible.
- Click on a different page e.g. go from models to controllers.
- Check that the logs and permissions items don't disappear and then reappear during navigation.

## Details

https://warthogs.atlassian.net/browse/WD-18854


Ninfa-Jeon (approved):
LGTM

PR #1843, opened 2025-02-03
fix: Compute latest revision from revisions array

Ninfa-Jeon:
## Done

Fixes the bug related to latest-revision.

## QA

- Update the value of a secret
- Remove the last added value
- The revision value should be changed to the next most latest revision

## Details

https://warthogs.atlassian.net/browse/WD-8783

### DIFF
--- a/src/components/secrets/RevisionField/RevisionField.tsx
+++ b/src/components/secrets/RevisionField/RevisionField.tsx
@@ -5,7 +5,7 @@ import type {
 } from "@canonical/react-components";
 import { FormikField, Select } from "@canonical/react-components";
 
-import { getSecretByURI } from "store/juju/selectors";
+import { getSecretByURI, getSecretLatestRevision } from "store/juju/selectors";
 import { useAppSelector } from "store/store";
 
 import { Label } from "./types";
@@ -22,6 +22,9 @@ const RevisionField = ({ modelUUID, secretURI, ...props }: Props) => {
   const secret = useAppSelector((state) =>
     getSecretByURI(state, modelUUID, secretURI),
   );
+  const latestRevision = useAppSelector((state) =>
+    getSecretLatestRevision(state, modelUUID, secretURI),
+  );
   return (
     <FormikField
       label={Label.REVISION}
@@ -29,7 +32,7 @@ const RevisionField = ({ modelUUID, secretURI, ...props }: Props) => {
       component={Select}
       options={
         [...(secret?.revisions ?? [])].reverse().map(({ revision }) => ({
-          label: `${revision}${revision === secret?.["latest-revision"] ? " (latest)" : ""}`,
+          label: `${revision}${revision === latestRevision ? " (latest)" : ""}`,
           value: revision.toString(),
         })) ?? []
       }
--- a/src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.test.tsx
+++ b/src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.test.tsx
@@ -186,6 +186,20 @@ describe("SecretContent", () => {
     expect(screen.getByText("a value")).toBeInTheDocument();
   });
 
+  it("can display content when no secrets", async () => {
+    state.juju.secrets = secretsStateFactory.build();
+    renderComponent(<SecretContent secretURI="secret:aabbccdd" />, {
+      state,
+      path,
+      url,
+    });
+    await userEvent.click(screen.getByRole("button", { name: Label.SHOW }));
+    expect(screen.queryByRole("heading", { name: "a key" })).toBeNull();
+    expect(
+      document.querySelector(".p-code-snippet__block"),
+    ).not.toBeInTheDocument();
+  });
+
   it("can display content errors", async () => {
     state.juju.secrets.abc123.content = modelSecretsContentFactory.build({
       loaded: true,
--- a/src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.tsx
+++ b/src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.tsx
@@ -24,6 +24,7 @@ import {
   getSecretsContentLoading,
   getModelUUIDFromList,
   getModelByUUID,
+  getSecretLatestRevision,
 } from "store/juju/selectors";
 import { useAppSelector, useAppDispatch } from "store/store";
 
@@ -42,6 +43,9 @@ const SecretContent = ({ secretURI }: Props) => {
   )?.wsControllerURL;
   const secret = useAppSelector((state) =>
     getSecretByURI(state, modelUUID, secretURI),
+  );
+  const latestRevision = useAppSelector((state) =>
+    getSecretLatestRevision(state, modelUUID, secretURI),
   );
   const content = useAppSelector((state) =>
     getSecretsContent(state, modelUUID),
@@ -90,9 +94,7 @@ const SecretContent = ({ secretURI }: Props) => {
           >
             {secret ? (
               <Formik<{ revision: string }>
-                initialValues={{
-                  revision: secret["latest-revision"].toString(),
-                }}
+                initialValues={{ revision: (latestRevision ?? "").toString() }}
                 onSubmit={(values) => {
                   getSecretContent(secretURI, Number(values.revision));
                 }}
--- a/src/panels/RemoveSecretPanel/Fields/Fields.test.tsx
+++ b/src/panels/RemoveSecretPanel/Fields/Fields.test.tsx
@@ -177,6 +177,30 @@ describe("Fields", () => {
     ).toBeInTheDocument();
   });
 
+  it("displays a message if secret does not exist", async () => {
+    state.juju.secrets.abc123 = modelSecretsFactory.build();
+    const handleRemoveSecret = vi.fn();
+    renderComponent(
+      <Formik<FormFields>
+        initialValues={{ removeAll: false, revision: "" }}
+        onSubmit={vi.fn()}
+      >
+        <Fields
+          secretURI="secret:aabbccdd"
+          hideConfirm={vi.fn()}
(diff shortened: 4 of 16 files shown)


src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.tsx
```diff
@@ -90,9 +92,7 @@ const SecretContent = ({ secretURI }: Props) => {
           >
             {secret ? (
               <Formik<{ revision: string }>
-                initialValues={{
-                  revision: secret["latest-revision"].toString(),
-                }}
+                initialValues={{ revision: latestRevision.toString() }}
```
huwshimi, line 95: Wherever we do `latestRevision.toString()` we probably need to make sure the value exists first.

src/utils/getLatestRevision.ts
```diff
@@ -0,0 +1,17 @@
+import type {
+  ListSecretResult,
+  SecretRevision,
+} from "@canonical/jujulib/dist/api/facades/secrets/SecretsV2";
+
+const getLatestRevision = (secret: ListSecretResult | null | undefined) => {
```
huwshimi, line 6:
```suggestion
const getLatestRevision = (secret?: ListSecretResult | null) => {
```

src/utils/getLatestRevision.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (!secret) return -1;
```
huwshimi, line 7: This should probably return null otherwise `-1` could get displayed to the user as the latest revision.

Also we tend not to use single line if statements (this should probably get added as a lint rule).

src/utils/getLatestRevision.ts
```diff
@@ -0,0 +1,17 @@
...
+  return (secret?.revisions ?? []).reduce(
+    (result: number, { revision }: SecretRevision) => {
+      if (revision > result) result = revision;
+      return result;
```
huwshimi, line 11: I think you can do this, which is slightly more terse, but both work:

```suggestion
      => Math.max(revision, result)
```

src/components/secrets/RevisionField/RevisionField.tsx
```diff
@@ -22,14 +23,15 @@ const RevisionField = ({ modelUUID, secretURI, ...props }: Props) => {
   const secret = useAppSelector((state) =>
     getSecretByURI(state, modelUUID, secretURI),
   );
+  const latestRevision = getLatestRevision(secret);
```
huwshimi, line 26: For cases like this where `getLatestRevision` is called on each render, you could wrap `getLatestRevision` in a selector that takes `secretURI` e.g. something like this:

```typescript
export const getSecretLatestRevision = createSelector(
  [getSecretByURI],
  (secret) => getLatestRevision(secret),
);
```

huwshimi (commented):
A few quick comments.

src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.tsx
Ninfa-Jeon: How about now?

src/utils/getLatestRevision.ts
Ninfa-Jeon: Done!

src/utils/getLatestRevision.ts
Ninfa-Jeon: Got it! Done.

src/utils/getLatestRevision.ts
Ninfa-Jeon: Smart. Done

src/components/secrets/RevisionField/RevisionField.tsx
Ninfa-Jeon: Does the current implementation respect this?

src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.tsx
huwshimi: Yep that looks good now.

src/panels/RemoveSecretPanel/Fields/Fields.tsx
```diff
@@ -53,8 +60,8 @@ const Fields = ({
         </>
       ) : (
         <p>
-          This secret has one revision ({secret?.["latest-revision"]}) and will
-          be completely removed.
+          This secret has one revision ({latestRevision}) and will be completely
```
huwshimi, line 63: `latestRevision` can be `null`, so while this is very unlikely, the logic in this component could change so it might be safer to do:

```suggestion
          This secret has one revision{latestRevision ? ` (${latestRevision})` : ""} and will be completely
```

Ninfa-Jeon: Done. Thanks!

src/utils/getLatestRevision.ts
```diff
@@ -0,0 +1,17 @@
+import type {
+  ListSecretResult,
+  SecretRevision,
+} from "@canonical/jujulib/dist/api/facades/secrets/SecretsV2";
+
+const getLatestRevision = (secret?: ListSecretResult | null) => {
+  if (!secret) {
+    return null;
+  }
+  return (secret?.revisions ?? []).reduce(
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done.

Ninfa-Jeon:
Todo: Add unit tests for newly added and uncovered code.

src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.test.tsx
```diff
@@ -186,6 +186,18 @@ describe("SecretContent", () => {
...
+  it("can display content when no secrets", async () => {
+    state.juju.secrets = secretsStateFactory.build();
+    renderComponent(<SecretContent secretURI="secret:aabbccdd" />, {
+      state,
+      path,
+      url,
+    });
+    await userEvent.click(screen.getByRole("button", { name: Label.SHOW }));
+    expect(screen.queryByRole("heading", { name: "a key" })).toBeNull();
+    expect(screen.queryByText("a value")).toBeNull();
```
huwshimi, line 198: Because we're not setting these values anywhere this sort of doesn't test anything, so it might be better to test that there are no code blocks:

```
expect(document.querySelector(".p-code-snippet__block")).not.toBeInTheDocument();
```

src/panels/RemoveSecretPanel/Fields/Fields.test.tsx
```diff
@@ -177,6 +177,36 @@ describe("Fields", () => {
...
+        />
+      </Formik>,
+      { state, path, url },
+    );
+    expect(
+      screen.queryByRole("checkbox", { name: Label.REMOVE_ALL }),
+    ).not.toBeInTheDocument();
+    expect(
+      screen.queryByRole("combobox", { name: Label.REVISION }),
+    ).not.toBeInTheDocument();
```
huwshimi, line 202: These checks are covered by the previous test.

src/store/juju/slice.ts
```diff
@@ -69,6 +69,7 @@ const getOrSetContentState = (state: JujuState, modelUUID: string) => {
       ...DEFAULT_MODEL_SECRETS_CONTENT,
     };
   }
+  // console.log("**", state.secrets[modelUUID].content?.errors); //issue
```
huwshimi, line 72: Debugging

src/store/juju/slice.ts
```diff
@@ -416,6 +417,7 @@ const slice = createSlice({
         } & WsControllerURLParam
       >,
     ) => {
+      console.log("**", action.payload.errors);
```
huwshimi, line 420: Debugging

src/utils/getLatestRevision.test.ts
```diff
@@ -0,0 +1,35 @@
...
+      ],
+    });
+    expect(getLatestRevision(secret)).toBe(2);
+  });
+
+  it("should return null when secret is null", () => {
+    expect(getLatestRevision()).toBeNull();
+  });
+
+  it("should return -1 when revisions are null", () => {
```
huwshimi, line 26: I don't think we ever want a case where -1 is returned so we should change this to return `null` if the array is empty.

src/utils/getLatestRevision.ts
```diff
@@ -0,0 +1,17 @@
+import type {
+  ListSecretResult,
+  SecretRevision,
+} from "@canonical/jujulib/dist/api/facades/secrets/SecretsV2";
+
+const getLatestRevision = (secret?: ListSecretResult | null) => {
+  if (!secret) {
```
huwshimi, line 7:
```suggestion
  if (!secret || !secret.revisions.length) {
```

huwshimi (commented):
A few more small suggestions here.

src/pages/EntityDetails/Model/Secrets/SecretContent/SecretContent.test.tsx
Ninfa-Jeon: Done

src/panels/RemoveSecretPanel/Fields/Fields.test.tsx
Ninfa-Jeon: Removed them

src/store/juju/slice.ts
Ninfa-Jeon: My bad, I missed it

src/store/juju/slice.ts
Ninfa-Jeon: Removed

src/utils/getLatestRevision.ts
Ninfa-Jeon: Thanks!

huwshimi (approved):
Thanks! Looks good.

PR #1845, opened 2025-02-04
feat(logging): Use loglevel library for error handling

Ninfa-Jeon:
## Done

- Installed  loglevel
- Replaced console.error with log.error
- Fixed and updated tests

## QA

- All tests should run

## Details

https://warthogs.atlassian.net/browse/WD-18724

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "fuse.js": "7.0.0",
     "lodash.isequal": "4.5.0",
     "lodash.mergewith": "4.6.2",
-    "loglevel": "^1.9.2",
+    "loglevel": "1.9.2",
     "prism-react-renderer": "2.4.1",
     "prismjs": "1.29.0",
     "react": "18.3.1",
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "fuse.js": "7.0.0",
     "lodash.isequal": "4.5.0",
     "lodash.mergewith": "4.6.2",
+    "loglevel": "^1.9.2",
     "prism-react-renderer": "2.4.1",
     "prismjs": "1.29.0",
     "react": "18.3.1",
--- a/src/components/App/App.test.tsx
+++ b/src/components/App/App.test.tsx
@@ -1,5 +1,4 @@
 import { render, screen } from "@testing-library/react";
-import log from "loglevel";
 import * as reactGA from "react-ga";
 import { Provider } from "react-redux";
 import configureStore from "redux-mock-store";
@@ -27,21 +26,9 @@ vi.mock("react-router", async () => {
   };
 });
 
-vi.mock("loglevel", async () => {
-  const actual = await vi.importActual("loglevel");
-  return {
-    ...actual,
-    error: vi.fn(),
-  };
-});
-
 const mockStore = configureStore([]);
 
 describe("App", () => {
-  beforeEach(() => {
-    vi.spyOn(log, "error").mockImplementation(() => vi.fn());
-  });
-
   afterEach(() => {
     vi.resetAllMocks();
     vi.unstubAllEnvs();
--- a/src/components/App/App.test.tsx
+++ b/src/components/App/App.test.tsx
@@ -1,4 +1,5 @@
 import { render, screen } from "@testing-library/react";
+import log from "loglevel";
 import * as reactGA from "react-ga";
 import { Provider } from "react-redux";
 import configureStore from "redux-mock-store";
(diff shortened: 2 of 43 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
     "fuse.js": "7.0.0",
     "lodash.isequal": "4.5.0",
     "lodash.mergewith": "4.6.2",
+    "loglevel": "^1.9.2",
```
huwshimi, line 68: We usually pin our deps to specific versions

```suggestion
    "loglevel": "1.9.2",
```

src/components/App/App.test.tsx
```diff
@@ -26,20 +27,22 @@ vi.mock("react-router", async () => {
   };
 });
 
+vi.mock("loglevel", async () => {
```
huwshimi, line 30: I don't think we get any benefit from testing that we're passing errors to loglevel and it increases our maintenance burden so I think we can remove these mocks/tests.

huwshimi (commented):
Not sure if you want to do this in the same PR, but we should probably do the same for our `console.log` calls as well.

We might want to default to silent in prod and maybe in tests as well.

For prod we could have a console.log when the app loads to output something like `'Run log.setLevel("trace") to output errors to the console.'` and then expose loglevel as a global `log` variable.

package.json
Ninfa-Jeon: Done.

src/components/App/App.test.tsx
Ninfa-Jeon: I've removed it from most spots except for the ones where we wait for the console error until making another check.

Ninfa-Jeon:
> Not sure if you want to do this in the same PR, but we should probably do the same for our `console.log` calls as well.
> 
> We might want to default to silent in prod and maybe in tests as well.
> 
> For prod we could have a console.log when the app loads to output something like `'Run log.setLevel("trace") to output errors to the console.'` and then expose loglevel as a global `log` variable.

While I'm working on setting the log levels, I'm not sure I've initialised it right in test setup file.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
```diff
@@ -163,7 +167,7 @@ describe("CharmsAndActionsPanel", () => {
     } = renderComponent(<CharmsAndActionsPanel />, { path, url, state });
     expect(juju.getCharmsURLFromApplications).toHaveBeenCalledTimes(1);
     await waitFor(() =>
-      expect(console.error).toHaveBeenCalledWith(
+      expect(logger.error).toHaveBeenCalledWith(
```
huwshimi, line 170: If you remove this then you can move `expect(container.querySelector(".p-panel__title")).toContainHTML(` on line 175 inside this `waitFor` instead and the test will pass.

src/panels/ConfigPanel/ConfigPanel.test.tsx
```diff
@@ -395,7 +402,7 @@ describe("ConfigPanel", () => {
     renderComponent(<ConfigPanel />, { state, path, url });
     expect(getApplicationConfig).toHaveBeenCalledTimes(1);
     await waitFor(() => {
-      expect(console.error).toHaveBeenCalledWith(
+      expect(logger.error).toHaveBeenCalledWith(
```
huwshimi, line 405: If you remove this then you just need to change line 410 to `const configErrorNotification = await screen.findByText(` to make this test pass.

huwshimi (commented):
A couple of suggestions here so that you can remove the remaining logging mocks.

huwshimi:
> While I'm working on setting the log levels, I'm not sure I've initialised it right in test setup file.

It seems like it's working for me, there are no errors being displayed in the console and if I add `logger.setLevel("DEBUG");` to a test file I see the errors.

src/panels/CharmsAndActionsPanel/CharmsAndActionsPanel.test.tsx
Ninfa-Jeon: Done.

src/panels/ConfigPanel/ConfigPanel.test.tsx
Ninfa-Jeon: Done.

Ninfa-Jeon:
> > While I'm working on setting the log levels, I'm not sure I've initialised it right in test setup file.
> 
> It seems like it's working for me, there are no errors being displayed in the console and if I add `logger.setLevel("DEBUG");` to a test file I see the errors.

Okay, I've not made further changes to this since it was reported to be working fine. Additionally, I have replaced `console.log` with `logger.log`

src/index.tsx
```diff
@@ -159,13 +165,13 @@ function bootstrap() {
...
       .then(unwrapResult)
-      .catch((error) => console.error(Label.POLLING_ERROR, error));
+      .catch((error) => logger.error(Label.POLLING_ERROR, error));
   }
 
   getRoot()?.render(
     <Provider store={reduxStore}>
       <StrictMode>
-        <App />
+        <App logLevel={logLevel} />
```
huwshimi, line 174: Is there any need to delay setting the default level until this component is rendered? It looks like that will mean that the errors in this file will bypass the default level (or at least have a race condition in the `POLLING_ERROR` case above).

src/testing/setup.ts
```diff
@@ -31,7 +35,7 @@ const animateMock = () => {
 if (!window.HTMLDivElement.prototype.animate) {
   window.HTMLDivElement.prototype.animate = animateMock;
 } else {
-  console.error(
+  logger.error(
```
huwshimi, line 38: I think this one can be a standard `console.error` as we want this to appear no matter the log level (and this is not user facing code).

src/index.tsx
Ninfa-Jeon: Moved it up a component

src/testing/setup.ts
Ninfa-Jeon: Reverted

huwshimi (approved):
Thanks for all the work on this.

PR #1847, opened 2025-02-06
feat(analytics): Implement usage analytics with react-ga4

Ninfa-Jeon:
## Done

- Upgraded react-ga to react-ga4
- Implemented usage analytics using event params

## QA

- Login to dashboard
- Dashboard should send values like dashboard version, controller version and whether you’re using JAAS on successful login

## Details

https://warthogs.atlassian.net/browse/WD-18723

### DIFF
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -1,4 +1,4 @@
-import { initialize, pageview } from "react-ga";
+import ReactGA from "react-ga4";
 
 import ConnectionError from "components/ConnectionError";
 import ErrorBoundary from "components/ErrorBoundary";
@@ -12,8 +12,11 @@ function App() {
   const isProduction = import.meta.env.PROD;
   const analyticsEnabled = useAppSelector(getAnalyticsEnabled);
   if (isProduction && analyticsEnabled) {
-    initialize("UA-1018242-68");
-    pageview(window.location.href.replace(window.location.origin, ""));
+    ReactGA.initialize("UA-1018242-68"); // TODO: should use the GA4 Measurement ID (which starts with "G-")
+    ReactGA.send({
+      hitType: "page_view",
+      page: window.location.href.replace(window.location.origin, ""),
+    });
   }
 
   return (
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -12,9 +12,17 @@ function App() {
   const isProduction = import.meta.env.PROD;
   const analyticsEnabled = useAppSelector(getAnalyticsEnabled);
   if (isProduction && analyticsEnabled) {
-    ReactGA.initialize("UA-1018242-68"); // TODO: should use the GA4 Measurement ID (which starts with "G-")
+    ReactGA.initialize("G-JHXHM8VXJ1", {
+      gtagOptions: {
+        custom_map: {
+          dimension1: "dashboardVersion",
+          dimension2: "controllerVersion",
+          dimension3: "isJuju",
+        },
+      },
+    });
     ReactGA.send({
-      hitType: "page_view",
+      hitType: "pageview",
       page: window.location.href.replace(window.location.origin, ""),
     });
   }
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -11,8 +11,11 @@ import "../../scss/index.scss";
 function App() {
   const isProduction = import.meta.env.PROD;
   const analyticsEnabled = useAppSelector(getAnalyticsEnabled);
-  if (isProduction && analyticsEnabled) {
-    ReactGA.initialize("G-JHXHM8VXJ1"); // send { gaOptions: { debug_mode: true } } as second param to enable debugView
+  if (analyticsEnabled) {
+    ReactGA.initialize(
+      "G-JHXHM8VXJ1",
+      isProduction ? {} : { gaOptions: { debug_mode: true } },
+    );
     ReactGA.send({
       hitType: "pageview",
       page: window.location.href.replace(window.location.origin, ""),
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -12,13 +12,7 @@ function App() {
   const isProduction = import.meta.env.PROD;
   const analyticsEnabled = useAppSelector(getAnalyticsEnabled);
   if (isProduction && analyticsEnabled) {
-    ReactGA.initialize("G-JHXHM8VXJ1", {
-      gaOptions: {
-        dimension1: "dashboardVersion",
-        dimension2: "controllerVersion",
-        dimension3: "isJuju",
-      },
-    });
+    ReactGA.initialize("G-JHXHM8VXJ1"); // send { gaOptions: { debug_mode: true } } as second param to enable debugView
     ReactGA.send({
       hitType: "pageview",
       page: window.location.href.replace(window.location.origin, ""),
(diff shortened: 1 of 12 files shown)


src/hooks/useAnalytics.tsx
```diff
@@ -7,21 +7,28 @@ type AnalyticMessage = {
...
+    action = "",
+    eventParams = {},
+  }: AnalyticMessage) => {
     const isProduction = import.meta.env.PROD;
     if (!isProduction || !analyticsEnabled) {
       return;
     }
     if (path) {
-      pageview(path);
+      ReactGA.send({ hitType: "page_view", page: path });
```
huwshimi, line 26: It looks like this should be `pageview`.

src/components/App/App.tsx
```diff
@@ -12,8 +12,11 @@ function App() {
   const isProduction = import.meta.env.PROD;
   const analyticsEnabled = useAppSelector(getAnalyticsEnabled);
   if (isProduction && analyticsEnabled) {
-    initialize("UA-1018242-68");
-    pageview(window.location.href.replace(window.location.origin, ""));
+    ReactGA.initialize("UA-1018242-68"); // TODO: should use the GA4 Measurement ID (which starts with "G-")
```
huwshimi, line 15: It looks like we also need to initialise the dimensions for these to be sent: https://github.com/codler/react-ga4/issues/2#issuecomment-1398617476.

src/components/LogIn/LogIn.tsx
```diff
@@ -71,6 +79,20 @@ export default function LogIn() {
...
+  useEffect(() => {
+    if (userIsLoggedIn) {
+      sendAnalytics({
+        category: "Authentication",
+        action: "User Login",
+        eventParams: {
+          dashboardVersion: appVersion ?? "",
+          controllerVersion: controllerVersion ?? "",
+          isJuju: (!!isJuju).toString(),
+        },
```
huwshimi, line 91: Looking at how dimensions work it looks like we can send them will all events so that for any event we can see what version of the dashboard the event was from etc.

To do that we could move these additional params (and related selectors) into the `useAnalytics` hook and append them to all events.

src/hooks/useAnalytics.tsx
Ninfa-Jeon: My bad. Fixed now

src/components/App/App.tsx
Ninfa-Jeon: I've followed this and another issue for refactoring this part. Need to confirm if it works now.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: I've moved the additional params and their respective selectors to the useAnalytics hook so that they are always appended to a ReactGA.event call. The tests for the hook may need some additions after this change. Need to check.

src/components/LogIn/LogIn.tsx
```diff
@@ -71,6 +73,15 @@ export default function LogIn() {
     });
   }, [visitURLs]);
 
+  useEffect(() => {
+    if (userIsLoggedIn) {
+      sendAnalytics({
+        category: "Authentication",
+        action: "User Login",
```
huwshimi, line 80: It might be helpful to also include the type of authentication:

```suggestion
        action: `User Login (${config?.authMethod})`,
```

huwshimi (commented):
For some reason the dimensions are still not included in `dataLayer` so we might need to investigate this a bit more.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Done

src/hooks/useAnalytics.test.ts
```diff
@@ -55,20 +58,26 @@ describe("useAnalytics", () => {
...
+      page: "/some/path",
+    });
   });
 
   it("can send events", () => {
     vi.spyOn(store, "useAppSelector").mockImplementation(
       vi.fn().mockReturnValue(true),
     );
-    const { result } = renderHook(() => useAnalytics());
+    const { result } = renderWrappedHook(() => useAnalytics());
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done

Ninfa-Jeon:
> For some reason the dimensions are still not included in `dataLayer` so we might need to investigate this a bit more.

I tried various things too but they don't seem to be working either. For now, pushing the successful changes. Investigating this.

Ninfa-Jeon:
> > For some reason the dimensions are still not included in `dataLayer` so we might need to investigate this a bit more.
> 
> I tried various things too but they don't seem to be working either. For now, pushing the successful changes. Investigating this.

Okay, so I've got the values to show up in dataLayer under user_properties. I have verified for both Juju and Jimm controller (PFA)
Screenshot 2025-02-18 at 3 07 31 PM
Screenshot 2025-02-18 at 3 11 15 PM

src/hooks/useAnalytics.test.ts
huwshimi, line 1: I think we also need a test that it sends the correct values for `{ controllerVersion: "", dashboardVersion: "", isJuju: "false" }` from the Redux store.

src/store/middleware/model-poller.ts
```diff
@@ -132,12 +133,24 @@ export const modelPollerMiddleware: Middleware<
           return;
         }
 
+        const isProduction = import.meta.env.PROD;
+        const analyticsEnabled = window.jujuDashboardConfig?.analyticsEnabled;
```
huwshimi, line 137: This should get the value from the redux store.
```suggestion
        const analyticsEnabled = getAnalyticsEnabled(reduxStore.getState());
```

src/utils/analytics.ts
```diff
@@ -0,0 +1,29 @@
+import ReactGA from "react-ga4";
+
+type EventParams = {
+  dashboardVersion: string;
+  controllerVersion: string;
+  isJuju: string;
+};
+
+type AnalyticsMessage = { path?: string; category?: string; action?: string };
```
huwshimi, line 9: This type is duplicated here and in the `useAnalytics` hook. This one could be exported and then import it in the hook.

src/utils/analytics.ts
```diff
@@ -0,0 +1,29 @@
...
+const analytics = (
+  analyticsEnabled: boolean,
+  eventParams: EventParams,
+  { path, category = "", action = "" }: AnalyticsMessage,
+) => {
+  if (!analyticsEnabled) {
+    return;
+  }
+  if (path) {
+    ReactGA.send({ hitType: "pageview", page: path });
```
huwshimi, line 20: We should pass `...eventParams,` to the pageview events as well. This data won't be available until after the first few pageview events but it should help us get some idea of the versions being used.

src/hooks/useAnalytics.test.ts
Ninfa-Jeon: Done

src/store/middleware/model-poller.ts

src/utils/analytics.ts
Ninfa-Jeon: Done.

src/utils/analytics.ts
Ninfa-Jeon: Done.

src/utils/analytics.ts
```diff
@@ -0,0 +1,33 @@
...
+  isJuju: string;
+};
+
+export type AnalyticsMessage = {
+  path?: string;
+  category?: string;
+  action?: string;
+};
+
+const analytics = (
```
huwshimi, line 15: It might be a good idea to have a comment here explaining that in a component you can use the useAnalytics hook which fetches the additional even data.

Ninfa-Jeon: Done!

huwshimi (approved):
This looks good. Thanks!

PR #1871, opened 2025-02-28
feat: Implement rebac feature flag via query params

Ninfa-Jeon:
## Done

This commit introduces query parameter-based feature flags to control the enabling and disabling of rebac

- Added `useFeatureFlags` hook to manage `enable-flag` and `disable-flag` query parameters.
- Implemented `getEnabledFlag` and `getDisabledFlag` functions within the hook to determine flag status.
- Updated relevant components to utilize the feature flags and adjust rebac behavior accordingly.

BREAKING CHANGE: The rebac logic is now controlled by query parameters. Existing behaviour may be affected if the query parameters are not set appropriately.
NOTE: The `enable-flag=rebac` takes priority over `disable-flag=rebac` to avoid looping

## QA

- Try accessing the dashboard via local controller
- All should be fine
- Login to dashboard via JIMM
- Without any query params, it should hide rebac-related fields
- Add query param `enable-flag=rebac`
- Local storage should now show a `rebac` key with value `true`
- All rebac-related fields should appear
- Add query param `disable-flag=rebac`
- Local storage should now show a `rebac` key with value `false`
- All rebac-related fields should be hidden

## Details

https://warthogs.atlassian.net/browse/WD-18923


src/store/middleware/model-poller.test.ts
```diff
@@ -293,6 +294,7 @@ describe("model poller", () => {
   });
 
   it("updates the controller features if JIMM >= 4", async () => {
+    localStorage.setItem("rebac", JSON.stringify("true"));
```
fasih-mehmood, line 297: You don't have to use `JSON.stringify` here as "true" is already a string.

```suggestion
    localStorage.setItem("rebac", "true");
```

src/components/LogIn/LogIn.tsx
```diff
@@ -26,6 +28,8 @@ import UserPassForm from "./UserPassForm";
 import { ErrorResponse, Label, TestId } from "./types";
 
 export default function LogIn() {
+  const { getEnabledFlag, getDisabledFlag } = useFeatureFlags();
```
huwshimi, line 31: I think it might be better to move this hook into the `App` component as it isn't related to authentication.

src/components/LogIn/LogIn.tsx
```diff
@@ -38,6 +42,14 @@ export default function LogIn() {
...
   const visitURLs = useAppSelector(getVisitURLs);
 
+  useEffect(() => {
+    if (getEnabledFlag("rebac")) {
+      setRebacFlag(true);
+    } else if (getDisabledFlag("rebac")) {
+      setRebacFlag(false);
+    }
+  }, [getEnabledFlag, getDisabledFlag, setRebacFlag]);
+
```
huwshimi, line 52: I think this `useEffect` and the `useLocalStorage` above can live inside the `useFeatureFlags` hook as I don't think we need to handle getting the query params separately from setting local storage.

src/components/LogIn/LogIn.tsx
```diff
@@ -26,6 +28,8 @@ import UserPassForm from "./UserPassForm";
 import { ErrorResponse, Label, TestId } from "./types";
 
 export default function LogIn() {
+  const { getEnabledFlag, getDisabledFlag } = useFeatureFlags();
+  const [_, setRebacFlag] = useLocalStorage<boolean>("rebac", false);
```
huwshimi, line 32: This should use a const for the `"rebac"` key so we can define it in one place. Like we do here:

https://github.com/canonical/juju-dashboard/blob/602b27744e1c9e48545cc8b5728f1bb1a6469037/src/pages/AdvancedSearch/SearchForm/SearchForm.tsx#L39-L41

src/hooks/useFeatureFlags.ts
huwshimi, line 1: This file needs some tests.

src/store/middleware/model-poller.ts
```diff
@@ -152,13 +152,16 @@ export const modelPollerMiddleware: Middleware<
           }),
         );
         const jimmVersion = conn.facades.jimM?.version ?? 0;
+        const rebacEnabled = JSON.parse(
```
huwshimi, line 155: A small thing but this might be better to be named `rebacFlagEnabled` to distinguish it from whether the controller supports it etc.

src/store/middleware/model-poller.ts
```diff
@@ -152,13 +152,16 @@ export const modelPollerMiddleware: Middleware<
...
+          window.localStorage.getItem("rebac") ?? "false",
+        );
         reduxStore.dispatch(
           generalActions.updateControllerFeatures({
             wsControllerURL,
             features: {
               auditLogs: jimmVersion >= 4,
               crossModelQueries: jimmVersion >= 4,
-              rebac: jimmVersion >= 4,
+              rebac: rebacEnabled && jimmVersion >= 4,
```
huwshimi, line 164: We should add a test that this can be controlled via the feature flag.

Ninfa-Jeon:
> Don't mind me, just passing through here 😄
> 
> I recently implemented `localStorage` based feature flags in [Canonical Console](https://github.com/canonical/console.canonical.com) so as soon as I read "feature flag" in the PR title, I became interested to see how other devs implement it.
> 
> Flags in query parameter is an interesting way to implement feature flagging. Definitely a lot easier to explain to end users then to have them open the dev tools and run a command to update the local storage xD
> 
> Great work!

Thank you @fasih-mehmood ! The idea for this was Huw's so its all thanks to him haha.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Somehow, I first implemented this in `App` component but it wouldn't work. Will retry.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Makes sense. Working on it

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: we cannot place the useFeatureFlags hook in App.tsx because it uses useQueryParam hook that in turn uses useSearchParams hook. These hooks can only be placed in a component that is wrapped in a router, which essentially, comes under the App.tsx




a different approach could be to use browser router from react-router-dom but that would require a lot of refactor

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Since we are now moving to a broader implementation of this that essentially would just place all enable-flag params into an array of enabled flags in the local storage, we no longer need this.

Ninfa-Jeon: All the enabled flags could be accessed via the "flags" key in local storage. Will be moving that to a const instead.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Done.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Done.

src/hooks/useFeatureFlags.ts
Ninfa-Jeon: Done.

src/components/LogIn/LogIn.tsx
Ninfa-Jeon: Moved the hook call to `<CaptureRoutes/>`

src/components/LogIn/LogIn.tsx
```diff
@@ -26,6 +27,7 @@ import UserPassForm from "./UserPassForm";
 import { ErrorResponse, Label, TestId } from "./types";
 
 export default function LogIn() {
+  useFeatureFlags();
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done

src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+    let result = enabledFlags.filter((flag) => !disabledParams.includes(flag));
+
+    // append enabled params to flag list
+    result = enabledParams.reduce((acc, param) => {
+      if (!disabledParams.includes(param) && !acc.includes(param)) {
+        return [...acc, param];
+      }
+      return acc;
+    }, result);
+    return result;
```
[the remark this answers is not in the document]
Ninfa-Jeon: Thanks for this

src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+    "disable-flag": [],
+  });
+  const [enabledFlags, setEnabledFlags] = useLocalStorage<string[]>(
+    ENABLED_FLAGS,
+    [],
+  );
+
+  const getFinalFlags = useCallback(() => {
+    const enabledParams = queryParams["enable-flag"] ?? [];
+    const disabledParams = queryParams["disable-flag"] ?? [];
```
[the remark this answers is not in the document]
Ninfa-Jeon: Good catch

src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
+import { useCallback, useEffect, useMemo } from "react";
+
+import { ENABLED_FLAGS } from "consts";
+
+import useLocalStorage from "./useLocalStorage";
+import { useQueryParams } from "./useQueryParams";
+
+type FlagQueryParams = {
+  "enable-flag": string[] | null;
+  "disable-flag": string[] | null;
```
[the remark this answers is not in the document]
Ninfa-Jeon: Right.

src/store/middleware/model-poller.ts
```diff
@@ -170,13 +170,16 @@ export const modelPollerMiddleware: Middleware<
           }),
         );
         const jimmVersion = conn.facades.jimM?.version ?? 0;
+        const rebacFlagEnabled = JSON.parse(
+          window.localStorage.getItem("flags") ?? "[]",
+        ).includes("rebac");
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done.

src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,48 @@
...
+      return acc;
+    }, result);
+    return result;
+  }, [queryParams, enabledFlags]);
+
+  const finalFlags = useMemo(() => getFinalFlags(), [getFinalFlags]);
+
+  useEffect(() => {
+    // deep comparison to prevent unnecessary re-renders
+    if (JSON.stringify(finalFlags) !== JSON.stringify(enabledFlags)) {
```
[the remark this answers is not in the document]
Ninfa-Jeon: Modified the logic and moved it to `useLocalStorage`

src/hooks/useFeatureFlags.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  it("should initialize with empty flags from local storage", () => {
+    vi.spyOn(LocalStorage, "default").mockReturnValue([[], vi.fn()]);
+    vi.spyOn(QueryParams, "useQueryParams").mockReturnValue([
+      { "enable-flag": null, "disable-flag": null },
+      vi.fn(),
+    ]);
+
+    renderHook(() => useFeatureFlags());
+
+    expect(LocalStorage.default).toHaveBeenCalledWith("flags", []);
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done, will keep in mind.

src/featureFlags.ts
```diff
@@ -0,0 +1,3 @@
+export enum FeatureFlags {
```
huwshimi, line 1: I think this type can just go in `src/types.ts`

src/hooks/useFeatureFlags.ts
```diff
@@ -0,0 +1,38 @@
...
+    const enabled = Array.from(new Set([...enabledFlags, ...enabledParams]));
+    // Filter out any flags that should be disabled.
+    return enabled.filter((flag) => !disabledParams.includes(flag));
+  }, [queryParams, enabledFlags]);
+
+  const finalFlags = useMemo(() => getFinalFlags(), [getFinalFlags]);
+
+  useEffect(() => {
+    setEnabledFlags(finalFlags);
+  }, [setEnabledFlags, finalFlags, enabledFlags]);
```
huwshimi, line 37: I guess I forgot to submit this comment yesterday, but the effect of the `useCallback` then `useMemo` and then `useEffect` is the same as wrapping it all in a single `useEffect.`

src/hooks/useLocalStorage.ts
```diff
@@ -23,9 +24,11 @@ function useLocalStorage<V>(
...
   const setValue = (value: V) => {
     try {
-      const stringified = JSON.stringify(value);
-      setStoredValue(value);
-      window.localStorage.setItem(key, stringified);
+      if (!fastDeepEqual(storedValue, value)) {
+        const stringified = JSON.stringify(value);
+        setStoredValue(value);
+        window.localStorage.setItem(key, stringified);
+      }
```
huwshimi, line 31: We should have a test for this.

src/utils/isFeatureFlagEnabled.ts
```diff
@@ -0,0 +1,8 @@
+import { ENABLED_FLAGS } from "consts";
+
+const isFeatureFlagEnabled = (featureName: string) =>
```
huwshimi, line 3: Nice, thanks!

huwshimi (commented):
A couple more things and then I think this will be done!

huwshimi (approved):
This is great, thanks!

PR #1872, opened 2025-03-04
fix: overflowing tooltip on model page

### LINKED ISSUE #1866: Filter buttons are broken
The buttons in the search and filter menu were broken in a recent Vanilla update.

Image

### LINKED ISSUE #1867: No secrets message missing spacing
The message when there are no secrets has lost its padding (in the configure app panel).

Image

### LINKED ISSUE #1870: Long model data is not truncated
When viewing the model details the side panel content breaks its container if it contains long strings.

This should probably be truncated or wrapped.

Image

andogq:
## Done

- Fix `EntityInfo` overflowing items by importing `truncated-tooltip` into top level SCSS file
  - Include new test case in `TruncatedTooltip` to ensure the style is overridden
- Update `EntityInfo` tests
  - Clean up references to `topology`
  - Add new test for empty entity data
- Clean up duplicated SCSS includes (close #1866, close #1867)
- Turn on CSS source maps in dev mode

## Details

- closes #1870 
- https://warthogs.atlassian.net/jira/software/c/projects/WD/issues/WD-19878

## Screenshots

image

### DIFF
--- a/src/components/EntityInfo/EntityInfo.test.tsx
+++ b/src/components/EntityInfo/EntityInfo.test.tsx
@@ -1,17 +1,11 @@
 import { screen } from "@testing-library/react";
-import { vi } from "vitest";
 
 import { renderComponent } from "testing/utils";
 
 import EntityInfo from "./EntityInfo";
 
-vi.mock("components/Topology", () => {
-  const Topology = () => <div className="topology"></div>;
-  return { default: Topology };
-});
-
 describe("Entity info", () => {
-  it("renders the expanded topology on click", () => {
+  it("renders the data", () => {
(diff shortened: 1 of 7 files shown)


huwshimi (commented):
This is looking good, but while we're at it maybe we can have a look at why there is some space between the element that is being hovered and the tooltip:

Screenshot 2025-03-04 at 4 18 08 pm

Also there appears to be some padding missing from the tooltip. Usually they appear like this:

Screenshot 2025-03-04 at 4 22 40 pm

huwshimi (approved):
All good now, thanks!

PR #1873, opened 2025-03-05
fix: apply word break in config panel

### LINKED ISSUE #1868: Log config text breaks panel
When a config field has long text that doesn't get broken onto a new line it causes the split panel to change size.

Image

andogq:
## Done

- use `break-word` in `pre` containing configuration description, to prevent wrapping from changing panel aside width

## Details

- close #1868 
- WD-19882

## Screenshots

image


Ninfa-Jeon (approved):
LGTM

PR #1874, opened 2025-03-05
fix: correctly disable routes for disabled features

huwshimi:
## Done

- Correctly disable routes for disabled features.
- Check user permissions if the feature is enabled.

## QA

- Connect to a local Juju controller.
- Visit one of the disabled features by typing the URL e.g. /search or /logs or /permissions (these won't be in the side nav).
- It should not require you to log in and instead display a 404 page.
- Connect to JIMM.
- Make sure rebac is enabled with `?enable-flag=rebac`
- You should see the search, logs and permissions nav items.
- Clicking on them should display the pages.
- Enter `?disable-flag=rebac` and then visit /permissions
- It should display a 404 page.

## Details

https://warthogs.atlassian.net/browse/WD-19883

### DIFF
--- a/src/components/CheckPermissions/CheckPermissions.test.tsx
+++ b/src/components/CheckPermissions/CheckPermissions.test.tsx
@@ -7,7 +7,7 @@ import { renderComponent } from "testing/utils";
 import CheckPermissions from "./CheckPermissions";
 
 describe("CheckPermissions", () => {
-  it("displays the loading state", async () => {
+  it("displays the loading state", () => {
     renderComponent(
       <CheckPermissions loading>secret content</CheckPermissions>,
     );
@@ -16,14 +16,14 @@ describe("CheckPermissions", () => {
     ).toBeInTheDocument();
   });
 
-  it("displays a page not found message if it's not allowed", async () => {
+  it("displays a page not found message if it's not allowed", () => {
     renderComponent(
       <CheckPermissions allowed={false}>secret content</CheckPermissions>,
     );
     expect(screen.getByText(PageNotFoundLabel.NOT_FOUND)).toBeInTheDocument();
   });
 
-  it("displays a page not found message if it's not allowed", async () => {
+  it("displays a page not found message if it's not allowed", () => {
     renderComponent(
       <CheckPermissions allowed>secret content</CheckPermissions>,
     );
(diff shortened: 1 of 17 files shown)


src/components/Routes/Routes.tsx
```diff
@@ -53,25 +32,23 @@ export function Routes() {
...
-      path: urls.search,
-      element: <AdvancedSearch />,
-    });
-  }
-
-  if (requiresAuthentication || rebacEnabled) {
-    authenticatedRoutes.push({
-      path: `${urls.permissions}/*`,
-      element: <Permissions />,
-    });
```
huwshimi, line 74: The reason the pages could be accessed directly via the URL is that it would take you to the page, fetch the data to determine if the feature was enabled, and then if it wasn't enabled React Router would remove the route from the array but wouldn't update the current route and display the 404 page (but if you tried to navigate to a page after the data was loaded it wouldn't be in the array so then it would display the 404).

Instead I've moved the 404 handling to the actual views so they display the 404 after checking whether the feature is enabled and the user's permissions.

src/pages/Logs/Logs.test.tsx
```diff
@@ -44,6 +79,33 @@ describe("Logs", () => {
     vi.restoreAllMocks();
   });
 
+  it("doesn't display logs if the feature is disabled", async () => {
```
Ninfa-Jeon, line 82: ditto! we don't need `async` for these tests either

Ninfa-Jeon (approved):
LGTM, just minor suggestions

src/components/Routes/Routes.tsx
Ninfa-Jeon: Makes sense, thanks for the clarification!

src/components/CheckPermissions/CheckPermissions.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { screen } from "@testing-library/dom";
+
+import { LoadingSpinnerTestId } from "components/LoadingSpinner";
+import { PageNotFoundLabel } from "pages/PageNotFound";
+import { renderComponent } from "testing/utils";
+
+import CheckPermissions from "./CheckPermissions";
+
+describe("CheckPermissions", () => {
+  it("displays the loading state", async () => {
```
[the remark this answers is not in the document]
huwshimi: Done.

src/pages/Logs/Logs.test.tsx
huwshimi: Done.

PR #1875, opened 2025-03-06
fix: unit link generation

### LINKED ISSUE #1869: Clicking a unit in the advanced search causes error
When you do a cross-model search that contains a unit and then click the unit name it displays this error:

Image

This might be because it hasn't loaded the model data at that point (might need to refresh the search page before clicking on the unit to clear the model data so that this can be reproduced).

andogq:
## Done

- correctly generate unit links in advanced search
- create not found page for units instead of crashing

## QA

- Pull code
- Open http://0.0.0.0:8036/search?q=.
- Expand root -> 0 -> units, click on a unit (eg `source/4`)
- Should correctly navigate to source page
- Navigate to junk unit page (add something to end of current URL) for not found page

## Details

- WD-19966
- close #1869 

## Screenshots

https://github.com/user-attachments/assets/14059413-baca-44a3-b337-ac320904d38b

### DIFF
--- a/src/pages/EntityDetails/Unit/Unit.tsx
+++ b/src/pages/EntityDetails/Unit/Unit.tsx
@@ -1,10 +1,11 @@
 import { MainTable } from "@canonical/react-components";
 import { useMemo } from "react";
 import { useSelector } from "react-redux";
-import { useParams } from "react-router";
+import { Link, useParams } from "react-router";
 
 import EntityInfo from "components/EntityInfo";
 import InfoPanel from "components/InfoPanel";
+import NotFound from "components/NotFound";
 import type { EntityDetailsRoute } from "components/Routes";
 import type { ApplicationData, MachineData } from "juju/types";
 import {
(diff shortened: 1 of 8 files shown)


src/pages/EntityDetails/Unit/Unit.tsx
```diff
@@ -38,23 +41,28 @@ export default function Unit() {
   const machines = useSelector(getModelMachines(modelUUID));
   const isK8s = useAppSelector((state) => isKubernetesModel(state, modelUUID));
 
+  const unit = useMemo(
+    () => (unitIdentifier && units?.[unitIdentifier]) || null,
+    [units, unitIdentifier],
+  );
```
huwshimi, line 47: There's probably no value in a `useMemo` here (it might even be slower) as retrieving the unit will be very fast.

huwshimi (commented):
Thanks for this fix, just a couple of small comments.

huwshimi (approved):
Thanks this looks good now!

PR #1878, opened 2025-03-07
fix: show error on initial model error

andogq:
## Done

- previously `modelsLoaded` took precedence over checking `modelError`
- includes test to ensure spinner is hidden and no other content is displayed on the page

## QA

- Pull code
- Run `dotrun clean && dotrun serve`
- Open http://0.0.0.0:8036/
- Apply patch to `model-poller` (see below)
- Open dashboard

<details>
<summary>Patch with error</summary>

```patch
diff --git a/src/store/middleware/model-poller.ts b/src/store/middleware/model-poller.ts
index 39d9b6e6..7012da47 100644
--- a/src/store/middleware/model-poller.ts
+++ b/src/store/middleware/model-poller.ts
@@ -217,6 +217,7 @@ export const modelPollerMiddleware: Middleware<
               const models = await conn.facades.modelManager?.listModels({
                 tag: identity,
               });
+              throw new Error("model loading failure");
               if (models) {
                 reduxStore.dispatch(
                   jujuActions.updateModelList({ models, wsControllerURL }),
```

</details>

## Details

- Close #1865
- WD-19967

## Screenshots

image

PR #1879, opened 2025-03-07
refactor(tests): migrate test queries from rebac-admin

andogq:
## Done

- copy queries into `src/testing`
- update a handful of tests to use each available query
- includes suggestion from #1878 which was missed in original PR


huwshimi (approved):
This is great, thanks for migrating those queries.

PR #1890, opened 2025-03-12
fix: flash of charm actions on load

### LINKED ISSUE #1557: select-charms-and-actions panel flickers for a split second
**Describe the bug**
When accessing the `select-charms-and-actions` panel on a slow internet connection, there might be a flicker for a split second showing different content for the panel. In this split second, the details of every charm action is being displayed. Bug is most easily reproduced when setting the network speed to Slow 3G.

**Screenshots**
Content of the panel during the flicker:
Screenshot 2023-08-08 at 17 58 51

Content of the panel after the flicker:
Screenshot 2023-08-08 at 17 58 21

andogq:
## Done

Since the charm action item height was set in JS, there was a moment of time where the default item height was displayed before React kicked in and set it to the correct value.

This fix includes a few changes:

- No more magic values: animation points are pulled out of CSS variables, rather than hard coded pixel values

- Height enforced via CSS: JS is only used for triggering animations, there are no manual overrides of `el.style.height` in JS

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Select an application, and click 'Run Action'

## Details

- Close #1557
- WD-20172

## Screenshots

I've slowed down the recording for some of these which have caused some frames to go weird. This doesn't seem to be the case when actually using it (see the last video).

### Previous

https://github.com/user-attachments/assets/bd2bcd4f-54a4-436c-82bb-6bebce50f90d

### Fixed

https://github.com/user-attachments/assets/3e3ac127-9927-481e-9069-26a1245882ce

### Real-time

https://github.com/user-attachments/assets/014be8d2-b914-40e6-bbd7-b4c31977489e

### DIFF
--- a/src/components/RadioInputBox/RadioInputBox.tsx
+++ b/src/components/RadioInputBox/RadioInputBox.tsx
@@ -20,67 +20,64 @@ export default function RadioInputBox({
   const inputBoxRef = useRef<HTMLDivElement>(null);
   const containerRef = useRef<HTMLDivElement>(null);
 
-  // 20, 40 are magic numbers that align nicely with the heights.
-  const initialHeight = 40;
-  const paddingNumber = 20;
-
-  useEffect(() => {
-    if (inputBoxRef.current !== null) {
-      // Due to the 'closed' hook animation running on initial load we cannot
-      // set the initial height in the css but have to do it on the first load.
-      inputBoxRef.current.style.height = initialHeight + "px";
-    }
-  }, [inputBoxRef]);
+  // Used to track whether this radio box has been opened before. See its usage for more
+  // information.
+  const hasBeenOpened = useRef(false);
 
   useEffect(() => {
     setOpened(selectedInput === name);
   }, [selectedInput, name]);
 
+  // Control the open/close animation based on `opened`. This hook will trigger an animation
+  // between the collapsed and maximum height. Since it's not possible to transition to/from
+  // `initial` height (where the height is set by the content), we have to capture the content's
+  // height and manually set it from JS.
+  //
+  // The 'base' height is controlled in CSS via the `:not([aria-expanded="true"])` selector,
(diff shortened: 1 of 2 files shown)


src/components/RadioInputBox/RadioInputBox.tsx
```diff
@@ -20,67 +20,64 @@ export default function RadioInputBox({
...
-      startHeight = container.offsetHeight + paddingNumber;
-      duration = startHeight;
+      // An open attempt has been made, indicate that this has happened.
+      hasBeenOpened.current = true;
     }
-    const animation = wrapper.animate(
+
+    // Default animation points start collapsed, and expand to content height.
+    const points = [
+      // Collapsed height
```
huwshimi, line 60:
```suggestion
      // Collapsed height.
```

src/components/RadioInputBox/RadioInputBox.tsx
```diff
@@ -20,67 +20,64 @@ export default function RadioInputBox({
...
+      "var(--initial-height)",
+      // Maximum content height (inner content + vertical spacing)
```
huwshimi, line 62:
```suggestion
      // Maximum content height (inner content + vertical spacing).
```

src/components/RadioInputBox/_radio-input-box.scss
```diff
@@ -1,9 +1,28 @@
 .radio-input-box {
-  border: 1px solid $color-mid-light;
+  $border-width: 1px;
+  $v-padding: 0.5rem;
+
+  // This are the default for `label` in Vanilla.
```
huwshimi, line 5:
```suggestion
  // This is the default for `label` in Vanilla.
```

huwshimi (approved):
This works really well. I tried in a few browsers and they all seem to behave. I've made a couple of very small suggestions on the comments.

PR #1894, opened 2025-03-20
refactor: use native Redux store in tests

andogq:
## Done

Refactored tests to remove `redux-mock-store` which has been deprecated. The mock store had the following features which had to be addressed:

- Listing dispatched actions: This was only used by a subset of tests, so a store created with `createStore` exported from `testing/utils` includes middleware that will capture all dispatched actions.
- Reducers were not run: Stores would not deviate from their initial value regardless of any dispatched actions. Since a real store is now being used, some tests need to be updated to correctly handle state changes (most commonly, loading states).

See WD-20358


andogq:
Waiting on #1897 to re-run tests with Node 22

src/testing/utils.tsx
```diff
@@ -151,3 +155,45 @@ export const renderWrappedHook = <Result, Props>(
   });
   return { router, result, store };
 };
+
+export function createStore(
```
huwshimi, line 159: A tiny detail, but there's quite a bit of unnecessary whitespace in this function.

src/testing/utils.tsx
```diff
@@ -151,3 +155,45 @@ export const renderWrappedHook = <Result, Props>(
...
+  state: RootState,
+  opts: { trackActions: true },
```
huwshimi, line 161: We generally try to avoid abbreviated or single character names.

src/testing/utils.tsx
```diff
@@ -151,3 +155,45 @@ export const renderWrappedHook = <Result, Props>(
...
+): [EnhancedStore<RootState>, UnknownAction[]];
+export function createStore(
+  state: RootState,
+  opts?: { trackActions?: boolean },
+): EnhancedStore<RootState>;
```
huwshimi, line 166: I think this return type would need to be `EnhancedStore<RootState> | [EnhancedStore<RootState>, UnknownAction[]]` or else change opts to `opts?: { trackActions: false },`

huwshimi (approved):
Thanks for figuring all that out. A couple of tiny comments on the util.

PR #1902, opened 2025-03-26
refactor: replace async-limiter with a blocking loop

huwshimi:
## Done

- Remove async-limiter.
- Use a blocking loop to limit the concurrency to 1 (which is what async-limiter was used for previously).

## QA

- Go to the model list and check that all your models appear in the list.


src/juju/api.ts
```diff
@@ -344,80 +343,69 @@ export async function fetchAllModelStatuses(
...
-  const queue = new Limiter({ concurrency: 1 });
   let modelErrorCount = 0;
-  modelUUIDList.forEach((modelUUID) => {
-    queue.push(async (done) => {
-      if (isLoggedIn(getState(), wsControllerURL)) {
-        try {
-          const modelWsControllerURL = getModelByUUID(
-            getState(),
+  // Use for/of so that the awaits are blocking so only one model gets polled at a time.
+  for (const modelUUID of modelUUIDList) {
```
huwshimi, line 348: Most of this is indentation changes except for this for loop and I removed the `done()` calls.

andogq (approved):

PR #1908, opened 2025-03-28
feat: implement `OIDCAuth` subclass

andogq:
## Done

Implement `OIDCAuth` subclass.

## Details

- WD-20705

### DIFF
--- a/src/auth/OIDCAuth.test.ts
+++ b/src/auth/OIDCAuth.test.ts
@@ -1,0 +1,110 @@
+import type { Mock } from "vitest";
+
+import * as jimmListeners from "juju/jimm/listeners";
+import * as jimmThunks from "juju/jimm/thunks";
+import { actions as generalActions } from "store/general";
+
+import { Auth } from "./Auth";
+import { OIDCAuth } from "./OIDCAuth";
+
+describe("OIDCAuth", () => {
+  let dispatch: Mock;
+
+  beforeEach(() => {
+    dispatch = vi.fn();
+    new OIDCAuth(dispatch);
+  });
+
+  afterEach(() => {
+    // @ts-expect-error - Resetting singleton for each test run.
+    delete Auth.instance;
+  });
+
+  it("bootstrap", async () => {
+    // Since `OIDCAuth` is created with a mixin, fetch the anonymous class's prototype to spy on.
+    const pollingAuthBootstrapMock = vi.spyOn(
+      Object.getPrototypeOf(OIDCAuth).prototype,
+      "bootstrap",
+    );
+    const addWhoamiListenerMock = vi.spyOn(jimmListeners, "addWhoamiListener");
+    await Auth.instance.bootstrap();
+    expect(addWhoamiListenerMock).toHaveBeenCalledOnce();
+    expect(pollingAuthBootstrapMock).toHaveBeenCalledOnce();
+  });
+
+  it("logout", async () => {
+    const logoutMock = vi.spyOn(jimmThunks, "logout");
+    await Auth.instance.logout();
+    expect(dispatch).toBeCalledTimes(2);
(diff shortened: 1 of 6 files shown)


src/auth/OIDCAuth.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        appThunks,
+        "connectAndStartPolling",
+      );
+    });
+
+    it("connection success", async () => {
+      dispatch.mockResolvedValue({ payload: {} });
+
+      await Auth.instance.bootstrap();
+
```
huwshimi, line 42: A tiny note, but there's a bit of unnecessary whitespace in some of these tests.

src/auth/OIDCAuth.test.ts
```diff
@@ -0,0 +1,145 @@
...
+import { Auth } from "./Auth";
+import { OIDCAuth } from "./OIDCAuth";
+
+describe("OIDCAuth", () => {
+  let dispatch: Mock;
+
+  beforeEach(() => {
+    // @ts-expect-error - Resetting singleton for each test run.
+    delete Auth.instance;
```
huwshimi, line 16: We should probably also call this in `afterAll` so that we don't leave a singleton for future test files (should also do this in the base class tests).

src/auth/OIDCAuth.test.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    dispatch = vi.fn();
+    new OIDCAuth(dispatch);
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
```
huwshimi, line 24: This shouldn't be necessary as we set the config to do this automatically: https://github.com/canonical/juju-dashboard/blob/main/vite.config.ts#L93.

src/auth/OIDCAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+  override async logout(): Promise<void> {
+    this.dispatch(jimmListeners.pollWhoamiStop());
+    await this.dispatch(jimmThunks.logout());
+  }
+
+  override async beforeControllerConnect({
+    wsControllerURL,
+  }: ControllerData): Promise<boolean> {
+    try {
+      // TODO: Hoist loading state into model-poller
```
huwshimi, line 46: With todos like this we usually try to leave the ticket number so we know what they're related to and can find them in the future (also very minor thing but we try to use proper sentences, so start with a capital, end with a full stop).

src/auth/OIDCAuth.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  override async bootstrap(): Promise<void> {
+    jimmListeners.addWhoamiListener(listenerMiddleware.startListening);
+
+    try {
+      // Try and connect automatically.
+      unwrapResult(await this.dispatch(appThunks.connectAndStartPolling()));
+    } catch (error) {
+      logger.error(Label.POLLING_ERROR, error);
+    }
```
huwshimi, line 34: Do you think we should dedupe this somehow as this will also be done for Candid?

src/auth/OIDCAuth.test.ts
andogq: I assume you mean `afterEach`, as the singleton needs to be re-instantiated for each test run.

huwshimi: We could use `afterEach` if you like, but then we'd also need to add a `beforeAll` to make sure we clear the instance before the first test.

src/auth/OIDCAuth.ts
andogq: I've pulled out this logic into a `PollingAuth` class, which `OIDCAuth` (and `CandidAuth`) both extend instead.

huwshimi (approved):
Thanks for the mixin improvements.

PR #1915, opened 2025-04-02
feat: utilise `Auth` methods in place of `AuthMethod`

andogq:
## Done

- Utilise `Auth` methods (WD-20708)
- Update `js-libjuju` to `8.0.1`

### `LocalAuth`

- Ensure local juju multipass is running
- `yarn start`
- http://localhost:8036
- Login with local user account
- Verify dashboard functionality
- Logout of user in bottom left corner

### `OIDCAuth`

- Ensure access to JIMM instance
- `sudo yarn start-jaas --port 443`
- https://jimm-dashboard.k8s.dev.canonical.com
- Login with Google account
- Verify dashboard functionality
- Logout of user in bottom left corner

### `CandidAuth`

- Ensure access to [Juju controller with Candid](https://github.com/canonical/juju-dashboard/blob/main/HACKING.md#juju-controller-with-candid)
- `yarn start`
- http://localhost:8036
- Login with external account
- Verify dashboard functionality
- Logout of user in bottom left corner

## Details

Unfortunately this is a bit of a messy PR as it touches a lot of things.

The majority of the changes relate to tests, as it is now required that _an_ instance of `Auth` is initialised in the singleton for (many) of the tests to run. To deal with this, in some `beforeEach` hooks a specific instance of `Auth` is created based on the requirements of the test (OIDC/Candid/etc). Initialising the `Auth` instance in `setup.ts` seemed to cause a variety of errors across a large selection of tests.

### DIFF
--- a/src/juju/api.ts
+++ b/src/juju/api.ts
@@ -18,18 +18,17 @@ import type { ValueOf } from "@canonical/react-components";
 import { unwrapResult } from "@reduxjs/toolkit";
 import type { Dispatch } from "redux";
 
+import { Auth } from "auth";
 import bakery from "juju/bakery";
 import JIMM from "juju/jimm";
 import { actions as generalActions } from "store/general";
 import {
-  getConfig,
   getControllerConnection,
   getUserPass,
   getWSControllerURL,
   isLoggedIn,
 } from "store/general/selectors";
 import type { Credential } from "store/general/types";
-import { AuthMethod } from "store/general/types";
 import { actions as jujuActions } from "store/juju";
 import { addControllerCloudRegion } from "store/juju/thunks";
 import type {
@@ -74,7 +73,6 @@ export const CLIENT_VERSION = "3.0.0";
   @returns The configuration options.
 */
 export function generateConnectionOptions(
-  authMethod?: AuthMethod,
   usePinger = false,
   onClose: ConnectOptions["closeCallback"] = () => null,
 ) {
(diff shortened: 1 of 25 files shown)


huwshimi (commented):
This is looking good. One general thing is that we should probably add tests that the different Auth.instance.something() methods are called when we expect them to be, as we don't want to be testing the result of those methods for each AuthMethod when they're called throughout the codebase.

src/juju/api.ts
```diff
@@ -95,26 +93,16 @@ export function generateConnectionOptions(
   if (usePinger) {
     facades.push(Pinger);
   }
-  return {
+  const options = {
```
huwshimi, line 96: This could go back to directly return the object.

src/juju/api.ts
```diff
@@ -95,26 +93,16 @@ export function generateConnectionOptions(
...
     bakery,
     closeCallback: onClose,
     debug: false,
     facades,
-    oidcEnabled: authMethod === AuthMethod.OIDC,
+    loginWithSessionCookie: false,
```
huwshimi, line 101: It looks like this is an optional param so we can remove this line I think.

src/store/middleware/model-poller.test.ts
```diff
@@ -166,71 +167,79 @@ describe("model poller", () => {
...
+          controllers: [[wsControllerURL, undefined]],
+          isJuju: true,
+          poll: 0,
+        }),
+      );
+      expect(fakeStore.dispatch).toHaveBeenCalledWith(pollWhoamiStart());
+      expect(loginWithBakerySpy).toHaveBeenCalled();
+    });
+
+    it("handles whoami errors when logging", async () => {
```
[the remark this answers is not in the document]
huwshimi: What did you think about removing this test?

src/testing/utils.tsx
```diff
@@ -76,6 +79,8 @@ export const wrapComponent = (
     options && "store" in options
       ? options.store
       : createStore(options?.state ?? rootStateFactory.build());
+  const config = store.getState().general.config || configFactory.build();
+  initialiseAuthFromConfig(config as WindowConfig, store.dispatch);
```
huwshimi, line 83: `WindowConfig` is defined as `type WindowConfig = Omit<Config, "authMethod">` so once we remove authMethod from config in [WD-20710](https://warthogs.atlassian.net/browse/WD-20710) we can also remove `as WindowConfig` so maybe add a TODO to remember to do that?

src/store/middleware/model-poller.test.ts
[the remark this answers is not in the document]
andogq: Ah my bad, I think I may've missed it in a rebase.

huwshimi (approved):
Looks great!

PR #1916, opened 2025-04-03
feat: hoist login state

andogq:
WD-20709

## Done

- Move spinner logic out of `OIDCForm` to `LogIn`

### `LocalAuth`

- Ensure local juju multipass is running
- `yarn start`
- http://localhost:8036
- Login with local user account

### `OIDCAuth`

- Ensure access to JIMM instance
- `sudo yarn start-jaas --port 443`
- https://jimm-dashboard.k8s.dev.canonical.com
- Login with Google account

### `CandidAuth`

- Ensure access to [Juju controller with Candid](https://github.com/canonical/juju-dashboard/blob/main/HACKING.md#juju-controller-with-candid)
- `yarn start`
- http://localhost:8036
- Login with external account

PR #1918, opened 2025-04-03
feat: remove `AuthMethod`

andogq:
## Done

- remove references to `AuthMethod`

- change references of `WindowConfig` to `Config`

- alter handful of tests to behave without `config.authMethod`

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Make sure everything behaves

## Details

WD-20710

> [!note]
> Builds on #1916, so there's an extra commit included

### DIFF
--- a/src/auth/utils.ts
+++ b/src/auth/utils.ts
@@ -1,5 +1,5 @@
+import type { Config } from "store/general/types";
 import type { AppDispatch } from "store/store";
-import type { WindowConfig } from "types";
 
 import { CandidAuth, LocalAuth, OIDCAuth } from ".";
 
@@ -8,7 +8,7 @@ import { CandidAuth, LocalAuth, OIDCAuth } from ".";
  * implementation.
  */
 export function initialiseAuthFromConfig(
-  config: WindowConfig,
+  config: Config,
   dispatch: AppDispatch,
 ) {
   if (!config.isJuju) {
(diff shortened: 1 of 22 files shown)


src/auth/utils.ts
```diff
@@ -1,14 +1,20 @@
...
-import type { WindowConfig } from "types";
 
 import { CandidAuth, LocalAuth, OIDCAuth } from ".";
 
+export enum AuthMethod {
+  CANDID = "candid",
+  LOCAL = "local",
+  OIDC = "oidc",
+}
+
```
huwshimi, line 11: I think this would be more consistent if it was in a `types.ts` file.

src/index.tsx
```diff
@@ -93,13 +81,7 @@ export const renderApp = () => {
...
 
 function bootstrap() {
-  const windowConfig = window.jujuDashboardConfig;
-  const config = windowConfig
-    ? {
-        ...windowConfig,
-        authMethod: getAuthMethod(windowConfig),
-      }
-    : null;
+  const config = window.jujuDashboardConfig || null;
```
huwshimi, line 84: It looks like this would be handled fine if it was `undefined` so I think we can remove `|| null`.

huwshimi (approved):
A couple of tiny suggestions.

PR #1924, opened 2025-04-07
feat: add JIMM LXD OIDC workflow

andogq:
## Done

- Create E2E workflow with JIMM LXD
- ~~Remove unused input parameter from `e2e-juju-machine-charm-candid-auth.yml` (TBD whether to keep this)~~
- Use huwshimi/jimm as upstream for JIMM action (TODO: remove this once canonical/jimm#1587 is merged)

## QA

https://github.com/canonical/juju-dashboard/actions/runs/14322025289
https://github.com/canonical/juju-dashboard/actions/runs/14322025289/job/40140648008

## Details

WD-20942

### DIFF
--- a/.github/actions/setup-jimm/action.yml
+++ b/.github/actions/setup-jimm/action.yml
@@ -27,7 +27,8 @@ runs:
     - name: Checkout JIMM repo
       uses: actions/checkout@v4
       with:
-        repository: canonical/jimm
+        repository: huwshimi/jimm # TODO: (canonical/jimm#1587) Replace this with `canonica/jimm` when PR is merged
+        ref: add-microk8s-action
         path: ./jimm
     - name: Update dashboard location for JIMM
       shell: bash
--- a/.github/actions/setup-jimm/action.yml
+++ b/.github/actions/setup-jimm/action.yml
@@ -1,0 +1,56 @@
+name: Set up JIMM
+description: Set up JIMM with provided configuration.
+inputs:
+  jimm-version:
+    description: Version of JIMM to use.
+    required: false
+    default: dev
+  juju-channel:
+    description: Channel of Juju to use.
+    required: false
+    default: 3/stable
+  dashboard-location:
+    description: URL of the running dashboard.
+    required: false
+    default: http://127.0.0.1:8000
+  ghcr-secret:
+    description: GHCR secret.
+    required: true
+outputs:
+  controller-name:
+    description: Name of the Juju controller
+    value: ${{ steps.test-server.outputs.controller-name }}
+
+runs:
+  using: composite
+  steps:
+    - name: Checkout JIMM repo
+      uses: actions/checkout@v4
+      with:
+        repository: canonical/jimm
+        path: ./jimm
+    - name: Update dashboard location for JIMM
+      shell: bash
+      run:
+        yq -i '
+          .services.jimm-base.environment.JIMM_DASHBOARD_FINAL_REDIRECT_URL = "${{ inputs.dashboard-location }}" |
+          .services.jimm-base.environment.CORS_ALLOWED_ORIGINS = "${{ inputs.dashboard-location }}" |
+          .services.jimm-base.environment.JIMM_DASHBOARD_LOCATION = "${{ inputs.dashboard-location }}"'
+          "jimm/docker-compose.common.yaml"
+    - name: Setup Go
+      uses: actions/setup-go@v4
+      with:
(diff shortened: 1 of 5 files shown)


.github/actions/setup-jimm/action.yml
```diff
@@ -0,0 +1,62 @@
+name: Set up JIMM
+description: Set up JIMM with provided configuration.
+inputs:
+  jimm-checkout-location:
+    description: Path to checkout the JIMM repository to.
+    required: false
+    default: jimm
```
huwshimi, line 7: Is this here just as a way to store the page as a variable (and same with `dashboard-location`)? If so maybe we should set these as scoped env vars to make it clear these aren't supposed to be passed in: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow.

.github/actions/setup-jimm/action.yml
```diff
@@ -0,0 +1,62 @@
...
+  jimm-version:
+    description: Version of JIMM to use.
+    required: false
+    default: dev
+  juju-channel:
+    description: Channel of Juju to use.
+    required: false
+    default: 3.6/edge
```
huwshimi, line 15: If this workflow passes with "3/stable" maybe that would be a better default (the k8s version has trouble with 3.6/edge).

.github/actions/setup-jimm/action.yml
```diff
@@ -0,0 +1,62 @@
...
+      shell: bash
+      run: cd "${{ inputs.jimm-checkout-location }}" && go mod vendor
+    - name: Start JIMM (pull request)
+      uses: ./jimm/.github/actions/test-server # NOTE: Can't use `inputs.jimm-checkout-location` as it isn't expanded in this field.
+      id: test-server
+      with:
+        jimm-version: ${{ inputs.jimm-version }}
+        juju-channel: ${{ inputs.juju-channel }}
+        ghcr-pat: ${{ inputs.ghcr-secret }}
+        dump-logs: true
```
huwshimi, line 59: I don't think we need this on by default. It will dump the logs if it fails.

.github/workflows/end-to-end.yml
```diff
@@ -13,3 +13,6 @@ jobs:
   juju-k8s-local:
     name: Test Juju, k8s-charm and local-auth
     uses: ./.github/workflows/e2e-juju-k8s-charm-local-auth.yml
+  jimm-machine:
+    name: Test JIMM machine charm
```
huwshimi, line 17: Maybe this should mention OIDC as well to be consistent and also so that remember which runs are testing OIDC when looking through the list of jobs.

huwshimi (approved):

.github/actions/setup-jimm/action.yml
andogq: I did actually intend this to be an input variable, in case `jimm` conflicted with a top-level directory in the `juju-dashboard` repo. I'll change it over to env variables though, and we can just update the action if that's ever the case.

andogq: This actually isn't possible, that `env` property is only for workflows (not composite actions), see [this forum post](https://github.com/orgs/community/discussions/51280) for confirmation. I'll just remove and hard-code all those paths.

.github/workflows/e2e-juju-machine-charm-candid-auth.yml
```diff
@@ -2,12 +2,6 @@ name: "E2E tests: Juju, machine charm and local auth"
 
 on:
   workflow_call:
-    inputs:
-      admin-password:
-        required: false
-        type: string
-        description: The password to use when logging in as the admin user.
-        default: password1
```
Ninfa-Jeon, line 10: We require this as well as the `change-user-password` action that was removed from the workflow previously. That is because we need to `juju login/logout` to share models and such between users.

andogq: No problem, I've dropped that commit and re-added the step to the workflow.

Ninfa-Jeon: Thanks! Also, if my PR for models gets merged before this one, all you have to do is rebase because I'm using this flow haha.

huwshimi:
@andogq I think we can close this now, unless there's anything you still want to land from this PR?

PR #1926, opened 2025-04-07
feat(e2e): add basic auth tests for juju workflows

Ninfa-Jeon:
## Done

- Introduces basic authentication integration tests for all Juju workflows
- Verify the login functionality using username and password for Candid as well as Local Auth
- Implement helpers for authentication
- Inject the helpers as fixtures into the test

## QA

- The authentication tests should pass for all workflows
- Check the test results for all available End-to-End test workflows:
    1. https://github.com/canonical/juju-dashboard/actions/runs/14304424635/job/40085019396?pr=1926
    2. https://github.com/canonical/juju-dashboard/actions/runs/14304424635/job/40085019912?pr=1926
    3. https://github.com/canonical/juju-dashboard/actions/runs/14304424635/job/40085020348?pr=1926

## Details

https://warthogs.atlassian.net/browse/WD-20934

### DIFF
--- a/e2e/fixtures/auth-setup.ts
+++ b/e2e/fixtures/auth-setup.ts
@@ -1,0 +1,16 @@
+import { test as base } from "@playwright/test";
+
+import { AuthHelpers } from "../helpers/auth-helpers";
+
+type AuthFixtures = {
+  authHelpers: AuthHelpers;
+};
+
+export const test = base.extend<AuthFixtures>({
+  authHelpers: async ({ page }, use) => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    await use(new AuthHelpers(page));
+  },
+});
+
+export default test;
--- a/e2e/helpers/auth-helpers.ts
+++ b/e2e/helpers/auth-helpers.ts
@@ -1,0 +1,40 @@
+import { type Page } from "@playwright/test";
+
+export class AuthHelpers {
+  readonly page: Page;
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  async loginLocal() {
+    await this.page.goto("/");
+    await this.page.getByRole("textbox", { name: "Username" }).fill("admin");
+    await this.page
+      .getByRole("textbox", { name: "Password" })
+      .fill("password1");
+    await this.page
+      .getByRole("button", { name: "Log in to the dashboard" })
+      .click();
+  }
+
+  async loginCandid() {
+    await this.page.goto("/");
+    const popupPromise = this.page.waitForEvent("popup");
+    await this.page
+      .getByRole("link", { name: "Log in to the dashboard" })
+      .click();
+    const popup = await popupPromise;
+    await popup.getByRole("link", { name: "static" }).click();
+    await popup.getByRole("textbox", { name: "Username" }).fill("user1");
+    await popup.getByRole("textbox", { name: "Password" }).fill("password1");
+    await popup.getByRole("button", { name: "Login" }).click();
+  }
+
+  async login() {
+    if (process.env.AUTH_MODE === "candid") {
+      await this.loginCandid();
+    } else {
+      await this.loginLocal();
+    }
+  }
+}
--- a/e2e/base/init.spec.ts
+++ b/e2e/base/init.spec.ts
@@ -1,6 +1,6 @@
 import { expect } from "@playwright/test";
 
-import { test } from "../fixtures/auth-setup";
+import { test } from "../fixtures/setup";
 
 test("Log in", async ({ page, authHelpers }) => {
   await authHelpers.login();
--- a/e2e/base/init.spec.ts
+++ b/e2e/base/init.spec.ts
@@ -1,6 +1,11 @@
-import { test, expect } from "@playwright/test";
+import { expect } from "@playwright/test";
 
-test("dashboard loads", async ({ page }) => {
-  await page.goto("/");
-  await expect(page).toHaveTitle(/Juju Dashboard/);
+import { test } from "../fixtures/auth-setup";
+
+test("Log in", async ({ page, authHelpers }) => {
+  await authHelpers.login();
+  await expect(page.getByRole("link", { name: "Models" })).toBeInViewport();
+  await expect(
+    page.getByRole("link", { name: "Controllers" }),
+  ).toBeInViewport();
 });
--- a/e2e/fixtures/setup.ts
+++ b/e2e/fixtures/setup.ts
@@ -1,14 +1,14 @@
+/* eslint-disable react-hooks/rules-of-hooks */
 import { test as base } from "@playwright/test";
 
 import { AuthHelpers } from "../helpers/auth-helpers";
 
-type AuthFixtures = {
+type Fixtures = {
   authHelpers: AuthHelpers;
 };
 
-export const test = base.extend<AuthFixtures>({
+export const test = base.extend<Fixtures>({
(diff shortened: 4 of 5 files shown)


e2e/base/login.spec.ts
huwshimi, line 1: Maybe you could just move this to init.spec.ts as we don't need a test for the login and once we implement our first real test we'll remove init.spect.ts.

e2e/base/login.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { test, expect } from "../fixtures/auth-setup";
+
+test("Log in", async ({ page, authHelpers }) => {
+  let login = authHelpers.loginJujuLocal;
+  if (process.env.AUTH_MODE === "candid") {
+    login = authHelpers.loginJujuCandid;
+  }
```
huwshimi, line 7: Can this happen as part of the fixture so we can just call `login()` in the test and the login method is abstracted, otherwise we'll need this logic in every test.

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  async loginJujuLocal() {
+    await this.page.goto("/");
+    await this.page.getByRole("textbox", { name: "Username" }).fill("admin");
+    await this.page
+      .getByRole("textbox", { name: "Password" })
+      .fill("password1");
```
huwshimi, line 11: We'll need a way to pass the username and password as environment variables from the workflow and also be able to override the credentials on a per-test basis for when we're testing with other users.

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+  async loginJujuCandid() {
+    await this.page.goto("/");
+    const popupPromise = this.page.waitForEvent("popup");
+    await this.page
+      .getByRole("link", { name: "Log in to the dashboard" })
+      .click();
+    const popup = await popupPromise;
+    await popup.getByRole("link", { name: "static" }).click();
+    await popup.getByRole("textbox", { name: "Username" }).fill("user1");
+    await popup.getByRole("textbox", { name: "Password" }).fill("password1");
```
huwshimi, line 26: Same thing here where we want to pass these in.

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+    await popup.getByRole("button", { name: "Login" }).click();
+    popup.getByText("You're logged in as user1");
```
huwshimi, line 28: Do we need this line?

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,34 @@
+import { type Page } from "@playwright/test";
+
+export class AuthHelpers {
+  readonly page: Page;
+
+  constructor(page: Page) {
+    this.page = page;
+  }
+
+  async loginJujuLocal() {
```
huwshimi, line 6: The auth methods aren't tied to the environment, so this can just be:

```suggestion
  async loginLocal() {
```

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+    await this.page.getByRole("textbox", { name: "Username" }).fill("admin");
+    await this.page
+      .getByRole("textbox", { name: "Password" })
+      .fill("password1");
+    await this.page
+      .getByRole("button", { name: "Log in to the dashboard" })
+      .click();
+  }
+
+  async loginJujuCandid() {
```
huwshimi, line 17:
```suggestion
  async loginCandid() {
```

huwshimi (commented):
This is looking good, I've made some suggestions to make this a bit easier to work with.

Can you please also create a new task for the OIDC login so we don't forget about that (the current Jira task will get closed by this PR)?

e2e/base/login.spec.ts
Ninfa-Jeon: will do!

e2e/base/login.spec.ts
Ninfa-Jeon: Makes sense. Moving the logic there.

e2e/helpers/auth-helpers.ts
Ninfa-Jeon: You're right. I will implement that logic once the tests start working well with the fixture.

e2e/helpers/auth-helpers.ts
Ninfa-Jeon: Hmmm...I think I forgot to add an assertion for this one. Will add it. Thanks for noticing.

e2e/helpers/auth-helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+    await popup.getByRole("textbox", { name: "Password" }).fill("password1");
+    await popup.getByRole("button", { name: "Login" }).click();
+  }
+
+  async login() {
+    let login = this.loginLocal;
+    if (process.env.AUTH_MODE === "candid") {
+      login = this.loginCandid;
+    }
+    await login();
```
huwshimi, line 35: I think we can just call the functions above:

```suggestion
    if (process.env.AUTH_MODE === "candid") {
      await this.loginCandid();
    } else {
      await this.loginLocal();
    }
```

e2e/fixtures/auth-setup.ts
```diff
@@ -0,0 +1,18 @@
...
+};
+
+export const test = base.extend<AuthFixtures>({
+  authHelpers: async ({ page }, use) => {
+    const authHelpers = new AuthHelpers(page);
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    await use(authHelpers);
+  },
+});
+export { expect } from "@playwright/test";
```
huwshimi, line 16: I see that the fixture examples show this export but it seems like this might be confusing as we're not actually overriding "expect" and it seems a bit strange that we'd export "expect" but not anything else from the module.

Ninfa-Jeon: it is weird...and somehow...every single article, documentation or issue about this exports "expect"...i think it is probably to not have too many imports???....will try to see what happens when we do remove this and update

Ninfa-Jeon: Looks like we can remove this export. [This](https://github.com/microsoft/playwright/issues/14184) issue confirms that it is only to organize imports in our test files and nothing else.

e2e/helpers/auth-helpers.ts
Ninfa-Jeon: Done.

e2e/helpers/auth-helpers.ts
Ninfa-Jeon: For the change of credentials per test, I am planning to use a different set of fixtures. They will perform the juju-related actions like add-model, add-user, switching users to check access, etc. We can have one source of truth for doing these operations, i.e., the fixture could be used both in local debugging as well as the CI. Leveraging the `execSync`, we should be able to easily change credentials.

Ninfa-Jeon:
> This is looking good, I've made some suggestions to make this a bit easier to work with.
> 
> Can you please also create a new task for the OIDC login so we don't forget about that (the current Jira task will get closed by this PR)?

Here you go: https://warthogs.atlassian.net/browse/WD-21246

huwshimi (approved):

PR #1927, opened 2025-04-07
feat(e2e): implement 'List controllers' test suite

Ninfa-Jeon:
## Done

- Introduced an end-to-end test suite to verify the functionality of listing controllers in the dashboard
- Depends on https://github.com/canonical/juju-dashboard/pull/1926

## QA

- All tests should pass

## Details

https://warthogs.atlassian.net/browse/WD-20947


Ninfa-Jeon:
@huwshimi I think we can keep this test suite under` /base` or `/common` or directly under `/e2e` as this is something we would want to test for all workflows. What do you think?

e2e/juju/controllers.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+  const controllersTab = page.getByRole("link", { name: "Controllers" });
+  await expect(controllersTab).toBeInViewport();
+  await controllersTab.click();
+  await expect(
+    page.getByText("Model status across controllers"),
+  ).toBeInViewport();
+  await expect(
+    page.getByRole("columnheader", { name: "DEFAULT" }),
+  ).toBeInViewport();
+});
```
Ninfa-Jeon, line 16: it would be nice to add another assertion with the controller name. Something like `expect("mycontroller").toBeInViewport()`. Is there a way we could centrally use these names between CI and local? @huwshimi

huwshimi: Yes! You can expose the controller via an environment variable, in fact it's already be available as `CONTROLLER_NAME`.

e2e/juju/controllers.spec.ts
huwshimi, line 1: This test should run for juju and jimm so it can live in `e2e/base`

e2e/juju/controllers.spec.ts
Ninfa-Jeon: Will do the appropriate changes, thanks!

e2e/juju/controllers.spec.ts
Ninfa-Jeon: Noted.

Ninfa-Jeon: Done.

e2e/base/controllers.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  await expect(controllersTab).toBeInViewport();
+  await controllersTab.click();
+  await expect(
+    page.getByText("Model status across controllers"),
+  ).toBeInViewport();
+  await expect(
+    page.getByRole("columnheader", { name: "DEFAULT" }),
+  ).toBeInViewport();
+  if (process.env.CONTROLLER_NAME)
+    await expect(page.getByText(process.env.CONTROLLER_NAME)).toBeInViewport();
```
huwshimi, line 17: This should check if the controller name is in the controllers table otherwise it will pass if the text is found anywhere on the page (which is quite likely when the controller name is "test").

e2e/base/controllers.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  const controllersTab = page.getByRole("link", { name: "Controllers" });
+  await expect(controllersTab).toBeInViewport();
+  await controllersTab.click();
+  await expect(
+    page.getByText("Model status across controllers"),
+  ).toBeInViewport();
+  await expect(
+    page.getByRole("columnheader", { name: "DEFAULT" }),
+  ).toBeInViewport();
+  if (process.env.CONTROLLER_NAME)
```
huwshimi, line 16: We probably want this to fail if the controller name isn't found otherwise the expect will be skipped and the test will pass.

e2e/base/controllers.spec.ts
Ninfa-Jeon: makes sense...will modify the matcher

e2e/base/controllers.spec.ts
Ninfa-Jeon: you're right...

Ninfa-Jeon: Modified this so that if the controller name isn't found in env variables, we throw an error.

huwshimi (approved):
Thanks for the improvements!

PR #1928, opened 2025-04-07
feat(e2e): implement 'List models' test suite

Ninfa-Jeon:
## Done

- Introduced an end-to-end test suite to verify the functionality of listing models in the dashboard
- Tests that:
     - user can view owned models
     - user can view shared models
- Implemented the `JujuHelpers` that is injected as a fixture
- `JujuHelpers` takes care of creating models and users, granting cloud access and sharing models
- Additionally, `jujuHelpers.cleanup()` acts as a teardown after each test to remove any users, models and cloud permissions.

## QA

- All tests should pass
- The list of models should include owned as well as shared models
- View the following to verify the cleanup action:
     - https://github.com/canonical/juju-dashboard/actions/runs/14401627062/job/40388497620?pr=1928
     - https://github.com/canonical/juju-dashboard/actions/runs/14401627062/job/40388498051?pr=1928
     - https://github.com/canonical/juju-dashboard/actions/runs/14401627062/job/40388497258?pr=1928

## Details

https://warthogs.atlassian.net/browse/WD-20946


e2e/juju/models.spec.ts
huwshimi, line 1: This should run for juju and jimm so it should live in `e2e/base`

e2e/juju/models.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+import { test } from "../fixtures/setup";
+
+test("List Models", async ({ page, authHelpers, modelHelpers }) => {
+  await authHelpers.login();
+  await modelHelpers.addModel("foo");
+
+  await expect(
+    page.getByRole("columnheader", { name: "RUNNING" }),
+  ).toBeInViewport();
+  await expect(page.getByText("foo")).toBeInViewport();
```
huwshimi, line 12: This should probably check that this is in the model table so that we know this is the model name.

huwshimi (commented):
Looks like the checks for shared models are missing:

```
Requirements: 
- A model created by the user.
- A model shared with them by another user.

Steps
- Go to the model list.
- Check that their models are in the list.
- Check that models that have been shared with them are in the list.
```

e2e/base/init.spec.ts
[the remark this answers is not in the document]
Ninfa-Jeon: Will do

e2e/juju/models.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../fixtures/setup";
+
+test("List Models", async ({ page, authHelpers, modelHelpers }) => {
+  await authHelpers.login();
+  await modelHelpers.addModel("foo");
```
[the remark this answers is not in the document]
Ninfa-Jeon: Umfortunately at the moment I am commenting it out...I am thinking to add a conditional flow to this because even if I add the model once, this would fail on a rerun locally.

e2e/juju/models.spec.ts
Ninfa-Jeon: Will add the check, thanks!

Ninfa-Jeon: Checking for model name within table grid cells.

e2e/juju/models.spec.ts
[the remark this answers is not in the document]
Ninfa-Jeon: This can be skipped with env variables but I'm more interested in being able to skip this non-idempotent steps on CI. I see that the tests are retried in case they fail which is good for some cases. However in cases where we add a user or a model, we cannot retry those.

Ninfa-Jeon:
> Looks like the checks for shared models are missing:
> 
> ```
> Requirements: 
> - A model created by the user.
> - A model shared with them by another user.
> 
> Steps
> - Go to the model list.
> - Check that their models are in the list.
> - Check that models that have been shared with them are in the list.
> ```

Hi @huwshimi I have added tests for missing cases.

e2e/juju/models.spec.ts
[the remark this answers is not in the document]
huwshimi: It looks like beforeAll/afterAll get run between retries: https://playwright.dev/docs/test-retries.

So for tests that require specific setup we could put them in their own `describe` and calling Juju in beforeAll and then doing cleanup in afterAll.

e2e/helpers/model-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+  ) {
+    execSync(`juju grant ${userName} ${accessType} ${modelName}`);
+  }
+
+  async addSharedModel(modelName: string, userName: string) {
+    const currentUserName = execSync(
+      `juju whoami | grep "User:" | sed 's/^User: //'`,
+    )
+      .toString()
+      .trim();
```
huwshimi, line 76: We should pass in the admin username via an env var (also so we can use that locally) and make sure to log in as them otherwise it will just try to use the last user to log in who might not have permission to add users/models.

e2e/helpers/model-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+    const output = (await this.addUser(userName)).toString();
+    const registrationString = output.split("\n")[2].trim().split(" ")[2];
```
huwshimi, line 78: I'm not sure if this is a whole lot easier, but to save extracting this registration string it looks like you can just set the new user's password and then they can log.

```
juju add-user bob
echo password1 | juju change-user-password bob --no-prompt
juju logout
juju login -u bob
[enter password1]
```

e2e/helpers/model-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+    await this.addCredential();
+    await this.addModel(modelName);
+    await this.grantModelAccess(
+      currentUserName,
+      ModelAccessTypes.READ,
+      modelName,
+    );
+    await this.jujuLogout();
+    await this.jujuLogin(currentUserName, process.env.PASSWORD);
+    execSync("juju switch controller");
```
huwshimi, line 91: I think we should log in as the correct user at the start of this function instead.

e2e/helpers/model-helpers.ts
Ninfa-Jeon: I can try this. Thanks!

e2e/helpers/model-helpers.ts
Ninfa-Jeon: The admin already logged in as a part of previous test so another login might fail

Ninfa-Jeon: Switching between users asked me to switch to a controller

e2e/juju/models.spec.ts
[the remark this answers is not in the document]
Ninfa-Jeon: I tried a beforeAll for logging in previously in this test suite but looks like the fixture for login can only be called from within the test block and not beforeAll. I'm certain I've seen fixtures being called from beforeAll but I suppose they weren't Page Object Model (POM) implementations. Will try to investigate if we can transform our fixture that way.

e2e/helpers/model-helpers.ts
Ninfa-Jeon: i think i may have found a resolution to this...working on it

Ninfa-Jeon: hold on, the switch controller is necessary...the other user is logging out and admin is logging in once again. the admin needs to switch to controller.

e2e/helpers/model-helpers.ts
Ninfa-Jeon: this worked! will be replacing register with change-user-password.

e2e/helpers/model-helpers.ts
Ninfa-Jeon: will do

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  async addSharedModel(modelName: string, userName: string) {
+    await this.addUser(userName);
+    await this.grantCloud(userName, CloudAccessTypes.ADD_MODEL);
+    await this.jujuLogout();
+    await this.jujuLogin(userName, "password2");
+    await this.addCredential();
+    await this.addModel(modelName, userName);
+    await this.grantModelAccess(ModelAccessTypes.READ, modelName);
+    await this.adminLogin();
```
huwshimi, line 112: We don't know which user will be logged in at the start of this function, so I'm not sure we should end the function by logging in as the admin.

For functions that require you to be an admin it might make more sense to log in as the admin at the start of the function, and then we're not relying on previous functions having left everything in a certain state.

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+  ) {
+    this.cleanupStack.push({ resourceName, type, owner });
+  }
+
+  async addModel(
+    modelName: string,
+    owner: string | undefined = process.env.USERNAME,
+  ) {
+    execSync(`juju add-model ${modelName}`);
+    await this.addToCleanupStack(modelName, ResourceType.MODEL, owner);
```
huwshimi, line 47: This is great.

e2e/fixtures/setup.ts
```diff
@@ -2,15 +2,45 @@
...
+export enum ResourceType {
+  MODEL = "MODEL",
+  USER = "USER",
+  CLOUD = "CLOUD",
+}
+
+export type Resource = {
+  resourceName: string | CloudAccessTypes;
+  type: ResourceType;
+  owner: string | undefined;
```
huwshimi, line 31:
```suggestion
  owner?: string;
```

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async addUser(userName: string) {
+    execSync(
+      `juju add-user --controller ${process.env.CONTROLLER_NAME} ${userName}`,
+    );
+    await this.addToCleanupStack(userName, ResourceType.USER, "");
```
huwshimi, line 54:
```suggestion
    await this.addToCleanupStack(userName, ResourceType.USER);
```

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  async cleanup() {
+    console.log("Running Juju cleanup...");
+    await this.adminLogin();
+
+    while (this.cleanupStack.length > 0) {
+      const resource = this.cleanupStack.pop();
+      if (resource) {
+        const { resourceName, type, owner = "" } = resource;
```
huwshimi, line 122:
```suggestion
        const { resourceName, type, owner } = resource;
```

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,138 @@
...
+    execSync("juju logout");
+  }
+
+  async jujuLogin(
+    userName: string | undefined = process.env.USERNAME,
+    password: string | undefined = process.env.PASSWORD,
+  ) {
+    if (!userName || !password) {
+      throw new Error("Cannot login without credentials");
+    }
```
huwshimi, line 77: Maybe for a follow-up PR you could write these env vars to a file before playwright starts so that we know they exist, to save having these checks.

huwshimi (commented):
Thanks for the cleanup actions, this is getting there.

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: Makes sense. Will move this into the test suite instead so that each test is able to decide for itself which user is intended to see the changes

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: Will do!

e2e/helpers/dummy-credentials.yaml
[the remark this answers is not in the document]
Ninfa-Jeon: Okay, so I have changed the dummy-credentials to a template file. We replace the username and password based on what's supplied in arguments and generate a temporary file to be used to add credential. Then we unlink the file.

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,153 @@
+import { execSync } from "child_process";
```
andogq, line 1: Since there's `async` functions everywhere, it may be worthwhile to use an asynchronous version of `exec` in order to free up the event loop whilst waiting for `juju` commands (don't ask my why they haven't put a promise version in like `fs` D:).

```js
import { promisify } from "node:util";
import child_process from "node:child_process";

const exec = promisify(child_process.exec);
```

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  constructor(cleanupStack?: Resource[]) {
+    this.cleanupStack = cleanupStack || [];
+  }
+
+  async removeModel(
+    modelName: string,
+    owner: string | undefined = process.env.USERNAME,
+  ) {
+    execSync(`juju destroy-model ${owner}/${modelName} --force --no-prompt`);
```
andogq, line 22:
```suggestion
    await exec(`juju destroy-model ${owner}/${modelName} --force --no-prompt`);
```

Same with all the other `execSync`...

andogq, line 22: I would wrap these interpolations (and all the other ones in this file) in `'` to prevent spaces or special characters (aside from other `'`) from breaking the shell. This isn't proper sanitisation, but since we're not accepting user input it shouldn't be a massive deal.

```js
await exec(`juju destroy-model '${owner}/${modelName}' --force --no-prompt`)
```

e2e/helpers/juju-helpers.ts
```diff
@@ -0,0 +1,153 @@
...
+    await this.addToCleanupStack(modelName, ResourceType.MODEL, owner);
+  }
+
+  async addUser(userName: string) {
+    execSync(
+      `juju add-user --controller ${process.env.CONTROLLER_NAME} ${userName}`,
+    );
+    await this.addToCleanupStack(userName, ResourceType.USER);
+    execSync(
+      `{ echo password2; echo password2; } | juju change-user-password ${userName}`,
```
andogq, line 57: Should these passwords be moved to env vars anywhere, so they can be injected?

andogq (approved):
Looking good overall! A couple _very_ minor suggestions, so take them or leave them depending on what you think!

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: We are planning this for a future PR. For now, maybe I can move them to a `const` since we are not expecting another secondary user besides John-Doe at the moment.

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: Done! Thanks for noticing the potential issue.

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: Okay so I tried the env vars thing but midway through I got a doubt that I'd like to clarify before moving forward. Closing this for now; will take it in next PR. Thanks!

PR #1938, opened 2025-04-16
feat(e2e): Add test suite for Juju Web CLI

Ninfa-Jeon:
## Done

- Added test suite for Juju Web CLI

## QA

- All tests should pass

## Details

https://warthogs.atlassian.net/browse/WD-20953

### DIFF
--- a/e2e/base/models.spec.ts
+++ b/e2e/base/models.spec.ts
@@ -2,28 +2,30 @@ import { expect } from "@playwright/test";
 
 import { test } from "../fixtures/setup";
 
-test.beforeAll(async ({ jujuHelpers }) => {
-  await jujuHelpers.jujuLogin();
-  await jujuHelpers.addModel("foo");
-  await jujuHelpers.addSharedModel("bar", "John-Doe");
-  await jujuHelpers.adminLogin();
-});
+test.describe("Models", () => {
+  test.beforeAll(async ({ jujuHelpers }) => {
+    await jujuHelpers.jujuLogin();
+    await jujuHelpers.addModel("foo");
+    await jujuHelpers.addSharedModel("bar", "John-Doe");
+    await jujuHelpers.adminLogin();
+  });
 
-test.afterAll(async ({ jujuHelpers }) => {
-  await jujuHelpers.cleanup();
-});
+  test.afterAll(async ({ jujuHelpers }) => {
+    await jujuHelpers.cleanup();
+  });
 
-test("List created and shared models", async ({ page, authHelpers }) => {
-  await authHelpers.login();
+  test("List created and shared models", async ({ page, authHelpers }) => {
+    await authHelpers.login();
 
-  await expect(
-    page
-      .locator("tr", { hasText: "foo" })
-      .and(page.locator("tr", { hasText: "admin" })),
-  ).toBeInViewport();
-  await expect(
-    page
-      .locator("tr", { hasText: "bar" })
-      .and(page.locator("tr", { hasText: "John-Doe" })),
-  ).toBeInViewport();
+    await expect(
+      page
+        .locator("tr", { hasText: "foo" })
+        .and(page.locator("tr", { hasText: "admin" })),
+    ).toBeInViewport();
+    await expect(
+      page
+        .locator("tr", { hasText: "bar" })
+        .and(page.locator("tr", { hasText: "John-Doe" })),
+    ).toBeInViewport();
+  });
 });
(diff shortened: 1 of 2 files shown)


e2e/base/models.spec.ts
```diff
@@ -2,28 +2,30 @@ import { expect } from "@playwright/test";
 
 import { test } from "../fixtures/setup";
 
-test.beforeAll(async ({ jujuHelpers }) => {
-  await jujuHelpers.jujuLogin();
-  await jujuHelpers.addModel("foo");
-  await jujuHelpers.addSharedModel("bar", "John-Doe");
-  await jujuHelpers.adminLogin();
-});
+test.describe("Models", () => {
```
Ninfa-Jeon, line 5: not putting the `beforeAll` and `afterAll` within a `describe` block was somehow having the models leak between different `spec` files, especially for k8s-charm that uses clusters.

e2e/juju/web-cli.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../fixtures/setup";
+
+test("Web CLI", async ({ page, jujuHelpers, authHelpers }) => {
+  await jujuHelpers.jujuLogin();
+  await jujuHelpers.addModel("foo");
```
Ninfa-Jeon, line 7: keeping the initial setup and cleanup for models within the `test` block since this suite will only contain one test and we should avoid `beforeAll` and `afterAll` blocks outside of `describe` in case they break other test suites.

Ninfa-Jeon:
For some reason, the `candid` flow doesn't allow me to run commands on Web CLI.  It gives me an `Unable to authenticate.` error be it local debugging or the CI.

huwshimi:
> For some reason, the `candid` flow doesn't allow me to run commands on Web CLI. It gives me an `Unable to authenticate.` error be it local debugging or the CI.

Interesting if it isn't working locally then it could be an actual bug or maybe the web CLI websocket can't be authenticated with Candid.

Ninfa-Jeon:
> > For some reason, the `candid` flow doesn't allow me to run commands on Web CLI. It gives me an `Unable to authenticate.` error be it local debugging or the CI.
> 
> Interesting if it isn't working locally then it could be an actual bug or maybe the web CLI websocket can't be authenticated with Candid.

Could be the case. I will skip the test for Candid auth in the case for now

e2e/juju/web-cli.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+import { test } from "../fixtures/setup";
+
+test("Web CLI", async ({ page, authHelpers }) => {
+  // Skipping candid auth as web CLI websocket can't be authenticated with it.
+  test.skip(process.env.AUTH_MODE === "candid");
+
+  await authHelpers.login();
+  await page
+    .locator('[href*="/models/admin/controller"]', { hasText: "controller" })
+    .click();
```
huwshimi, line 12: You could just open the page directly: `await this.page.goto("/models/admin/controller");`.

huwshimi (approved):
One small suggestion.

e2e/juju/web-cli.spec.ts
Ninfa-Jeon: Was trying this originally, but with the current authHelper, it was redirecting to `/models` and asking to authenticate again.

Ninfa-Jeon: That kind of ended up turning into a loop and failing the test. Probably for the future, will change authHelper a bit.

Ninfa-Jeon:
> > For some reason, the `candid` flow doesn't allow me to run commands on Web CLI. It gives me an `Unable to authenticate.` error be it local debugging or the CI.
> 
> Interesting if it isn't working locally then it could be an actual bug or maybe the web CLI websocket can't be authenticated with Candid.

After verifying with Ubuntu SSO Provider, I can confirm that this is happening with Candid on both flows. [Here](https://github.com/canonical/candid/issues/202) is the issue I raised for this.

huwshimi:
> > > For some reason, the `candid` flow doesn't allow me to run commands on Web CLI. It gives me an `Unable to authenticate.` error be it local debugging or the CI.
> > 
> > 
> > Interesting if it isn't working locally then it could be an actual bug or maybe the web CLI websocket can't be authenticated with Candid.
> 
> After verifying with Ubuntu SSO Provider, I can confirm that this is happening with Candid on both flows. [Here](https://github.com/canonical/candid/issues/202) is the issue I raised for this.

OK good to know, I think it's a bug in the dashboard so we should file it on the dashboard repo instead.

PR #1939, opened 2025-04-21
feat(e2e): Add test case for model access

Ninfa-Jeon:
## Done

- Added test case for verifying that a user cannot view a model without access

## QA

- All tests should pass

## Details

https://warthogs.atlassian.net/browse/WD-20951

### DIFF
--- a/e2e/helpers/auth-helpers.ts
+++ b/e2e/helpers/auth-helpers.ts
@@ -10,7 +10,6 @@ export class AuthHelpers {
     userName: string = process.env.USERNAME ?? "",
     password: string = process.env.PASSWORD ?? "",
   ) {
-    await this.page.goto("/");
     await this.page.getByRole("textbox", { name: "Username" }).fill(userName);
     await this.page.getByRole("textbox", { name: "Password" }).fill(password);
     await this.page
@@ -22,7 +21,6 @@ export class AuthHelpers {
     userName: string = "user1",
     password: string = "password1",
   ) {
-    await this.page.goto("/");
     const popupPromise = this.page.waitForEvent("popup");
     await this.page
       .getByRole("link", { name: "Log in to the dashboard" })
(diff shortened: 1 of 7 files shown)


e2e/helpers/auth-helpers.ts
```diff
@@ -10,7 +10,6 @@ export class AuthHelpers {
     userName: string = process.env.USERNAME ?? "",
     password: string = process.env.PASSWORD ?? "",
   ) {
-    await this.page.goto("/");
```
Ninfa-Jeon, line 13: This needs to be done case-wise to leverage `page.goto` as we like. In these tests, everytime we use a `page.goto`, a login screen pops up. Ideally, simply logging in once should suffice but since we were going to `"/"` before every login, it created a loop.

e2e/juju/web-cli.spec.ts
```diff
@@ -6,10 +6,8 @@ test("Web CLI", async ({ page, authHelpers }) => {
   // Skipping candid auth as web CLI websocket can't be authenticated with it.
   test.skip(process.env.AUTH_MODE === "candid");
 
+  await page.goto("/models/admin/controller");
   await authHelpers.login();
-  await page
-    .locator('[href*="/models/admin/controller"]', { hasText: "controller" })
-    .click();
```
Ninfa-Jeon, line 12: we no longer need to do this as we are now able to use `page.goto`

package.json
```diff
@@ -83,7 +83,7 @@
     "@eslint/compat": "1.2.7",
     "@eslint/eslintrc": "3.3.1",
     "@eslint/js": "9.23.0",
-    "@playwright/test": "1.51.1",
+    "@playwright/test": "1.52.0",
```
Ninfa-Jeon, line 86: Bumped playwright to resolve version discrepancy of downloaded browsers for Playwright causing CI to fail. [Here](https://github.com/microsoft/playwright/issues/4033#issuecomment-900629454) is the issue.

Ninfa-Jeon:
I can skip the test for `candid` auth if we don't have a way to add more users to the list of known users at the moment.

huwshimi (commented):
Thanks @Ninfa-Jeon, I think the page.goto change is a good idea.

Feel free to add the skip for candid and then reenable it as part of the work to manage candid/jimm users.

huwshimi (approved):

PR #1940, opened 2025-04-22
feat: test auth abstraction

andogq:
## Done

- Remove references to `SECONDARY_USER` - this is handled by the auth implementations
- Update `setup-candid` action with option to merge config during setup
- Add docs about accessing playwright UI in github action
- Use `ActionStack` to handle running/rolling back stateful actions
- Implement Local and Candid auth

## Details

- WD-21540
- https://github.com/canonical/juju-dashboard/actions/runs/14771198514


e2e/helpers/user/candid.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (id !== 0) {
+          throw new Error("only admin with ID 0 available in Juju");
+        }
+
+        const users = await Candid.getConfigUsers();
+        const dashboardSecret = users.find((user) => user.username === "admin");
+
+        if (!dashboardSecret) {
+          throw new Error(
+            "no user with username `admin` provided in canid config",
```
Ninfa-Jeon, line 57:
```suggestion
            "no user with username `admin` provided in candid config",
```

e2e/helpers/user/juju.ts
```diff
@@ -0,0 +1,213 @@
...
+    await page.getByRole("textbox", { name: "Password" }).fill(password);
+    await page.getByRole("button", { name: "Log in to the dashboard" }).click();
+    return page;
+  }
+
+  async loginCli(user: Credential) {
+    const { username, password } = user.cli();
+
+    await exec("juju logout");
+    await exec(`echo '${password}' | juju login -u '${username}' --no-prompt`);
```
Ninfa-Jeon, line 37: for a scenario where a login URL is printed to the console, my idea is to read it as the output of `exec` and make playwright open it using `page.goto`. I figured that simply navigating to a candid login page and logging in as the intended user does not have the same effect as logging in with the URL returned as a result of `juju login command`

e2e/helpers/juju-helpers.ts
```diff
@@ -117,17 +100,38 @@ export class JujuHelpers {
...
-    await this.addCredential(userName, "password2");
-    await this.addModel(modelName, userName);
-    await this.grantModelAccess(ModelAccessTypes.READ, modelName);
+  async addSharedModel(modelName: string, user: Credential) {
+    const admin = await UserStore.get.getCredential(UserKind.Admin);
+
+    await this.jujuLogin(admin);
+    await this.grantCloud(user.cli().username, CloudAccessType.ADD_MODEL);
+    await this.jujuLogin(user);
+    // TODO: Is this needed for regular juju?
```
Ninfa-Jeon, line 127: yes! a new user cannot add models until their credential is added. The dummy-credentials file takes care of this for both `localhost` and `microk8s` clouds

e2e/helpers/juju-helpers.ts
```diff
@@ -117,17 +100,38 @@ export class JujuHelpers {
...
-    await this.grantCloud(userName, CloudAccessTypes.ADD_MODEL);
-    await this.jujuLogout();
-    await this.jujuLogin(userName, "password2");
-    await this.addCredential(userName, "password2");
-    await this.addModel(modelName, userName);
-    await this.grantModelAccess(ModelAccessTypes.READ, modelName);
+  async addSharedModel(modelName: string, user: Credential) {
+    const admin = await UserStore.get.getCredential(UserKind.Admin);
+
+    await this.jujuLogin(admin);
```
Ninfa-Jeon, line 124: do we still have any utility for `adminLogin` after this change? or is that a better choice instead of this? I will leave the choice up to you for this one

andogq:
To run Playwright UI from the action:

- Start the UI server: `yarn playwright test --ui --ui-port 1234`

- Forward the port (standard SSH port forward doesn't work with tmate): `ssh -R 80:localhost:1234 no-key@localhost.run`

e2e/candid-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+      admin:
+        name: Admin
+        email: admin@example.com
+        password: password
+        groups: []
+      user0:
+        name: User 0
+        email: user0@example.com
+        password: password0
+        groups: []
```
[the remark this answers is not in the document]
andogq: Any users we put here will be merged into the candid config before it runs! So either we can add all the users here that we may need, or we can investigate an alternate provider.

e2e/helpers/user/candid.ts
andogq: 🙏

e2e/helpers/juju-helpers.ts
andogq: Perfect, thank you for the clarification!

e2e/helpers/juju-helpers.ts
andogq: I originally got rid of it as I didn't want it to 'feel' like the admin is different from regular users. Instead, a user is just a value that we can pass around, if that makes sense.

e2e/helpers/user/juju.ts
andogq: I think this is a good plan. I'll investigate it today!

e2e/candid-config.yaml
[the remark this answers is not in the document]
Ninfa-Jeon: Makes sense!

e2e/helpers/juju-helpers.ts
Ninfa-Jeon: It does! In case `adminLogin` is no more used, you can clean it up.

andogq: Good point, all gone now!

e2e/helpers/user/juju.ts
andogq: Good news: this worked! This is it if you're interested https://github.com/canonical/juju-dashboard/pull/1940/files#diff-f9579a5c0d4752ecff5a36ed67e4c0ec0c773e58053eccf4fed4fba60f490bddR88-R120

e2e/helpers/auth/backends/oidc.ts
```diff
@@ -14,6 +18,53 @@ type Secret = {
...
+
+  debug(): string {
+    return "mocked OIDC user";
+  }
+  result(): OidcUser {
+    return new OidcUser(this.username, this.password);
+  }
+}
+
+export class OidcUser extends LocalUser {
```
huwshimi, line 43: Currently we're quite tied to OIDC = JIMM/jimmctl in our implementation. If/when OIDC is an authentication option outside of JIMM we may need to revisit some of this to enable OIDC with Juju scenarios.

e2e/helpers/auth/backends/Local.ts
```diff
@@ -0,0 +1,76 @@
...
+    await exec(
+      `{ echo ${this.password}; echo ${this.password}; } | juju change-user-password '${this.username}'`,
+    );
+  }
+
+  async rollback(jujuCli: JujuCli) {
+    await jujuCli.loginLocalCliAdmin();
+    await exec(
+      `juju remove-user --yes --quiet --controller '${jujuCli.controller}' '${this.username}'`,
+    ).catch(() => {});
```
huwshimi, line 35: Is this catch here in case the user doesn't exist? If so it might be worth leaving a comment.

e2e/utils/juju.ts
```diff
@@ -0,0 +1,7 @@
+import { exec } from ".";
+
+export async function login(username: string, password: string) {
+  // TODO: Don't do this when OIDC is fixed
```
huwshimi: Is this TODO for the OIDC implementation?

huwshimi (commented):
This is great, other than the note about cleaning up for afterEach vs afterAll I've mostly just made some suggestions for code consistency.

I think the use of (sub)classes here makes this all quite clear.

e2e/base/models.spec.ts
```diff
@@ -1,55 +1,62 @@
 import { expect } from "@playwright/test";
 
 import { test } from "../fixtures/setup";
+import { ActionStack } from "../helpers/action";
+import { AddModel, GrantModelAccess } from "../helpers/actions";
+import type { User } from "../helpers/auth";
+import { ModelGrantPermission, type Model } from "../helpers/objects";
+
+// TODO: Replace with model randomisation
```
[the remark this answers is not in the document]
Ninfa-Jeon: I will move the code here and close my PR if that helps.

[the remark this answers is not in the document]
andogq: Yes, although it'd likely be much simpler now. Just need to generate a random string in the constructor, instead of taking it in as a parameter. I would add it into this PR (and can if needed), but there's already so much happening and I wanted to limit (even more) creep.

e2e/base/models.spec.ts
```diff
@@ -1,55 +1,62 @@
...
+      sharedModel = add(new AddModel(user2, SHARED_MODEL));
+
+      add(new GrantModelAccess(sharedModel, user1, ModelGrantPermission.Read));
+    });
   });
 
-  test.afterAll(async ({ jujuHelpers }) => {
-    await jujuHelpers.cleanup();
+  test.afterAll(async () => {
+    await actions.rollback();
```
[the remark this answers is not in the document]
andogq: I'd likely advocate for multiple `ActionStack` instances, and use the built-in `test.beforeAll`/`test.beforeEach` methods.

```ts
let describeActions: ActionStack;
let testActions: ActionStack;

test.beforeAll(async () => {
    describeActions = new ActionStack(...);
    await describeActions.prepare((add) => {
        // TODO: Set up describe here
    });
});

test.beforeEach(async () => {
    testActions = new ActionStack(...);
    await testActions.prepare((add) => {
        // TODO: Set up test here
    });
});

// Tests...

test.afterEach(async () => {
    await testActions.rollback();
});

test.afterAll(async () => {
    await describeActions.rollback();
});
```

e2e/helpers/actions/grantModelAccess.ts
```diff
@@ -0,0 +1,38 @@
...
+    private user: User,
+    private access: ModelGrantPermission,
+  ) {}
+
+  async run(_: JujuCli) {
+    await this.model.owner.cliLogin();
+
+    await exec(
+      `juju grant '${this.user.cliUsername}' '${this.access}' '${this.model.name}'`,
+    );
```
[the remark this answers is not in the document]
andogq: Yeah I'm a bit of a fiend for the enter key it seems...

e2e/base/models.spec.ts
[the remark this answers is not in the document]
huwshimi: Ah yes, good idea!

e2e/helpers/auth/backends/oidc.ts
andogq: Very true, I can imagine we'd need to break out a lot of this into OIDC provider-specific implementations...

e2e/helpers/auth/backends/Local.ts
andogq: It (hopefully) shouldn't be needed anymore, everything should be kept in sync now.

e2e/utils/juju.ts
andogq: I've added the ticket number!

e2e/helpers/auth/backends/Candid.ts
```diff
@@ -102,9 +100,13 @@ export class CandidUser extends LocalUser {
     // Begin the login
     const loginProc = exec("juju login");
 
+    if (loginProc.child.stderr === null) {
```
huwshimi, line 103: Can this also be other falsy values e.g. undefined?

huwshimi (approved):

e2e/base/models.spec.ts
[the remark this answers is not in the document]
Ninfa-Jeon: Maybe for a future PR. We can keep the TODO here till then.

e2e/helpers/auth/backends/Candid.ts
andogq: Hard to say actually. The LSP type has `stderr: Stream.Readable`, but [the docs](https://nodejs.org/api/child_process.html#subprocessstderr) also say that it can be `null` or `undefined` if the process can't be spawned. I'll fix it up anyway.

e2e/helpers/auth/backends/oidc.ts
```diff
@@ -14,8 +18,51 @@ type Secret = {
 // anything, it does not need to match the real controller names.
 const JIMM_CONTROLLER = "jimm-k8s";
 
+export class CreateOidcUser implements Action<OidcUser> {
```
Ninfa-Jeon, line 21: I wonder if the following suggestion sits well with the conventions we currently use. I leave it up to you to decide if we need to make this change :) 
```suggestion
export class CreateOIDCUser implements Action<OIDCUser> {
```

PR #1942, opened 2025-04-22
chore(ci): add JIMM workflow

huwshimi:
## Done

- Add a workflow for JIMM.

## Details

https://warthogs.atlassian.net/browse/WD-20941


andogq (approved):
Nice work! I've left one suggestion about moving the device login from a playwright test into a regular script, but that's up to you whether you think it's worthwhile or not!

.github/actions/setup-jimm/action.yml
```diff
@@ -0,0 +1,173 @@
...
+        # which will remain as blocked as we're not using an external identity provider.
+        juju wait-for model iam --query='forEach(applications, app => (app.name == "kratos-external-idp-integrator" && app.status=="blocked") || (app.name != "kratos-external-idp-integrator" && app.status=="active"))' || true # continue if this gets stuck because it stops getting status updates and times out
+        juju offer hydra:oauth
+        juju offer self-signed-certificates:send-ca-cert
+    - name: Create admin user
+      shell: bash
+      run: |
+        juju config kratos enforce_mfa=False
+        juju wait-for unit kratos/0 || true # continue if this gets stuck because it stops getting status updates and times out
+        IDENTITY_ID=$(juju run --format=json kratos/0 create-admin-account email='${{ inputs.admin-username }}' password=${{ inputs.admin-password }} username=admin | yq .kratos/0.results.identity-id)
```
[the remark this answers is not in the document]
huwshimi: My assumption here was that by setting the username to the same name as the `admin` user that was created when bootstrapping Juju, that these two users were being associated, but I can't actually see anything in the docs to support that.

We log in using the `email` (`inputs.admin-username`) and it seems like `username` just needs to be a unique value (but we don't use it anywhere). I don't think it matters what we use so I've changed it to `inputs.admin-username`.

.github/actions/setup-jimm/cli-login/login.spec.ts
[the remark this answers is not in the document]
huwshimi: Done.

.github/actions/setup-jimm/action.yml
```diff
@@ -0,0 +1,173 @@
...
+        # Start the login. This will output the device code to a file and puts the login process into the background while it waits for us to log in.
+        juju login test-jimm.local:443 -c jimm-k8s >~/login.txt 2>&1 &
+        # Wait for the login message to be output.
+        timeout 1m grep -m 1 'enter code' <(tail -f ~/login.txt)
+        # Get the device code from th efile.
+        export DEVICE_CODE=$(grep -oP "(?<=code\ )(.+)(?=\ to)" ~/login.txt); echo "Device code: $DEVICE_CODE"
+        # Use Playwright to step through the login flow.
+        yarn playwright test -c ${{ github.workspace }}/.github/actions/setup-jimm/cli-login
+        # Wait for the login process to complete.
+        timeout 1m grep -m 1 Welcome <(tail -f ~/login.txt)
```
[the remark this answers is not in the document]
huwshimi: Done.

andogq (approved):

.github/actions/setup-jimm/action.yml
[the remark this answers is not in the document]
andogq: Thanks for the clarification, that seems reasonable to me!

.github/actions/setup-jimm/cli-login.ts
```diff
@@ -0,0 +1,15 @@
+import { OIDC } from "../../../e2e/helpers/user/oidc";
+
+const username = process.env.USERNAME;
+const password = process.env.PASSWORD;
```
Ninfa-Jeon, line 4: We are no longer using these environment variables. Check: https://github.com/canonical/juju-dashboard/pull/1943#discussion_r2062910702
```suggestion
const username = process.env.ADMIN_USERNAME;
const password = process.env.ADMIN_PASSWORD;
```

.github/actions/setup-machine-charm/action.yml
```diff
@@ -21,6 +21,14 @@ inputs:
...
     description: Git reference to use when cloning charm repository.
     required: false
+  controller-app:
+    required: false
+    description: The Juju controller application.
+    default: controller
+  controller-model:
+    required: false
+    description: The model containing the Juju controller.
+    default: controller
```
Ninfa-Jeon, line 31: I wonder if this change was needed as currently we don't have a JIMM with machine charm workflow

e2e/helpers/auth-helpers.ts
```diff
@@ -40,9 +40,25 @@ export class AuthHelpers {
...
+    password: string = "test",
+  ) {
+    await this.page.goto("/");
+    await this.page
+      .getByRole("link", { name: "Log in to the dashboard" })
+      .click();
+    await this.page.getByRole("textbox", { name: "Email" }).fill(userName);
+    await this.page.getByRole("textbox", { name: "Password" }).fill(password);
+    await this.page.getByRole("button", { name: "Sign in" }).click();
+    return this.page;
```
Ninfa-Jeon, line 54: We don't need to return the `page` object as it will be available to the `test` function. It was impossible to get a hold of the `popup` element in case of candid flow and as it was required to invalidate a faulty auth attempt, we were returning it. I doubt we need this.

e2e/helpers/user/oidc.ts
```diff
@@ -0,0 +1,65 @@
...
+  username: string;
+  password: string;
+};
+
+// The name to use to identify the authenticated jimm controller. This can be
+// anything, it does not need to match the real controller names.
+const JIMM_CONTROLLER = "jimm-k8s";
+
+export class OIDC {
+  static async loginCli(user: Secret): Promise<void> {
```
Ninfa-Jeon, line 18:
```suggestion
  static async loginCLI(user: Secret): Promise<void> {
```

.github/actions/setup-jimm/cli-login.ts
huwshimi: These are different from the admin user so I think it's fine to keep as different names, they're only used by this script: https://github.com/canonical/juju-dashboard/blob/main/.github/actions/setup-jimm/cli-login.ts#L3 which is used to log in so we can add the JIMM workloads controller as part of the JIMM set up.

.github/actions/setup-machine-charm/action.yml
huwshimi: You're right, I added this for consistency with the k8s action, but we don't actually need this so we could remove it.

e2e/helpers/auth-helpers.ts
huwshimi: Ah OK, we can remove that then. Thanks!

e2e/helpers/user/oidc.ts
huwshimi: Thanks, I think there are a few things that will need to be updated to match the new names in Tom's PR. We might as well wait until that lands and we extend the base class which will complain about the different names.

.github/actions/setup-jimm/cli-login.ts
Ninfa-Jeon: Understood. Thanks for the clarification!

PR #1948, opened 2025-04-29
fix: don't trigger e2e from other labels

huwshimi:
## Done

- Fix an issue where changing an unrelated label would trigger the e2e workflows if the PR already had the `e2e` label applied.

## Details

https://warthogs.atlassian.net/browse/WD-21780


andogq (approved):
Shame that those conditions can't be hoisted in workflows. Not the end of the world though!

PR #1952, opened 2025-05-01
chore(ci): add e2e tests for accessing audit logs

huwshimi:
## Done

- Test that audit logs can be viewed on the top level page and in model details.

## Details

https://warthogs.atlassian.net/browse/WD-20955

### DIFF
--- a/e2e/jimm/audit-logs.spec.ts
+++ b/e2e/jimm/audit-logs.spec.ts
@@ -1,0 +1,63 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../fixtures/setup";
+import { ActionStack } from "../helpers/action";
+import { GiveControllerAccess, AddModel } from "../helpers/actions";
+import type { User } from "../helpers/auth";
+import type { Model } from "../helpers/objects";
+import { ControllerPermission } from "../helpers/objects";
+
+test.describe("audit logs", () => {
+  let actions: ActionStack;
+  let user: User;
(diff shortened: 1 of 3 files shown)


e2e/jimm/audit-logs.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    await expect(page.getByRole("tab", { name: "Audit logs" })).toHaveAttribute(
+      "aria-selected",
+      "true",
+    );
+    await expect(
+      page.locator("td", { hasText: user.dashboardUsername }).first(),
+    ).toBeVisible();
+    await expect(
+      page.locator("td", { hasText: "less than a minute ago" }).first(),
+    ).toBeVisible();
```
Ninfa-Jeon, line 61: if these are going to be in the same row, we could do something like this:
```suggestion
    await expect(
      page
        .locator("td", { hasText: user.dashboardUsername })
        .and(page.locator("td", { hasText: "less than a minute ago" })),
    ).toBeVisible();
```

Ninfa-Jeon: we could do this for the checks above as well

Ninfa-Jeon (approved):
LGTM

e2e/jimm/audit-logs.spec.ts
huwshimi: Thanks!

PR #1956, opened 2025-05-05
feat(e2e): Update test suite for models

Ninfa-Jeon:
## Done

- Updated the `models.spec.ts` test suite
- The test to verify list of created + shared models is now run from a secondary user's perspective that does not have superuser permissions
- Removed hard-coded model names and added incrementing model name.

## Details

https://warthogs.atlassian.net/browse/WD-21920

### DIFF
--- a/e2e/base/models.spec.ts
+++ b/e2e/base/models.spec.ts
@@ -6,20 +6,15 @@ import { AddModel, GrantModelAccess } from "../helpers/actions";
 import type { User } from "../helpers/auth";
 import { ModelGrantPermission, type Model } from "../helpers/objects";
 
-const MODEL1 = "models-foo1";
-const MODEL2 = "models-foo2";
-const SHARED_MODEL = "models-bar";
-
 test.describe("Models", () => {
   // TODO: implement OIDC fixtures WD-21779.
-  // Skipping non-local auth tests: Only admin login supported for Candid/OIDC currently.
-  test.skip(process.env.AUTH_MODE !== "local");
+  test.skip(process.env.AUTH_MODE === "oidc");
 
   let actions: ActionStack;
   let user1: User;
   let user2: User;
   let user1Model: Model;
-  let user1Model2: Model;
+  let sharedModel: Model;
   let user2Model: Model;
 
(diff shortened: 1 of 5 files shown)


e2e/helpers/actions/addModel.ts
```diff
@@ -9,11 +9,7 @@ import { Model } from "../objects";
 export class AddModel implements Action<Model> {
   public model: Model;
 
-  constructor(
-    owner: User,
-    // TODO: Generate random name
-    modelName: string,
-  ) {
+  constructor(owner: User, modelName: string = generateRandomModelName()) {
```
andogq, line 12: I think we should make it always randomly generate a name, remove the option to provide a custom name.

e2e/base/models.spec.ts
```diff
@@ -6,18 +6,17 @@ import { AddModel, GrantModelAccess } from "../helpers/actions";
 import type { User } from "../helpers/auth";
 import { ModelGrantPermission, type Model } from "../helpers/objects";
 
-const MODEL = "models-foo";
-const SHARED_MODEL = "models-bar";
-
 test.describe("Models", () => {
   // TODO: implement OIDC fixtures WD-21779.
-  test.skip(process.env.AUTH_MODE === "oidc");
+  // Skipping non-local auth tests: Only admin login supported for Candid/OIDC currently.
```
andogq, line 11: What do you mean by this? We can just create as many users as we want, and log in as any of them.

Ninfa-Jeon: simply moved the case-specific skip to the describe block since they were both using same login. if it works then its not needed below either...have removed it to see if it works

e2e/juju/model-access-control.spec.ts
```diff
@@ -30,8 +28,8 @@ test.describe("Model Access Control", () => {
   });
 
   test("Can change model permissions", async ({ page }) => {
-    // Skipping non-local auth tests: Only admin login supported for Candid/OIDC currently.
```
Ninfa-Jeon, line 33: this is no longer an issue with candid

PR #1957, opened 2025-05-05
fix: trigger e2e tests on PR creation if label is present

Ninfa-Jeon:
## Done

- Updated the e2e workflow to trigger when a PR is initially created and it already has the 'e2e' label applied.

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -20,10 +20,10 @@ on:
         required: true
         type: choice
         options:
-        - branch
-        - release
-        - charm
-        - bundled
+          - branch
+          - release
+          - charm
+          - bundled
       source-identifier:
         description: |
           Identifier for the dashboard source:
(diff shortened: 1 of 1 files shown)


.github/workflows/end-to-end.yml
```diff
@@ -57,28 +57,32 @@ jobs:
     if: |
       github.event_name != 'pull_request' || (
         (github.event.action == 'labeled' && github.event.label.name == 'e2e') ||
-        (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'e2e')))
+        (github.event.action == 'opened' && contains(github.event.pull_request.labels.*.name, 'e2e'))
```
huwshimi, line 60: Did this change make the tests run when you created the PR? It looks like `github.event.action != 'labeled'` should have matched `github.event.action == 'opened'`.

Ninfa-Jeon: i did work but i saw that the tests were being skipped when i change any other label after PR creation...trying to work that out

Ninfa-Jeon: looks like this is something that cannot be avoided

PR #1964, opened 2025-05-09
feat(e2e): Implement Secrets test suite

Ninfa-Jeon:
## Done

- Introduced Secrets test suite to verify that a user can create and use a secret

## Details

https://warthogs.atlassian.net/browse/WD-20948

### DIFF
--- a/e2e/base/secrets.spec.ts
+++ b/e2e/base/secrets.spec.ts
@@ -1,0 +1,99 @@
+import { expect } from "@playwright/test";
+
+import { test } from "../fixtures/setup";
+import { ActionStack } from "../helpers/action";
+import { AddModel, DeployApplication } from "../helpers/actions";
+import type { User } from "../helpers/auth";
+import type { Model, Application } from "../helpers/objects";
+import { generateRandomName } from "../utils";
+
+test.describe("Secrets", () => {
+  let actions: ActionStack;
+  let user: User;
+  let model: Model;
+  let application: Application;
+
+  test.beforeAll(async ({ jujuCLI, testOptions }) => {
+    test.setTimeout(300000);
+    actions = new ActionStack(jujuCLI);
+
+    await actions.prepare((add) => {
+      user = add(jujuCLI.createUser());
+      model = add(new AddModel(user));
+      application = add(new DeployApplication(model, testOptions.provider));
+    });
+  });
+
+  test.afterAll(async () => {
+    await actions.rollback();
+  });
+
+  test("Can create and use a secret", async ({ page }) => {
+    await user.dashboardLogin(page, "/models?enable-flag=rebac");
+
+    // Go to the secrets tab inside the model
+    await page.getByRole("link", { name: model.name }).click();
+    await page.getByRole("link", { name: "Secrets" }).click();
+
+    await page.getByRole("button", { name: "Add secret" }).click();
+    await expect(page.getByTestId("secret-form-panel")).toBeInViewport();
+
+    // Add a secret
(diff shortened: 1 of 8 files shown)


e2e/base/secrets.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      .getByRole("button", { name: "Add secret" })
+      .click();
+
+    // Verify the secret was added and copy the URI
+    await expect(page.getByTestId("secret-form-panel")).not.toBeInViewport();
+    await expect(page.locator("tr", { hasText: secret })).toBeInViewport();
+    const secretURI = await page
+      .locator("tr", { hasText: secret })
+      .locator("td.has-hover")
+      .innerText();
```
Ninfa-Jeon, line 57: The copy to clipboard and paste from clipboard action is not supported for HTTP in Playwright due to security reasons.

e2e/base/secrets.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+      .fill(secretURI);
+    await page.getByRole("button", { name: "Save and apply" }).click();
+
+    await expect(
+      page.getByText("Are you sure you wish to apply these changes?"),
+    ).toBeInViewport();
+    await page.getByRole("button", { name: "Yes, apply changes" }).click();
+
+    if (await page.getByText("Grant secrets?").isVisible())
+      await page.getByRole("button", { name: "Yes" }).click();
```
Ninfa-Jeon, line 80: this dialog does not always open

e2e/helpers/auth/Users.ts
```diff
@@ -59,7 +59,7 @@ export abstract class User {
    */
   abstract dashboardLogin(
     page: Page,
-    url: string,
+    url?: string,
```
Ninfa-Jeon, line 62: made this optional for cases where we reload without changing the location

e2e/base/secrets.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    await expect(page.getByTestId("config-panel")).not.toBeInViewport();
+
+    // Reload the page and verify the secret was saved
+    await page.reload();
+    if (await page.getByText("Log in to the dashboard").isVisible())
+      await user.dashboardLogin(page);
```
Ninfa-Jeon, line 87: we don't always see a login popup

e2e/base/secrets.spec.ts
huwshimi: Was that in CI or locally? It won't need to grant the secret if it has already been granted.

e2e/base/secrets.spec.ts
huwshimi: It won't need to log in again for Candid or OIDC, but it will for Local. I've added a method to abstract reloading the page and handle it correctly for each backend: https://github.com/canonical/juju-dashboard/pull/1966/files#diff-373044533a240d1100391eb22ae679fd4c40a75fe6058f6fa40b9fcd749eb136R54.

e2e/base/secrets.spec.ts
Ninfa-Jeon: okay so i think it may have been a local issue while I was still implementing it

huwshimi:
I've built upon this PR: https://github.com/canonical/juju-dashboard/pull/1971.

Mostly the changes are:
- Resolved conflicts with main.
- Used label and test id enums to make it easier to maintain.
- Used the `reloadDashboard` method I added in another PR for reloading the page.

PR #1965, opened 2025-05-12
chore(CI): Make e2e runs more reliable

huwshimi:
## Done

- Increase timeout when running actions on a unit.
- Wait for the vault cert file to be created.
- Retry `charmcraft pack` if it fails.
- Replace `|| true` with a retry so that it doesn't continue on real failures (if it times out the first time wait-for is called then that usually means the status updates stopped and the retry will run wait-for again and this time it will succeed as it will already be in the right state).
- Added logs to all workflows.

## Details

https://warthogs.atlassian.net/browse/WD-21941

### DIFF
--- a/.github/actions/failure-log/action.yml
+++ b/.github/actions/failure-log/action.yml
@@ -1,0 +1,56 @@
+name: Failure log
+description: Output logs if the workflow fails.
+inputs:
+  juju-env:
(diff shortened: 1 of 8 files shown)


Ninfa-Jeon (approved):
Great work!

PR #1966, opened 2025-05-13
chore(ci): add e2e test for configuring an application

huwshimi:
## Done

- Add support for reloading the dashboard and logging back in.
- Added configure application test.

Passing run:
https://github.com/canonical/juju-dashboard/actions/runs/14986276376?pr=1966

## Details

https://warthogs.atlassian.net/browse/WD-20945

### DIFF
--- a/e2e/helpers/objects/application.ts
+++ b/e2e/helpers/objects/application.ts
@@ -1,25 +1,35 @@
 /**
  * Charm to deploy based on provider.
  */
-export const CharmName = {
-  localhost: "anbox-cloud-dashboard",
-  microk8s: "hello-kubecon",
-};
+export enum CharmName {
+  localhost = "anbox-cloud-dashboard",
+  microk8s = "hello-kubecon",
+}
 
 /**
  * Action to run.
  */
-export const ActionName = {
-  "anbox-cloud-dashboard": "backup",
-  "hello-kubecon": "pull-site",
-};
+export enum ActionName {
+  "anbox-cloud-dashboard" = "backup",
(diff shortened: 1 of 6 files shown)


e2e/helpers/objects/application.ts
```diff
@@ -1,25 +1,35 @@
...
  * Charm to deploy based on provider.
  */
-export const CharmName = {
-  localhost: "anbox-cloud-dashboard",
-  microk8s: "hello-kubecon",
-};
+export enum CharmName {
+  localhost = "anbox-cloud-dashboard",
+  microk8s = "hello-kubecon",
+}
```
Ninfa-Jeon, line 7: thanks! I was unsure which of the two to use

e2e/helpers/objects/application.ts
```diff
@@ -1,25 +1,35 @@
...
+export enum ActionName {
+  "anbox-cloud-dashboard" = "backup",
+  "hello-kubecon" = "pull-site",
+}
+
+/**
+ * Config property to change.
+ */
+export enum ConfigName {
+  "anbox-cloud-dashboard" = "location",
```
Ninfa-Jeon, line 21: just adding a note that `ua_token` can also be used as it is essentially a deprecated configuration option

Ninfa-Jeon (approved):
LGTM

e2e/helpers/objects/application.ts
huwshimi: Ah yeah I decided not to use that one as I figured it might get removed.

PR #1968, opened 2025-05-13
chore(ci): use urls definition and automatically add feature flags

huwshimi:
## Done

- Use the URLs defined in the dashboard in the e2e tests to make them more easily to change in one place.
- Automatically add the feature flag.

### DIFF
--- a/e2e/helpers/auth/backends/Local.ts
+++ b/e2e/helpers/auth/backends/Local.ts
@@ -3,7 +3,7 @@ import fs from "node:fs/promises";
 import type { Page } from "@playwright/test";
 
 import type { User } from "..";
-import { juju } from "../../../utils";
+import { addFeatureFlags, juju } from "../../../utils";
 import { exec } from "../../../utils/exec";
 import type { Action } from "../../action";
 import type { JujuCLI } from "../../juju-cli";
@@ -53,7 +53,7 @@ export class LocalUser implements User {
   }
 
   async dashboardLogin(page: Page, url: string) {
-    await page.goto(url);
+    await page.goto(addFeatureFlags(url));
     await this.enterCredentials(page);
   }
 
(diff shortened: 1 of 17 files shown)


e2e/utils/addFeatureFlags.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Add required feature flags to a URL.
+ */
+export const addFeatureFlags = (url: string): string =>
+  [url, "enable-flag=rebac"].join(url.includes("?") ? "&" : "?");
```
andogq, line 5: It would be nice to see this done via `URL`, just to help with readability.

e2e/helpers/auth/backends/Local.ts
```diff
@@ -53,7 +53,7 @@ export class LocalUser implements User {
   }
 
   async dashboardLogin(page: Page, url: string) {
-    await page.goto(url);
+    await page.goto(addFeatureFlags(url));
```
andogq, line 56: Is ReBAC meant to be enabled in local?

andogq (approved):
Very minor nit, but this is a nice change. I forgot `urls` was a thing...

e2e/helpers/auth/backends/Local.ts
huwshimi: The feature flag is only used by JIMM at the moment, it's not tied to a specific authentication method so I figured we'd add the flags for all auth methods, and the feature flag just won't do anything unless the JIMM + OIDC workflow is run. This way if auth methods change, or if we add new feature flags we can do it in one spot.

Does that sound reasonable? I though about only appending the feature flag for the jimm env, but that meant passing around more data and the flag doesn't affect anything if the env isn't jimm...

e2e/utils/addFeatureFlags.ts
huwshimi: Unfortunately we only have the path and URL requires a hostname as well, so I've done a bit of manipulation with `URLSearchParams`. Do you think that's better?

andogq: Looks good to me!

andogq (approved):

PR #1969, opened 2025-05-13
chore(ci): use labels and test ids in e2e tests

huwshimi:
## Done

- Use the Label and TestId enums in the e2e tests to make it so the tests pass if the text/id is changed.

Passing run:
https://github.com/canonical/juju-dashboard/actions/runs/15008489507/job/42172661840?pr=1969

### DIFF
--- a/e2e/base/actions.spec.ts
+++ b/e2e/base/actions.spec.ts
@@ -6,6 +6,7 @@ import {
 } from "pages/EntityDetails/App/types";
 import { Label as ConfirmationDialogLabel } from "panels/ActionsPanel/ConfirmationDialog/types";
 import { TestId as ActionsPanelTestId } from "panels/ActionsPanel/types";
+import { ModelTab } from "urls";
 
 import { test } from "../fixtures/setup";
 import { ActionStack } from "../helpers/action";
@@ -74,7 +75,7 @@ test.describe("Actions", () => {
 
     // Go to the action logs
     if (process.env.AUTH_MODE === "oidc") {
-      await page.getByRole("link", { name: "Logs" }).click();
+      await page.getByRole("link", { name: ModelTab.LOGS }).click();
     } else {
       await page.getByTestId(AppTestId.SHOW_LOGS).click();
     }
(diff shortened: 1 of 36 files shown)


e2e/base/actions.spec.ts
```diff
@@ -1,5 +1,12 @@
 import { expect } from "@playwright/test";
 
+import {
+  Label as AppLabel,
+  TestId as AppTestId,
+} from "pages/EntityDetails/App/types";
+import { Label as ConfirmationDialogLabel } from "panels/ActionsPanel/ConfirmationDialog/types";
+import { TestId as ActionsPanelTestId } from "panels/ActionsPanel/types";
```
huwshimi, line 8: We have to import the enums from the types files instead of the index, otherwise the index will try to load the component as well, which won't be built.

e2e/jimm/rebac.spec.ts
```diff
@@ -29,7 +31,9 @@ test.describe("ReBAC Admin", () => {
       return user;
     });
     await user.dashboardLogin(page, "/permissions/users?enable-flag=rebac");
-    await expect(page.getByRole("link", { name: "Permissions" })).toBeVisible();
+    await expect(
+      page.getByRole("link", { name: PrimaryNavLabel.PERMISSIONS }),
+    ).toBeVisible();
     await expect(page.getByRole("heading", { name: "Users" })).toBeVisible();
```
andogq, line 37: I'm assuming we're not aiming for every string in the UI, but rather only those that have an enum already?

andogq (approved):
Nice fix
:shipit:

e2e/jimm/rebac.spec.ts
huwshimi: That was my approach, but now that I'm not in a rush I've done this properly and added the missing labels and test ids. I've added those changes in a second commit, if you wouldn't mind taking another look.

andogq (approved):
Thanks for adding all those labels!

PR #1974, opened 2025-05-21
fix: make the commands websocket authentication work with candid

huwshimi:
## Done

- Fix the origin that the macaroons are stored under when authenticating with the web CLI using Candid.

## QA

- Connect to a controller that uses Candid.
- Go to a model details page.
- Enter a command into the CLI e.g. `status`.
- A result should be returned from Juju.

## Details

https://warthogs.atlassian.net/browse/WD-22214

### DIFF
--- a/src/components/WebCLI/WebCLI.tsx
+++ b/src/components/WebCLI/WebCLI.tsx
@@ -173,7 +173,7 @@ const WebCLI = ({
       const origin = connection.current?.address
         ? new URL(connection.current?.address)?.origin
         : null;
-      const macaroons = origin ? bakery.storage.get(origin) : null;
+      const macaroons = origin ? bakery.storage.get(`${origin}/api`) : null;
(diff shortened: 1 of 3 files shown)


src/components/WebCLI/WebCLI.tsx
```diff
@@ -173,7 +173,7 @@ const WebCLI = ({
       const origin = connection.current?.address
         ? new URL(connection.current?.address)?.origin
         : null;
-      const macaroons = origin ? bakery.storage.get(origin) : null;
+      const macaroons = origin ? bakery.storage.get(`${origin}/api`) : null;
```
huwshimi, line 176: This needed to match the format as is provided in config.js which ends with `/api`

Ninfa-Jeon (approved):
Wonderful! Thanks for fixing this

PR #1976, opened 2025-05-26
fix: handle commands that return a single line

huwshimi:
## Done

- Fix an issue where a command that returned a single line of output would display the help instead of the result.
- Fix an issue with the help being triggered by all keys.

## QA

- Go to a model and type a command that returns a single line e.g. `run` or `config`.
- The error message should be displayed.
- Tab to the help icon and press enter or space and the help should get toggled. Pressing other keys should not toggle the help.

## Details

https://warthogs.atlassian.net/browse/WD-22300

### DIFF
--- a/src/components/WebCLI/WebCLI.tsx
+++ b/src/components/WebCLI/WebCLI.tsx
@@ -42,7 +42,8 @@ const WebCLI = ({
   protocol = "wss",
 }: Props) => {
   const connection = useRef<Connection | null>(null);
-  const [shouldShowHelp, setShouldShowHelp] = useState(false);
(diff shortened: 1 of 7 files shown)


src/components/WebCLI/WebCLI.tsx
```diff
@@ -222,7 +224,11 @@ const WebCLI = ({
     }
   };
 
-  const showHelp = () => {
+  const showHelp = (event: React.MouseEvent | React.KeyboardEvent) => {
+    // Only trigger the help from the space or enter keys (or mouse clicks).
```
huwshimi, line 228: I initially thought this was related, so I fixed it and left the change in as we need it anyway.

PR #1978, opened 2025-05-29
refactor: move application layout to the top route

huwshimi:
## Done

- Move application layout to the top route.
- Add handlers for displaying the CLI in the status area.

## QA

- Load the app and check that the application layout is displayed as before.
- Navigate to a model page and check that the CLI is displayed.

## Details

https://warthogs.atlassian.net/browse/WD-22421


src/components/CaptureRoutes/CaptureRoutes.tsx
```diff
@@ -17,7 +17,7 @@ const CaptureRoutes = (): JSX.Element => {
     });
   }, [location, sendAnalytics]);
 
-  return <Outlet />;
+  return null;
```
huwshimi, line 20: This component no longer wraps all the other components so doesn't need an outlet.

src/components/LogIn/LogIn.test.tsx
```diff
@@ -33,7 +33,6 @@ describe("LogIn", () => {
     });
     renderComponent(<LogIn />, {
       state,
-      path: "*",
```
huwshimi, line 36: Because of the wrapping component in the renderComponent util some of these paths had to change.

src/layout/BaseLayout/BaseLayout.test.tsx
huwshimi, line 1: The tests in this file were moved to MainContent.

src/layout/BaseLayout/BaseLayout.tsx
```diff
@@ -65,37 +55,9 @@ const BaseLayout = ({
           />
         }
         sideNavigation={<PrimaryNav />}
-        status={status}
+        status={<Status status={status} />}
       >
-        <div
```
huwshimi, line 70: This section became MainContent

src/layout/MainContent/MainContent.test.tsx
huwshimi, line 1: Moved from BaseLayout

src/layout/MainContent/MainContent.tsx
huwshimi, line 1: Moved from BaseLayout

src/components/CaptureRoutes/CaptureRoutes.tsx
andogq, line 1: If this isn't being used to actually render anything, would it be more sensible to just have it as a hook?

src/components/CaptureRoutes/CaptureRoutes.tsx
```diff
@@ -1,11 +1,11 @@
 import type { JSX } from "react";
 import { useEffect } from "react";
-import { Outlet, useLocation } from "react-router";
+import { useLocation } from "react-router";
 
 import useAnalytics from "hooks/useAnalytics";
 import useFeatureFlags from "hooks/useFeatureFlags";
 
-const CaptureRoutes = (): JSX.Element => {
+const CaptureRoutes = (): JSX.Element | null => {
```
andogq, line 8:
```suggestion
const CaptureRoutes = (): null => {
```

Should we have the signature reflect that it doesn't render anything? Or are we leaving the door open for it to render stuff in the future?

src/layout/Status/Status.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import type { Props } from "./types";
+import { StatusView } from "./types";
+
+const Status = ({ status }: Props) => {
+  switch (status) {
+    case StatusView.CLI:
+      return <JujuCLI />;
+    default:
+      return null;
```
andogq, line 11: Is it worth emitting some kind of 'not found', so we don't rely on junk states hiding the panel? I know the enum will help us get over the line here, but was just thought!

andogq (approved):
Huge change, but it makes sense to have this inverted. I've got a few minor questions, but I think they're more-so problems with me rather than the code.

src/components/CaptureRoutes/CaptureRoutes.tsx
huwshimi: Thanks, I've moved these calls into the application layout (I think they were there a long time ago but got moved to a component at some point when the layout got moved around).

src/layout/Status/Status.tsx
huwshimi: Thanks, done.

PR #1982, opened 2025-06-04
refactor: move dashboard-specific code out of WebCLI

Ninfa-Jeon:
## Done

- Moved dashboard-specific code like the `useAnalytics` and the `useStore` hooks out of the WebCLI
- Corrected a typo in `HACKING.md`

## Details

We are moving code strictly tied to the dashboard out of the WebCLI component to reduce tight coupling of the two. This is to aid in making the WebCLI component a standalone project someday.

## QA

- Login to the dashboard
- Go to Model details page and expand the WebCLI
- Try using it with a couple of commands, clicking help, etc. Everything should work fine.

## Details
https://warthogs.atlassian.net/browse/WD-22424

## Screenshots
Screenshot 2025-06-05 at 10 27 00 AM

### DIFF
--- a/src/components/JujuCLI/JujuCLI.tsx
+++ b/src/components/JujuCLI/JujuCLI.tsx
@@ -4,7 +4,12 @@ import { useParams } from "react-router";
 
 import type { EntityDetailsRoute } from "components/Routes";
 import WebCLI from "components/WebCLI";
-import { getUserPass, getIsJuju } from "store/general/selectors";
+import useAnalytics from "hooks/useAnalytics";
+import {
+  getUserPass,
+  getIsJuju,
+  getActiveUserTag,
+} from "store/general/selectors";
 import {
   getControllerDataByUUID,
   getModelInfo,
@@ -78,6 +83,7 @@ const processHelp = (messages: string[]) => {
 
 const JujuCLI = () => {
   const routeParams = useParams<EntityDetailsRoute>();
+  const sendAnalytics = useAnalytics();
   const { userName, modelName } = routeParams;
   const modelUUID = useAppSelector((state) =>
     getModelUUIDFromList(state, modelName, userName),
@@ -106,7 +112,18 @@ const JujuCLI = () => {
       .replace("ws://", "")
       .replace("wss://", "")
       .replace("/api", "") || null;
-  const wsProtocol = primaryControllerData?.[0].split("://")[0];
+  const wsProtocol = primaryControllerData?.[0].split("://")[0] ?? "wss";
+
+  const activeUser = useAppSelector((state) =>
+    getActiveUserTag(state, primaryControllerData?.[0]),
+  );
+
+  function onCommandSent(_command?: string) {
+    sendAnalytics({
+      category: "User",
+      action: "WebCLI command sent",
+    });
+  }
 
   useEffect(() => {
     if (isJuju && getMajorMinorVersion(modelInfo?.version) >= 2.9) {
@@ -126,6 +143,8 @@ const JujuCLI = () => {
       controllerWSHost={controllerWSHost}
       credentials={credentials}
       modelUUID={modelUUID}
+      onCommandSent={onCommandSent}
+      activeUser={activeUser}
       processOutput={{
         help: {
           // This should match "help" but not "help bootstrap" etc.
@@ -133,7 +152,7 @@ const JujuCLI = () => {
           process: processHelp,
         },
       }}
-      protocol={wsProtocol ?? "wss"}
+      protocol={wsProtocol}
     />
   );
 };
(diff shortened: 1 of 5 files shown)


src/components/JujuCLI/JujuCLI.tsx
```diff
@@ -14,6 +20,8 @@ import { getMajorMinorVersion } from "utils";
 
 const JujuCLI = () => {
   const routeParams = useParams<EntityDetailsRoute>();
+  const sendAnalytics = useAnalytics();
+  const storeState = useStore().getState();
```
huwshimi, line 24: While we're moving this code around we might as well fix this to call `getActiveUserTag` in a `useSelector) rather than getting the entire state here. I think it would just require a small change to `getActiveUserTag` to handle receiving undefined or null if `originalWSOrigin` isn't defined.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -212,10 +195,7 @@ const WebCLI = ({
           commands: [command],
         }),
       );
-      sendAnalytics({
-        category: "User",
-        action: "WebCLI command sent",
-      });
+      onCommandSent();
```
huwshimi, line 198: I think it would be nice to pass the command to this method, not that we use it now, but seems like a sensible thing to expose if this library was to be used externally from the dashboard.

Ninfa-Jeon: Done.

src/components/JujuCLI/JujuCLI.tsx
huwshimi: Right now TypeScript won't complain as we're passing `${originalWSOrigin}/api` which will always be a truthy string, but `originalWSOrigin` can be `null` which would be passed as `${null}/api`, so we need to check that `originalWSOrigin` has been set and if it hasn't then pass `null`.

Ninfa-Jeon: how about this?

src/components/JujuCLI/JujuCLI.tsx
```diff
@@ -106,7 +112,28 @@ const JujuCLI = () => {
...
+
+  const wsAddress = useMemo(() => {
+    if (!controllerWSHost || !modelUUID) {
+      return null;
+    }
+    return `${wsProtocol}://${controllerWSHost}/model/${modelUUID}/commands`;
+  }, [controllerWSHost, modelUUID, wsProtocol]);
+
+  const originalWSOrigin = wsAddress
+    ? `${new URL(wsAddress).origin}/api`
```
Ninfa-Jeon, line 125: instead of repeating the same check, i just moved `/api` up

src/components/JujuCLI/JujuCLI.tsx
```diff
@@ -106,7 +112,28 @@ const JujuCLI = () => {
...
+    : null;
+  const activeUser = useAppSelector((state) =>
+    getActiveUserTag(state, originalWSOrigin),
```
huwshimi, line 128: Looking at the logic that was originally implemented, it seems that it does a lot of logic to transform the websocket address a number of times, and the end result is exactly what we had to begin with, which is `primaryControllerData?.[0]` so I think this can just be `getActiveUserTag(state, primaryControllerData?.[0])` and then we don't need the `wsAddress` logic in this component.

huwshimi (commented):
I've made a couple of suggestions here that should simplify this.

src/components/JujuCLI/JujuCLI.tsx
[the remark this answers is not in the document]
Ninfa-Jeon: I've added one test each.

src/components/WebCLI/WebCLI.test.tsx
```diff
@@ -256,6 +258,7 @@ describe("WebCLI", () => {
         commands: ["status --color"],
       }),
     );
+    expect(props.onCommandSent).toBeCalledWith("status --color");
```
huwshimi, line 261: This particular test isn't really about sending commands, it just happens to send one so it'd probably be better to add a new one for testing this case.

src/components/WebCLI/WebCLI.tsx
```diff
@@ -27,6 +24,8 @@ type Props = {
   controllerWSHost: string;
   credentials?: Credential | null;
   modelUUID: string;
+  onCommandSent: (command?: string) => void;
+  activeUser: string | undefined;
```
huwshimi, line 28:
```suggestion
  activeUser?: string;
```

huwshimi (approved):
Just a couple of small things and then this is good to go!

src/components/WebCLI/WebCLI.tsx
Ninfa-Jeon: I can't believe the number of times I do this

PR #1987, opened 2025-06-13
fix(ci): send failure notification by default

andogq:
## Done

- send failure notification by default
- print action configuration to assist with debugging


Ninfa-Jeon (approved):
LGTM

PR #1988, opened 2025-06-13
fix: manually extract path parameters in panels

andogq:
## Done

- Fix broken actions panel

## QA

- Navigate through all the different panels, and ensure they run.

## Details

Due to #1978, `<Panels />` (rendered by `<BaseLayout />`) is no longer rendered within the sub-routers for the model detail pages. This means that the panels no longer inherit the route, and so `useParams` doesn't return the correct parameters.

This fix manually uses `useMatch` provided by react-router to re-parse the location with a manually-specified path. An alternate fix could involve re-structuring the render tree, to ensure that `<Panels />` is again rendered as a child of the router.


Ninfa-Jeon (approved):
LGTM!

Ninfa-Jeon:
Configure application panel:
Screenshot 2025-06-16 at 9 11 30 AM

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -4,12 +4,11 @@ import classnames from "classnames";
 import cloneDeep from "clone-deep";
 import type { JSX, MouseEvent } from "react";
 import { useCallback, useEffect, useRef, useState } from "react";
-import { useParams } from "react-router";
 
 import FadeIn from "animations/FadeIn";
 import CharmIcon from "components/CharmIcon";
 import Panel from "components/Panel";
-import type { EntityDetailsRoute } from "components/Routes";
+import { useModelAppParams, useModelIndexParams } from "components/hooks";
```
Ninfa-Jeon:
```suggestion
import { useModelAppParams } from "components/hooks";
```

Ninfa-Jeon (approved):
Thanks, it works now.

PR #1990, opened 2025-06-19
feat: get model list permissions when using JIMM

huwshimi:
## Done

- Fetch model list permissions and hide/show the access button when using JIMM/ReBAC.

## QA

- Set up a local JIMM
- Using the new instructions in this PR in HACKING.md add a new user e.g. username: `user2`, email: `user2@example.com`
- In your JIMM multipass, using the jimm-test user (the one that gets created when setting up JIMM) run the following commands:
```shell
juju switch jimm-dev # you might already be on this controller
juju add-model read-only-model
juju grant-cloud user2@example.com add-model localhost
juju grant user2@example.com read read-only-model
```
- Now you can log in as the new user and add a new model. When prompted to log in then open the link in a private browser and log in with user2 and the password you set.
```
juju logout
juju login
juju update-credential localhost localhost -c jimm-dev
juju add-model user2-model
```
Now visit the dashboard in your private browser session, with ReBAC enabled: http://jimm.localhost:8036/models?enable-flag=rebac.
- Log in as user2.
- In the model list you should see read-only-model and user2-model.
- Hover user2-model and you should see the Access button.
- Hover read-only-model and you should NOT see the Access button.

## Details

https://warthogs.atlassian.net/browse/WD-22731


eslint.config.d.mts
andogq, line 1: This one should probably be deleted too

andogq (approved):
Aside from deleting that eslint.config.d.mts file, looks all good to me!

eslint.config.d.mts
huwshimi: Oops thanks, that's weird, I even have `rm eslint.config.d.mts` in my scrollback, I wonder why it didn't get removed..

PR #1996, opened 2025-06-25
feat: persist CLI history

huwshimi:
## Done

- Support storing the history for each model.
- Store the history in Redux.
- Recreate the history when revisiting a model.

## QA

- Go to a model details page.
- Enter a command in the CLI.
- Navigate away from the model details and then back again and open the CLI and it should display the previous output.

## Details

https://warthogs.atlassian.net/browse/WD-22422


andogq (approved):

andogq:
Works well! A tiny thought I had: if the CLI was opened before navigating away from the model, it would be nice if it were restored to the same opened state upon return to that model. Doesn't have to block this PR though, can always be something we add later.

PR #1997, opened 2025-06-25
refactor: move output processing to own component

huwshimi:
## Done

Wrapped each command and its output in a memoised component so that they don't need to re-process their contents when new items are added.

## QA

- Go to a model details page.
- Submit a lot of commands in CLI and it shouldn't get slower.

## Details

https://warthogs.atlassian.net/browse/WD-23134

### DIFF
--- a/src/components/WebCLI/Output/OutputCommand/OutputCommand.test.tsx
+++ b/src/components/WebCLI/Output/OutputCommand/OutputCommand.test.tsx
@@ -1,4 +1,4 @@
-import { screen } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 
 import { renderComponent } from "testing/utils";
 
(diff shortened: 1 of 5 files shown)


src/components/WebCLI/Output/Output.test.tsx
```diff
@@ -124,174 +124,6 @@ describe("Output", () => {
     );
   });
 
-  it("should display the content with correct formatting", () => {
```
huwshimi, line 127: These tests have been moved to the new component tests.

src/components/WebCLI/Output/Output.tsx
```diff
@@ -166,31 +159,14 @@ const WebCLIOutput = ({
     }
   }, [content, contentChanged, height]);
 
-  const output = content.map(({ command, messages }, i) => {
-    let response: ReactNode = null;
```
huwshimi, line 170: This has been moved to the new component.

src/components/WebCLI/Output/OutputCommand/OutputCommand.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const arePropsEqual = (oldProps: Props, newProps: Props) => {
+  const { messages: oldMessages, ...oldRest } = oldProps;
+  const { messages: newMessages, ...newRest } = newProps;
+  for (const key in Object.keys(newRest)) {
+    if (newRest[key as keyof RestProps] !== oldRest[key as keyof RestProps]) {
+      return false;
+    }
+  }
```
andogq, line 55: Maybe too not-picky, but can we filter out the 'messages' key within the `for` loop? This would just save the double allocation when spreading into `oldRest`/`newRest` each time this check is performed.

src/components/WebCLI/Output/OutputCommand/OutputCommand.test.tsx
```diff
@@ -0,0 +1,127 @@
+import { screen } from "@testing-library/react";
+
+import { renderComponent } from "testing/utils";
+
+import OutputCommand from "./OutputCommand";
+
+describe("OutputCommand", () => {
```
andogq, line 7: Is it a little redundant to have a test ensuring that the process function isn't re-run if the content remains the same?

andogq (approved):
Nice change!

src/components/WebCLI/Output/OutputCommand/OutputCommand.test.tsx
huwshimi: I think it's a good idea as we're overriding the default memoise equality check. I've pushed some new tests but in the process of adding them I realised that a couple of the props are objects not functions so I've had to memoise them and changed the comparison check to always do a deep equal.

src/components/WebCLI/Output/OutputCommand/OutputCommand.tsx
huwshimi: This ended up always needing to do a deep equal as part of my refactor.

PR #2010, opened 2025-07-18
feat: dev toolbar

andogq:
## Done

- Create a dev toolbar
- TODO: Verify with JAAS

## QA

- `NODE_ENV=development yarn start`
- Click on `$_` button in bottom left corner
- Experiment with options
- `NODE_ENV=production yarn start`
- See that `$_` button is no longer present

## Details

Contains a handful of utilities to help with day-to-day development.

### Controller Debugging

image

Shows available controller information from the connection. Includes a `http` link to the controller, to easily accept HTTPS certificates. Gives a hint if controller can't be connected due to HTTP certificate problems.

image

### User Management

image

Store user credentials in local storage, and automatically login when app loads.

https://github.com/user-attachments/assets/e2bc0300-2efc-4549-8c0b-996614a23e11

### Feature flags

List available flags found in local storage, and perform CRUD operations on them. Doesn't deeply integrate with the application (eg doesn't list available feature flags, it's free form).

image

### Maximised

image

### Minimised

image


src/components/DevBar/DevBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+import classNames from "classnames";
+
+import useLocalStorage from "hooks/useLocalStorage";
+
+import { useWidgets } from "./widgets";
+
+const MINIMISED_KEY = "DEV_minimised";
+
+export default function DevBar() {
+  const [minimised, setMinimised] = useLocalStorage(MINIMISED_KEY, false);
```
Ninfa-Jeon, line 11: can you explain why we are using a `localStorage` item just to check if the component is minimised? Couldn't we simply use the `useState`?

src/components/DevBar/widgets/feature-flags.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export default {
+  Title: () => {
+    const [enabledFlags] = useLocalStorage<string[]>(ENABLED_FLAGS, []);
+
+    return (
+      <StatusTitle
+        title="Feature flags"
+        label={
+          // TODO: This won't get updated without a refresh, since it's just pulling from local storage.
```
Ninfa-Jeon, line 19: Will you be working on this `TODO` in this PR?

src/components/DevBar/DevBar.tsx
andogq: We certainly could use `useState`, however that would mean the minimised state wouldn't persist across page refreshes. Since this piece of UI is always present, I figured it might get annoying if we had to manually minimise it on each page load.

src/components/DevBar/widgets/feature-flags.tsx
andogq: I wasn't planning on it, since it would require touching the other feature flag code. It might be something if we ever expand the feature flag functionality, but I feel the existing system works for now. Open to feedback on that though!

Ninfa-Jeon: You're right about this. For a dev-setup this is already a massive upgrade. We can think about this in the future. Thanks for the work

huwshimi:
A couple of QA things (though I might be doing something wrong).

After I saved a user and set it to auto login, the dashboard shows the login screen with an error:

Screenshot 2025-07-23 at 4 05 05 pm

Trying to add a feature flag reloads the page:

Screen Recording 2025-07-23 at 4 05 31 pm

src/components/DevBar/widgets/juju-user.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+    const {
+      result,
+      result: { container },
+    } = renderComponent(<Title />, { state });
+
+    expect(container).toHaveTextContent("Juju User");
+
+    const spans = result.baseElement.getElementsByTagName("span");
+    expect(spans).toHaveLength(1);
+    expect(spans[0]).toHaveClass("show-status");
```
huwshimi, line 63: Same thing here about getting the element via the text so the order doesn't matter.

src/components/DevBar/DevBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const sections = useWidgets();
+
+  if (!import.meta.env.DEV) {
+    console.error("`<DevBar />` rendered in non-dev mode");
+    return null;
+  }
+
+  return (
+    <div className="dev-bar">
+      {minimised && (
```
huwshimi, line 21: I prefer to use a ternary for conditionally displaying jsx, otherwise falsy values sometimes get display on the page (e.g. "0"). Likely the local storage won't ever be set to something like `0` but I find it good practice to always use a ternary and then you don't have to consider what the values might be.

src/components/DevBar/DevBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+        <Button hasIcon onClick={() => setMinimised(false)} appearance="">
```
huwshimi, line 22: Not sure if this needs `appearance=""` does it?

src/components/LogIn/UserPassForm/login.ts
huwshimi, line 1: Now that this is in its own file it should probably have some tests.

src/scss/index.scss
```diff
@@ -63,6 +63,7 @@ $color-navigation-background: $color-brand;
 @import "../components/Banner/banner";
 @import "../components/ChipGroup/chip-group";
 @import "../components/ContentReveal/content-reveal";
+@import "../components/DevBar/dev-bar";
```
huwshimi, line 66: We'll be shipping this and the components in the public bundle, but I don't know if there's an easy way around that.

huwshimi (approved):
I've made a few code quality suggestions here. Even though this is just for dev, I think some of these are worth doing.

src/components/DevBar/widgets/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+export function useWidgets() {
+  const sections = WIDGETS.map(({ Title, Widget }) => {
+    const title = <Title />;
+
+    if (title === null) {
+      return null;
+    }
+
+    return { title, content: <Widget /> };
+  }).filter((widget) => widget !== null);
```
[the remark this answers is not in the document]
andogq: The line 19 `widget !== null` is required to allow TypeScript to narrow the type of the array items. Without it, it thinks the type is `({ title: Element; content: Element } | null)[]`.

src/components/DevBar/DevBar.tsx
andogq: Ah I pulled this from the storybook examples, but doesn't look like it makes a difference.

huwshimi (approved):

PR #2017, opened 2025-07-25
chore(CI): Docker JIMM workflow

huwshimi:
## Done

Add a workflow to test the dashboard with Docker JIMM.

## Details

https://warthogs.atlassian.net/browse/WD-23525
https://warthogs.atlassian.net/browse/WD-23524


.github/actions/run-playwright/action.yml
```diff
@@ -33,13 +33,18 @@ runs:
     - name: Get dashboard address
       if: ${{ inputs.run == 'true' }}
       id: dashboard
-      run: echo "address=http://$(address=$(juju show-unit dashboard/0 | yq '.dashboard/0.public-address'); if [ "$address" != "null" ]; then echo $address; else juju show-unit dashboard/0 | yq '.dashboard/0.address'; fi):8080" >> $GITHUB_OUTPUT
+      run: |
+        if [[ -z "${DASHBOARD_ADDRESS}" ]]; then
+          echo "address=http://$(address=$(juju show-unit dashboard/0 | yq '.dashboard/0.public-address'); if [ "$address" != "null" ]; then echo $address; else juju show-unit dashboard/0 | yq '.dashboard/0.address'; fi):8080" >> $GITHUB_OUTPUT
+        else
+          echo "address=http://$DASHBOARD_ADDRESS" >> $GITHUB_OUTPUT
```
huwshimi, line 40: This allows it to use the provided dashboard address as there is no juju unit to get an address from.

.github/actions/setup-docker-dashboard/action.yml
```diff
@@ -44,6 +39,11 @@ runs:
         };
         EOF
       shell: bash
+    - name: Build Juju Dashboard image
```
huwshimi, line 42: I realised it needed to build the image _after_ modifying the config, so the config is included in the image.

e2e/base/models.spec.ts
```diff
@@ -20,6 +20,8 @@ test.describe("Models", () => {
   let user2Model: Model;
 
   test.beforeAll(async ({ jujuCLI }) => {
+    // Give the beforeAll enough time to create the models:
+    test.setTimeout(300000);
```
huwshimi, line 24: It turns out the reason some of these were failing is because the `beforeAll` would time out before the setup was complete.

nginx.conf
```diff
@@ -20,7 +20,7 @@ server {
         try_files /config.js =404;
     }
 
-    location /static {
+    location /assets {
```
huwshimi, line 23: This path was incorrect, but we weren't using this nginx config (the charms have their own nginx config) so didn't realise until now.

Ninfa-Jeon (approved):
LGTM!

.github/actions/setup-docker-dashboard/action.yml
huwshimi: @Ninfa-Jeon do you mean the new `build.sh` scripts? If so, those are for building the charms, whereas this step builds a docker image that serves the dashboard (instead of being served by the charm). This image will be used by the JIMM team to run the dashboard within their docker compose env. Or am I missing something?

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -0,0 +1,84 @@
...
+          ghcr-pat: ${{ secrets.GITHUB_TOKEN }}
+          dump-logs: true
+      - name: Set the CLI admin's password
+        run: |
+          juju switch ${{ env.CONTROLLER_NAME }}
+          echo '${{ inputs.admin-password }}' | juju change-user-password --no-prompt
+      - name: Restart JIMM
+        run: docker restart jimm
+      - name: Install jimmctl
+        run: sudo snap install --channel=3 jimmctl
```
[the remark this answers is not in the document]
huwshimi: Oh yeah, thanks, I need to rebase this on main now that it has the changes from my other PR to update the e2e tests to use the jaas plugin. I'll update this step to install the jaas plugin instead.

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -0,0 +1,84 @@
...
+          is-juju: false
+          controller-ws-address: ws://jimm.localhost:17070/api
+      - name: Checkout JIMM repo
+        uses: actions/checkout@v4
+        with:
+          repository: canonical/jimm
+          path: jimm
+      - name: Update JIMM with the dashboard address
+        run: |
+          sed -i -e 's/JIMM_DASHBOARD_FINAL_REDIRECT_URL: "https:\/\/jaas.ai"/CORS_ALLOWED_ORIGINS: "http:\/\/${{ steps.docker-dashboard.outputs.address }}"\n      JIMM_DASHBOARD_LOCATION: "http:\/\/${{ steps.docker-dashboard.outputs.address }}"\n      JIMM_DASHBOARD_FINAL_REDIRECT_URL: "http:\/\/${{ steps.docker-dashboard.outputs.address }}"/g' jimm/docker-compose.common.yaml
```
andogq, line 47: We could configure this using `yq` which might be slightly simpler to reason about

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -0,0 +1,84 @@
...
+          jimm-version: dev
+          juju-channel: "3.6/edge"
+          ghcr-pat: ${{ secrets.GITHUB_TOKEN }}
+          dump-logs: true
+      - name: Set the CLI admin's password
+        run: |
+          juju switch ${{ env.CONTROLLER_NAME }}
+          echo '${{ inputs.admin-password }}' | juju change-user-password --no-prompt
+      - name: Restart JIMM
+        run: docker restart jimm
```
andogq, line 66: Should we have a step to wait until the healthcheck passes?

e2e/base/models.spec.ts
andogq: Would it be worth moving this timeout into the playwright config? And maybe only raise it for JIMM runs?

nginx.conf
andogq: Could this config be replaced with the one in the k8s-charm `src` dir?

andogq (approved):
A few minor comments, but looks good!

e2e/base/models.spec.ts
huwshimi: I'm a little cautious about setting the default this high as it will wait for any failure to resolve until it times out, so if it fails on something that has broken it will sit there waiting until the timeout even though it will never succeed. Does that sound sensible?

nginx.conf
huwshimi: Done.

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: "jimm/go.mod"
+      - name: Go vendor to speed up docker build
+        run: cd jimm && go mod vendor
+      - name: Start JIMM (pull request)
+        uses: ./jimm/.github/actions/test-server
+        with:
+          jimm-version: dev
+          juju-channel: "3.6/edge"
```
andogq, line 61: Could we pass this in as a parameter with a default parameter? Unsure if JIMM team will want to have control over this.

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -0,0 +1,90 @@
...
+          dump-logs: true
+      - name: Set the CLI admin's password
+        run: |
+          juju switch ${{ env.CONTROLLER_NAME }}
+          echo '${{ inputs.admin-password }}' | juju change-user-password --no-prompt
+      - name: Restart JIMM
+        run: docker restart jimm # Fixes a context cancelled issue when fetching models from the dashboard.
+      - name: Wait for JIMM container to be ready
+        run: while [ "$(docker inspect -f {{.State.Restarting}} jimm)" == "true" ]; do sleep 1; done
+        shell: bash
```
andogq, line 72: Do we need to worry about a timeout on this at all?

andogq (approved):
Two very minor thoughts, but looks good to merge!

.github/workflows/e2e-jimm-docker-keycloak.yml
huwshimi: Done.

PR #2035, opened 2025-08-12
feat: Implement client for model migration

Ninfa-Jeon:
## Done

- Added client for `migrateModel` facade method
- Added client for `listMigrationTargets` facade method
- Cleaned up `src/juju/jimm/api.ts` to make all functions consistent

## Details

https://warthogs.atlassian.net/browse/WD-23988

### DIFF
--- a/src/juju/jimm/JIMMV4.ts
+++ b/src/juju/jimm/JIMMV4.ts
@@ -57,6 +57,13 @@ export type CheckRelationsResponse = {
 };
 
 // As typed in JIMM:
+// https://github.com/canonical/jimm/blob/c82e0cbba03387031a46ec3a48494803dce16995/pkg/api/params/params.go#L482
+export type MigrateModelInfo = {
+  "model-tag": string;
+  "target-controller": string;
+};
+
+// As typed in JIMM:
 // https://github.com/canonical/jimm/blob/c82e0cbba03387031a46ec3a48494803dce16995/pkg/api/params/params.go#L217
 export type ListControllersResponse = {
   controllers: ControllerInfo[];
@@ -115,20 +122,13 @@ class JIMMV4 extends JIMMV3 {
     });
   }
 
-  migrateModel(
-    modelUUID: string,
-    targetController: string,
-  ): Promise<InitiateMigrationResults> {
+  migrateModel(specs: MigrateModelInfo[]): Promise<InitiateMigrationResults> {
     return new Promise((resolve, reject) => {
       const req = {
         type: "JIMM",
         request: "MigrateModel",
         version: 4,
-        params: {
-          specs: [
-            { "model-tag": modelUUID, "target-controller": targetController },
-          ],
-        },
+        params: { specs },
       };
       this._transport.write(req, resolve, reject);
     });
--- a/src/juju/jimm/api.ts
+++ b/src/juju/jimm/api.ts
@@ -1,7 +1,7 @@
-import type { ConnectionWithFacades } from "juju/types";
+import type { ConnectionWithFacades, Facades } from "juju/types";
 
 import type { FindAuditEventsRequest } from "./JIMMV3";
-import type { RelationshipTuple } from "./JIMMV4";
+import type { MigrateModelInfo, RelationshipTuple } from "./JIMMV4";
 
 export enum Label {
   NO_JIMM = "Not connected to JIMM.",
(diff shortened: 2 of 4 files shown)


src/juju/jimm/JIMMV4.ts
```diff
@@ -107,6 +117,41 @@ class JIMMV4 extends JIMMV3 {
...
+          ],
+        },
+      };
+      this._transport.write(req, resolve, reject);
+    });
+  }
+
+  listMigrationTargets(
+    modelTag: string,
+  ): Promise<{ controllers: ControllerInfo[] | null }> {
```
huwshimi, line 142: We should make a separate return type to match the return type in JIMM. It looks like in JIMM it is called `ListControllersResponse` https://github.com/canonical/jimm/blob/c82e0cbba03387031a46ec3a48494803dce16995/pkg/api/params/params.go#L217 (normally we just match the names to what they are in JIMM/Juju).

It'd also be helpful to leave a link to the type in JIMM so we know where to look if it changes.

This type can then also be reused as the return type for `listMigrationTargets`.

src/juju/jimm/JIMMV4.ts
```diff
@@ -107,6 +117,41 @@ class JIMMV4 extends JIMMV3 {
...
+  ): Promise<InitiateMigrationResults> {
+    return new Promise((resolve, reject) => {
+      const req = {
+        type: "JIMM",
+        request: "MigrateModel",
+        version: 4,
+        params: {
+          specs: [
+            { "model-tag": modelUUID, "target-controller": targetController },
+          ],
```
huwshimi, line 133: Because we're hard coding this to only supply one model in the array it would probably be nice to also also extract the single result from the array (if there's not an error) so that we don't have to deal with an array of results where we call this.

huwshimi (commented):
This is looking good, I've made a few small suggestions.

src/juju/jimm/api.ts
```diff
@@ -72,3 +80,40 @@ export const checkRelations = async (
...
+      conn.facades.jimM
+        .migrateModel(modelName, targetController)
+        .then((migrateModelResponse) => resolve(migrateModelResponse))
+        .catch((error) => reject(error));
+    } else {
+      reject(new Error(Label.NO_JIMM));
+    }
+  });
+}
+
```
andogq, line 100: These methods can be significantly simplified by just making them `async`. Same goes for `listMigrationTargets`.

```suggestion
export async function migrateModel(
  conn: ConnectionWithFacades,
  modelName: string,
  targetController: string,
) {
  if (!conn?.facades?.jimM) {
    throw new Error(Label.NO_JIMM);
  }
  
  return conn.facades.jimM
    .migrateModel(modelName, targetController);
}

```

src/juju/jimm/JIMMV4.ts
Ninfa-Jeon: Actually, this will need updates when we start working on the upgrade controller functionality. In that scenario, we will be migrating multiple models from a controller.

Ninfa-Jeon: I can, however, update the input parameter format to work with one or more models, though we will be using it for a one model at a time currently

src/juju/jimm/JIMMV4.ts
```diff
@@ -55,6 +55,16 @@ export type CheckRelationsResponse = {
...
+export type InitiateMigrationResult = {
+  error?: Error;
+  "migration-id": string;
+  "model-tag": string;
+};
+
+export type InitiateMigrationResults = {
+  results: InitiateMigrationResult[];
+};
+
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done!

src/juju/jimm/JIMMV4.ts
Ninfa-Jeon: Done!

src/juju/jimm/api.ts
```diff
@@ -24,51 +24,78 @@ export const endpoints = () => {
...
+    throw new Error(Label.NO_JIMM);
+  }
+  return await conn.facades.jimM.crossModelQuery(query);
 }
 
-export const checkRelation = async (
+/**
+ Check if a user has the specified relation to a resource. 
+ */
+export async function checkRelation(
```
huwshimi, line 53: Just a note that the dashboard typical uses arrow functions unless there are specific need for a different `this` scope (I know it was inconsistent in this file to begin with).

src/juju/jimm/api.ts
```diff
@@ -24,51 +24,78 @@ export const endpoints = () => {
...
-        .findAuditEvents(params)
-        .then((events) => resolve(events))
-        .catch((error) => reject(error));
-    } else {
-      reject(new Error(Label.NO_JIMM));
-    }
-  });
+  if (!conn.facades.jimM) {
+    throw new Error(Label.NO_JIMM);
+  }
```
huwshimi, line 33: As you're modifying all these anyway it would probably be a good time to do a little cleanup to create a wrapping function for these that does this check that jimm exists and then throws the error or calls the jimm method, rather than having this repeated in each function.

src/juju/jimm/JIMMV4.ts
huwshimi: OK that sounds good, it'll be more consistent with the JIMM API that way too.

Ninfa-Jeon: Done!

src/juju/jimm/api.ts
Ninfa-Jeon: Replaced with arrow functions

src/juju/jimm/api.ts
```diff
@@ -21,54 +43,76 @@ export const endpoints = () => {
...
+ * the provided async function.
+ */
+const withJimmCheck = async <T>(
+  conn: ConnectionWithFacades,
+  facadeMethod: (jimm: JIMMFacade) => Promise<T>,
+): Promise<T> => {
+  if (!conn.facades.jimM) {
+    throw new Error(Label.NO_JIMM);
+  }
+  return await facadeMethod(conn.facades.jimM);
```
andogq, line 57: You can directly return this `Promise` without `await`-ing it to avoid the extra task, but #meganit

andogq (approved):

src/juju/jimm/api.ts
```diff
@@ -1,7 +1,29 @@
...
+  crossModelQuery: (query: string) => Promise<CrossModelQueryResponse>;
+  checkRelation: (tuple: RelationshipTuple) => Promise<CheckRelationResponse>;
+  checkRelations: (
+    tuples: RelationshipTuple[],
+  ) => Promise<CheckRelationsResponse>;
+  migrateModel: (
+    migrationSpecs: MigrateModelInfo[],
+  ) => Promise<InitiateMigrationResults>;
+  listMigrationTargets: (modelTag: string) => Promise<ListControllersResponse>;
+};
```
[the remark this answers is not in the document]
Ninfa-Jeon: How about `NonNullable<Facades["jimM"]>`?

src/juju/jimm/api.ts
Ninfa-Jeon: Done!

huwshimi (approved):
Thanks for the improvements here!

PR #2038, opened 2025-08-14
fix: fix e2e infinite loop

huwshimi:
## Done

Don't refetch rebac relations if there is an error to prevent an infinite loop.

## Details

https://warthogs.atlassian.net/browse/WD-23889

### DIFF
--- a/src/hooks/useModelAccess.ts
+++ b/src/hooks/useModelAccess.ts
@@ -1,4 +1,4 @@
-import { useId } from "react";
+import { useRef } from "react";
 
 import { useCheckRelations } from "juju/api-hooks/permissions";
 import { JIMMRelation } from "juju/jimm/JIMMV4";
@@ -63,8 +63,8 @@ const useModelAccess = (modelUUID?: string | null, cleanup?: boolean) => {
   const permissions = useAppSelector((state) =>
     getReBACPermissions(state, relations),
   );
-  const requestId = useId();
-  useCheckRelations(requestId, relations, cleanup);
+  const requestId = useRef(new Date().getTime().toString());
+  useCheckRelations(requestId.current, relations, cleanup);
   if (isJuju) {
     return jujuAccess;
   }
--- a/src/juju/api-hooks/permissions.ts
+++ b/src/juju/api-hooks/permissions.ts
@@ -21,6 +21,8 @@ import {
   getReBACRelationshipsLoaded,
   getReBACRelationshipsLoading,
   getReBACPermissions,
+  getReBACRelationshipsErrors,
+  getReBACPermissionErrors,
 } from "store/juju/selectors";
 import { useAppSelector } from "store/store";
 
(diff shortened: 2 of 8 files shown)


src/hooks/useModelAccess.ts
```diff
@@ -63,8 +63,8 @@ const useModelAccess = (modelUUID?: string | null, cleanup?: boolean) => {
   const permissions = useAppSelector((state) =>
     getReBACPermissions(state, relations),
   );
-  const requestId = useId();
-  useCheckRelations(requestId, relations, cleanup);
+  const requestId = useRef(new Date().getTime().toString());
```
huwshimi, line 66: It turns out `useId` is only unique per component. This wasn't the problem, but it could have been in the future.

src/pages/ModelsIndex/ModelsIndex.tsx
```diff
@@ -62,21 +63,21 @@ export default function Models() {
   const modelsError = useAppSelector(getModelsError);
   const modelsLoaded = useAppSelector(getModelListLoaded);
   const hasSomeModels = useAppSelector(hasModels);
-  // loop model data and pull out filter panel data
   const modelData = useAppSelector(getModelData);
+  const modelUUIDs = useAppSelector(getModelUUIDs);
```
huwshimi, line 67: This was an improvement I made while debugging.

src/hooks/useModelAccess.ts
andogq: Maybe `crypto.randomUUID()` instead of a timestamp? It's unlikely, but timestamp may still conflict.

src/juju/api-hooks/permissions.ts
```diff
@@ -86,11 +92,15 @@ export const useCheckPermissions = (
...
   );
+  const errors = useAppSelector((state) =>
+    getReBACPermissionErrors(state, tuple),
+  );
   useReBAC(
     jujuActions.checkRelation,
     jujuActions.removeCheckRelation,
     loading,
     loaded,
+    !!errors,
```
andogq, line 103: What happens if this length is zero, instead of it just being nullish?

src/pages/ModelsIndex/ModelsIndex.tsx
```diff
@@ -62,21 +63,21 @@ export default function Models() {
...
   const { clouds, regions, owners, credentials } =
     useModelAttributes(modelData);
   const controllerUser = useAppSelector(getControllerUserTag);
-  const requestId = useId();
+  const requestId = useRef(new Date().getTime().toString());
```
andogq, line 71: Same here for crypto uuid.

andogq (approved):
Nice find! Small comment about using crypto random UUID, but looks good aside from that.

src/pages/ModelsIndex/ModelsIndex.tsx
huwshimi: Done.

src/hooks/useModelAccess.ts
huwshimi: Done

Ninfa-Jeon (approved):
Thanks, Huw!

PR #2057, opened 2025-08-22
refactor: Consolidate redundant code on models list page

Ninfa-Jeon:
## Done

- Created a new `Actions` column on the models list table.
- Moved the logic from within the `AccessColumn` component to a contextual menu in the new `ModelActions` component.
- The `ModelActions` menu will show the different model management actions that a user can trigger. If the user does not have sufficient permission, the respective option will appear disabled.
- Removed code for `AccessColumn`.
- Moved common code from `StatusGroup`, `OwnerGroup` and `CloudGroup` to a new `ModelTable` component.

## QA

- Check that the three models list views work correctly

## Details
https://warthogs.atlassian.net/browse/WD-25920


src/components/ModelActions/ModelActions.tsx
```diff
@@ -0,0 +1,55 @@
...
+    panel: string | null;
+  }>({
+    model: null,
+    panel: null,
+  });
+  const canConfigureModel = useCanConfigureModel(false, modelName, activeUser);
+  const isJuju = useAppSelector(getIsJuju);
+
+  return (
+    <ContextualMenu
```
Ninfa-Jeon, line 24: need to handle the case where the user does not have permission to view the access button. showing a menu with no options is poor; will aim to remove the entire column

src/components/ModelTableList/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+/**
+  Returns the model info and statuses in the proper format for the table data.
+  @param  models The list of models
+  @return The formatted table rows.
+*/
+function generateModelTableList(
+  models: ModelData[],
+  activeUsers: Record<string, string>,
+  controllers: Controllers | null,
+  groupBy: "status" | "owner" | "cloud",
```
huwshimi, line 38: It'd probably be a bit easier to maintain if these values were an enum so that we don't have to repeat this definition and also don't have to maintain magic strings.

src/components/ModelTableList/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+      ...(groupBy !== "owner"
+        ? [
+            {
+              "data-testid": TestId.COLUMN_OWNER,
+              content: (
+                <TruncatedTooltip message={owner}>{owner}</TruncatedTooltip>
+              ),
+            },
+          ]
+        : []),
```
huwshimi, line 84: I wonder if these would be a little more readable if we were to introduce a utility to conditionally display these. Maybe something like this?

```suggestion
      ...displayForGroups(groupBy, ["cloud", "status"], {
              "data-testid": TestId.COLUMN_OWNER,
              content: (
                <TruncatedTooltip message={owner}>{owner}</TruncatedTooltip>
              ),
            }),
```

What do you think?

src/components/ModelTableList/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,208 @@
...
+    headerOptions = {
+      showCloud: true,
+      showOwner: true,
+      ...(groupLabel === "Blocked" ? { showHeaderStatus: true } : {}),
+    };
+  } else if (groupBy === "owner") {
+    headerOptions = { showCloud: true, showStatus: true };
+  } else if (groupBy === "cloud") {
+    headerOptions = { showOwner: true, showStatus: true };
+  }
```
huwshimi, line 191: Just a small suggestion, if you think it's easier to read:

```suggestion
  const headerOptions = {
    showCloud: ["status", "owner"].includes(groupBy),
    showOwner: ["status", "cloud"].includes(groupBy),
    showStatus: ["owner", "cloud"].includes(groupBy),
      ...(groupLabel === "Blocked" && groupBy === "status" ? { showHeaderStatus: true } : {}),
  };
```

huwshimi (commented):
This is heading in a good direction, thanks! I've made a few small comments.

src/components/ModelTableList/ModelTable/ModelTable.tsx
Ninfa-Jeon: Thank you!

src/components/ModelTableList/ModelTable/ModelTable.tsx
Ninfa-Jeon: Done!

src/components/ModelActions/ModelActions.tsx
```diff
@@ -0,0 +1,54 @@
...
```
huwshimi, line 24: I know that the designs aren't final, but we can can make the button appear like the designs if we use the `p-icon--contextual-menu` icon and set the button appearance to base and also right align the action column.

Also, we should remove the margin at the bottom of the toggle button as it's adding extra space at the bottom of each row.

src/components/ModelActions/ModelActions.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+    renderComponent(
+      <ModelActions activeUser="eggman@external" modelName="test-model" />,
+    );
+    expect(
+      screen.getByRole("button", { name: "Toggle menu" }),
+    ).toBeInTheDocument();
+  });
+
+  it("shows option to manage access if user has permission", async () => {
+    const state = rootStateFactory.build({
```
huwshimi, line 33: It looks like the state for this and the test below is the same so it'd probably be good to set it up in a before each.

src/components/ModelActions/ModelActions.tsx
```diff
@@ -0,0 +1,54 @@
...
+      hasToggleIcon
+      links={[
+        {
+          children: Label.ACCESS,
+          className: "model-access",
```
huwshimi, line 29: I think we can just use the default styling for this so we can remove this class name and its styles.

src/components/ModelTableList/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,214 @@
...
+        content: (
+          <ModelSummary
+            modelData={model}
+            ownerTag={model.info?.["owner-tag"]}
+          />
+        ),
+        className: "u-overflow--visible",
+      },
+      // Conditionally include columns based on the group type
+      ...displayForGroups(groupBy, GroupBy.OWNER, {
```
huwshimi, line 96: I wonder if there's a name that might better fit this as it's excluding the group that's listed, whereas it reads like it's going to display it for the group. Maybe something like `excludeFromGroup`, `displayForOtherGroups` or `displayForSomeColumnsThatArentTheColumnThatsTheSecondArgumentOKGo`.

huwshimi (approved):
Thanks! Just a few small suggestions.

src/components/ModelTableList/ModelTable/ModelTable.tsx
Ninfa-Jeon: I REALLY like the third name

src/components/ModelActions/ModelActions.test.tsx
huwshimi: It's generally safer to set up state in a beforeEach so that we get a fresh state object for each test, otherwise anything that mutates state will cause subsequent tests to fail.

src/components/ModelActions/ModelActions.tsx
Ninfa-Jeon: Done!

src/components/ModelTableList/ModelTable/ModelTable.tsx
Ninfa-Jeon: `getConditionalCell` will never be as great as `displayForSomeColumnsThatArentTheColumnThatsTheSecondArgumentOKGo`

PR #2058, opened 2025-08-26
feat: env var configuration of dashboard image

huwshimi:
## Done

- Update the docker image to be able to be configured at run time using env vars.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- [Add additional steps]

## Details

https://warthogs.atlassian.net/browse/WD-26021


charms/k8s-charm/src/config.py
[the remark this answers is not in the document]
kian99: I think one option for K8s charms is to assume this config file exists in the container. Then the charm code can invoke it and the entrypoint code can also invoke it.

Unfortunately I don't think that works for machine charms, in which case if you take the mentioned route, you'd need to keep a copy of the file in the machine charm as you've done.

entrypoint
```diff
@@ -1,9 +1,16 @@
...
+# DASHBOARD_ROOT - the root directory that the dashboard is located in.
+# DASHBOARD_CONFIG_DIR - the location that the config.js.j2 and nginx.conf.j2 are located in.
+# DASHBOARD_IDENTITY_PROVIDER_URL - equivalent to the charm's identity-provider-url config option.
+# DASHBOARD_IS_JUJU - equivalent to the charm's is-juju config option.
+# DASHBOARD_ANALYTICS_ENABLED - equivalent to the charm's analytics-enabled config option.
+# DASHBOARD_PORT - equivalent to the charm's port config option.
+
+python3 /srv/config.py
 
 nginx -g "daemon off;"
```
kian99, line 16: I noticed that when I would do `docker compose down`, the dashboard container wouldn't exit immediately, only after 10 seconds. From docker docs,
>When executing docker compose down, Docker Compose by default sends a SIGTERM signal to containers, allowing them a grace period to shut down gracefully. If a container does not terminate within this grace period, a SIGKILL signal is sent to forcefully stop it. The default grace period for docker compose down is 10 seconds.

There is an easy fix. The common thing is to do `exec <your app>` in the entrypoint see https://stackoverflow.com/questions/32255814/what-purpose-does-using-exec-in-docker-entrypoint-scripts-serve.

kian99 (approved):
lgtm, thanks for this!

entrypoint
huwshimi: Fixed.

.github/workflows/e2e-jimm-docker-keycloak.yml
```diff
@@ -73,6 +73,8 @@ jobs:
           timeout_minutes: 1
           max_attempts: 2
           command: while [ "$(docker inspect -f {{.State.Restarting}} jimm)" == "true" ]; do sleep 1; done
+      - name: Remove the JIMM deployed dashboard
+        run: docker container stop juju-dashboard && docker container rm juju-dashboard
```
huwshimi, line 77: @kian99 Would it be possible to have jimm's docker-compose ([here](https://github.com/canonical/jimm/pull/1684/files#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R142)) be able to deploy from a local docker image using an env var or something? That would mean we could just build the image and let jimm handle the rest.

PR #2063, opened 2025-09-02
fix: alternate workaround for `node-util`, to remove `vite-plugin-node-polyfills`

andogq:
## Done

- remove `vite-plugin-node-polyfills`

### Development

- Ensure `identityProviderURL` is set in `config.local.js`, so that Candid is used
- `yarn start`
- Login with Candid

### Production

- `yarn build`
- `cp -f public/config.local.js build/config.js`
- `yarn vite preview`
- Login with Candid

## Details

Vite will inject `process.env.NODE_DEBUG` into the global scope in development, and will replace all occurrences during build. Ideally the dependency would use `import.meta` instead so vite could do this automatically, but it's quite old and unmaintained.

Will close #2055, close #2054 and close #1994 as those deps are removed.

### DIFF
--- a/e2e/fixtures/setup.ts
+++ b/e2e/fixtures/setup.ts
@@ -52,6 +52,8 @@ export type Resource = {
   owner?: string;
 };
 
+const JUJU_UPDATE_URL = "https://juju.is/latest.json";
+
 export const test = base.extend<Fixtures>({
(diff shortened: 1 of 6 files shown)


huwshimi (approved):
Thanks this is a big improvement.

PR #2065, opened 2025-09-02
fix(ci): use `pull_request_target` for PR actions

andogq:
For the actions that create a PR, use the `pull_request_target` event so that they can run with write permissions when triggered by pull requests that originate from a fork.


Ninfa-Jeon (approved):
LGTM

PR #2077, opened 2025-09-08
fix: log error config error

huwshimi:
## Done

- Log error instead of exiting when initial charm config is run. When deployed by a charm this config will be done later by the charm code so it's expected to fail the initial config (the env vars are only provided when running the docker image with docker compose).

## QA

- Pull code
- Open http://0.0.0.0:8036/
- [Add additional steps]

## Details

Fixes the issue here: https://github.com/canonical/juju-dashboard/actions/runs/17535094432/job/49797121189#step:12:382.

### DIFF
--- a/charms/k8s-charm/src/config.py
+++ b/charms/k8s-charm/src/config.py
@@ -1,6 +1,10 @@
+import logging
 import os
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


Ninfa-Jeon (approved):
Thanks for fixing this, Huw!

PR #2078, opened 2025-09-08
chore(lint): add no-redeclare rule

huwshimi:
## Done

- add [no-redeclare](https://typescript-eslint.io/rules/no-redeclare/) rule (there is a warning on the docs page about using with typescript but https://github.com/typescript-eslint/typescript-eslint/issues/11461 suggests that the rule is still valid).
- fix issues

### DIFF
--- a/src/components/Panel/Panel.tsx
+++ b/src/components/Panel/Panel.tsx
@@ -10,7 +10,7 @@ import type { AsideProps } from "components/Aside";
 
 type Props = PropsWithSpread<
   {
-    checkCanClose?: (e: KeyboardEvent | React.MouseEvent) => boolean;
+    checkCanClose?: (e: React.KeyboardEvent | React.MouseEvent) => boolean;
     drawer?: ReactNode;
     panelClassName?: string;
     splitContent?: ReactNode;
@@ -23,7 +23,7 @@ type Props = PropsWithSpread<
 const close = {
   // Close panel if Escape key is pressed when panel active
   onEscape: (
-    ev: KeyboardEvent,
+    ev: React.KeyboardEvent,
     onRemovePanelQueryParams: () => void,
     checkCanClose?: Props["checkCanClose"],
   ) => {
@@ -65,7 +65,7 @@ const Panel = forwardRef<HTMLDivElement, Props>(
   ) => {
     useListener(
       window,
-      (ev: KeyboardEvent) =>
+      (ev: React.KeyboardEvent) =>
         close.onEscape(ev, onRemovePanelQueryParams, checkCanClose),
       "keydown",
     );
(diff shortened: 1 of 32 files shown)


src/panels/ConfigPanel/BooleanConfig/BooleanConfig.tsx
```diff
@@ -12,7 +12,7 @@ export default function BooleanConfig({
   validate,
 }: ConfigProps): JSX.Element {
   function handleOptionChange(
-    ev: FormEvent<HTMLInputElement> | MouseEvent<HTMLInputElement>,
+    ev: FormEvent<HTMLInputElement> | React.MouseEvent<HTMLInputElement>,
```
Ninfa-Jeon, line 15:
```suggestion
    ev: React.FormEvent<HTMLInputElement> | React.MouseEvent<HTMLInputElement>,
```

src/panels/ConfigPanel/ConfigPanel.tsx
```diff
@@ -187,7 +187,7 @@ export default function ConfigPanel(): JSX.Element {
     setEnableSave(fieldChanged);
   }
 
-  function checkCanClose(event: KeyboardEvent | MouseEvent) {
+  function checkCanClose(event: KeyboardEvent | React.MouseEvent) {
```
Ninfa-Jeon, line 190:
```suggestion
  function checkCanClose(event: React.KeyboardEvent | React.MouseEvent) {
```

Ninfa-Jeon (approved):
I'm a bit confused as to why in some files the type `KeyboardEvent` is not required to be imported while in one of these, it was. I see more instances of this and `FormEvent` that didn't even need changing like we did for `MouseEvent`. Any idea why this discrepancy?

src/components/Panel/Panel.tsx
```diff
@@ -2,15 +2,15 @@ import type { PropsWithSpread } from "@canonical/react-components";
...
+import type { ReactNode } from "react";
 import { forwardRef, useId } from "react";
 
 import Aside from "components/Aside";
 import type { AsideProps } from "components/Aside";
 
 type Props = PropsWithSpread<
   {
-    checkCanClose?: (e: KeyboardEvent | MouseEvent) => boolean;
+    checkCanClose?: (e: KeyboardEvent | React.MouseEvent) => boolean;
```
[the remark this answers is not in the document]
huwshimi: You're right, I thought it wouldn't add the React attributes to listeners on the window, but I just checked and it does.

src/panels/ConfigPanel/BooleanConfig/BooleanConfig.tsx
huwshimi: I think this one is OK because there's no builtin type to conflict with.

src/panels/ConfigPanel/ConfigPanel.tsx
huwshimi: Done.

PR #2084, opened 2025-09-09
chore(lint): apply strict-boolean-expressions rule

huwshimi:
## Done

- Add the [strict-boolean-expressions](https://typescript-eslint.io/rules/strict-boolean-expressions/) rule.
- Fix issues.

### DIFF
--- a/src/components/hooks.test.tsx
+++ b/src/components/hooks.test.tsx
@@ -99,7 +99,10 @@ describe("useModelIndexParams", () => {
         />
       ),
     });
-    expect(result.current).toStrictEqual({});
+    expect(result.current).toStrictEqual({
+      modelName: null,
+      userName: null,
+    });
   });
 });
 
@@ -132,7 +135,11 @@ describe("useModelAppParams", () => {
         />
       ),
     });
-    expect(result.current).toStrictEqual({});
+    expect(result.current).toStrictEqual({
+      appName: null,
+      modelName: null,
+      userName: null,
+    });
   });
 });
 


src/pages/EntityDetails/counts.ts
```diff
@@ -25,12 +25,12 @@ const generateOfferCounts = (modelStatusData: ModelData) => {
   return { joined: offerCount };
 };
 
-const generateSecondaryCounts = <M = ModelData>(
+const generateSecondaryCounts = <M extends ModelData>(
```
andogq, line 28: This is has different semantics to setting the default type. Is this intentional?

src/testing/utils.tsx
```diff
@@ -174,11 +174,11 @@ export function createStore(
 ): [EnhancedStore<RootState>, UnknownAction[]];
 export function createStore(
   state: RootState,
-  options?: { trackActions?: false },
+  options?: { trackActions: false },
```
andogq, line 177: should this be `boolean`, or must it be `false` if it's provided?

andogq (approved):
i've got absolutely nothing constructive to add here :(

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -7,17 +7,17 @@ import { ModelTab } from "urls";
...
-    unitId,
-    machineId,
-    isNestedEntityPage,
+    userName = null,
+    modelName = null,
+    appName = null,
+    unitId = null,
+    machineId = null,
+    isNestedEntityPage = false,
   } = useEntityDetailsParams();
```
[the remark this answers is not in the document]
huwshimi: Done. Some of them are hooks from external libraries so can't change the return values, so I've left those ones (e.g. `useParams`).

src/pages/EntityDetails/counts.ts
huwshimi: Yeah, the type doesn't actually need a default as it's not optional and with my change to use `segment in modelStatusData` it needed to know that `M` was an object (which it does when it uses extends).

src/components/Breadcrumb/Breadcrumb.tsx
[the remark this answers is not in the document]
andogq: Thanks Huw! And that's no problem, very reasonable

PR #2091, opened 2025-09-09
fix: correctly look up model permissions

huwshimi:
## Done

- Fixes checking model permissions (it was trying to find a model owned by the current user instead of the model's real owner).

## QA

- Create a model and share it with another user (giving them write/admin permissions).
- Log in as the second user and go to the model list.
- Open the actions menu for the model.
- Check that the access button is enabled.


Ninfa-Jeon (approved):
Great work!

PR #2092, opened 2025-09-09
feat(deps): upgrade dependencies to fix vulnerabilities

Ninfa-Jeon:
## Done

- Upgraded `@playwright/test`
- Set correct versions for packages like `ops`, `jinja2` and `pydantic` in `requirements.txt` for the charms

## QA
- All E2E tests should pass as the affected dependencies are either `playwright` or charm-related that should pretty much be tested when they are built.

## Details

Scanned the dashboard using [safedep](https://github.com/safedep/vet) and tried to fix the reported vulnerabilities (especially the ones with a high impact score).

https://warthogs.atlassian.net/browse/WD-26682

## Screenshots

Before:
Screenshot 2025-09-09 at 2 55 14 PM

After:
Screenshot 2025-09-10 at 9 01 22 AM


huwshimi (approved):
Thanks for this. I think it would be safer to use `==` for the version numbers so we don't accidentally upgrade to breaking versions.

PR #2093, opened 2025-09-09
fix: Deleted models are not updated

Ninfa-Jeon:
## Done

- Fixed the bug where a deleted model still appeared in the models list until the dashboard was reloaded

## QA

- Add one model
- Verify that it is available on the dashboard's models list page
- Delete it
- Without reloading the dashboard, check with the network tab that the deleted model is removed as soon as the `listModels` endpoint is polled next
- Do the same with more than one model added
- Do the same with multiple models deleted at once

## Details

https://warthogs.atlassian.net/browse/WD-26700

### DIFF
--- a/src/store/juju/slice.ts
+++ b/src/store/juju/slice.ts
@@ -142,18 +142,24 @@ const slice = createSlice({
       state,
       action: PayloadAction<{ models: UserModelList } & WsControllerURLParam>,
     ) => {
-      const modelList = state.models;
+      const modelList: JujuState["models"] = {};
+      const modelDataList: JujuState["modelData"] = {};
       const userModels = action.payload.models["user-models"] ?? [];
       userModels.forEach((model) => {
-        modelList[model.model.uuid] = {
+        const uuid = model.model.uuid;
+        modelList[uuid] = {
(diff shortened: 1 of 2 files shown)


src/store/juju/slice.ts
```diff
@@ -142,18 +142,24 @@ const slice = createSlice({
       state,
       action: PayloadAction<{ models: UserModelList } & WsControllerURLParam>,
     ) => {
-      const modelList = state.models;
+      const modelList: JujuState["models"] = {};
+      const modelDataList: JujuState["modelData"] = {};
```
huwshimi, line 146: I think it could be helpful to explain why we recreate these, something about only including models that are in the updated list so that destroyed models are removed from the store.

huwshimi (commented):
Good work figuring out where this bug was coming from. This needs a couple of tests so that we're checking it removes destroyed models and that existing model data is retained.

src/store/juju/slice.ts
Ninfa-Jeon: I've added a comment to explain this. Thanks!

huwshimi (approved):

PR #2099, opened 2025-09-10
chore(lint): apply prefer-destructuring rule

huwshimi:
## Done

- Add prefer-destructuring rule.
- Fix issues (some were fixed by `--fix`).


src/components/DevBar/widgets/controller.test.tsx
```diff
@@ -70,7 +70,7 @@ describe("Widget", () => {
     const { result } = renderComponent(<Widget />, { state: connectedState() });
 
     expect(result.container.children).toHaveLength(2);
-    const grid = result.container.children[1];
+    const [, grid] = result.container.children;
```
andogq, line 73: I _really_ don't like this, as it feels like it's kinda going against our 1-char identifier rule (even though we're not binding to an identifier here).

Would it be possible to enforce binding to an identifier in these situations, even if the identifier is never used (such as if we prefix it with an `_`, it will indicate it's meant to be ignored). I feel this will help with clarity, especially if we're destructuring deeper into an array (eg `[,, third]`).

src/panels/ActionsPanel/ActionOptions/ActionOptions.tsx
```diff
@@ -32,7 +32,7 @@ export default function ActionOptions({ name, data, onValuesChange }: Props) {
       });
     });
     return collectOptions;
-  }, [action.params.properties, action.params.required]);
+  }, [action.params]);
```
andogq, line 35: Are we positive this will always be updated with a new reference, instead of using the actual fields? Additionally could this cause the hook to needlessly re-run?

andogq (commented):
My biggest concern is array destructuring. I'd be interested in some other rule that enforces we bind array elements when destructuring. Eg:

```ts
// Wrong
const [,, third] = [1, 2, 3];

// Right
const [_first, _second, third] = [1, 2, 3];
```

src/components/DevBar/widgets/controller.test.tsx
huwshimi: Done.

src/panels/ActionsPanel/ActionOptions/ActionOptions.test.tsx
```diff
@@ -51,7 +51,7 @@ describe("ActionOptions", () => {
       />,
     );
 
-    const props = OptionsInputSpy.mock.calls[0][0];
+    const [[props]] = OptionsInputSpy.mock.calls;
```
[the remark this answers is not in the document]
huwshimi: Done.

src/panels/ActionsPanel/ActionOptions/ActionOptions.tsx
huwshimi: Done.

andogq:
thanks for those changes Huw!

PR #2109, opened 2025-09-12
chore(lint): add require-await rule

huwshimi:
## Done

- Added the [require-await](https://typescript-eslint.io/rules/require-await/) rule.
- Disabled the rule for tests (lots of mocked things return a value instead of a promise).
- Fixed issues.

### DIFF
--- a/src/auth/Auth.ts
+++ b/src/auth/Auth.ts
@@ -71,6 +71,7 @@ export class Auth {
    *
    * @returns Whether to continue the connection attempt.
    */
+  // eslint-disable-next-line @typescript-eslint/require-await
   async beforeControllerConnect(
     controllerData: ControllerData,
   ): Promise<boolean> {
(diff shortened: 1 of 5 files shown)


src/auth/Auth.ts
```diff
@@ -71,6 +71,7 @@ export class Auth {
    *
    * @returns Whether to continue the connection attempt.
    */
+  // eslint-disable-next-line @typescript-eslint/require-await
```
andogq, line 74:
```ts
async beforeControllerConnect(
controllerData: ControllerData,
): Promise<boolean> {
return Promise.resolve(true);
}
```

could save us the directive?

andogq (approved):

andogq:
tiiiiny suggestion but not a big deal

src/auth/Auth.ts
huwshimi: Done.

PR #2122, opened 2025-09-29
feat: Destroy model middleware

Ninfa-Jeon:
## Done

- Added a new type of ToastCard - "info"
- Added `destroyModels` slice and related reducers + selector
- Added middleware support for `destroy-model` in `model-poller.ts`
- Extended `fullStatus` call to include storage in response
- Extended `fetchModelInfo` to query batch status for models
- Added unit tests

## QA

- Make sure the functionality and UI remains unaffected

## Details

https://warthogs.atlassian.net/browse/WD-27697

### DIFF
--- a/src/store/middleware/model-poller.ts
+++ b/src/store/middleware/model-poller.ts
@@ -538,6 +538,7 @@ export const modelPollerMiddleware: Middleware<
             (modelUUID) => !destroyedModels.includes(modelUUID),
           );
 
+          // TODO: Clear this timeout. Refer: WD-29374
           // Wait 10s then start again.
           await new Promise((resolve) => {
             setTimeout(() => {
--- a/src/store/middleware/model-poller.test.ts
+++ b/src/store/middleware/model-poller.test.ts
@@ -1,4 +1,5 @@
 import type { Client, Connection, Transport } from "@canonical/jujulib";
+import { waitFor } from "@testing-library/dom";
 import type { UnknownAction, MiddlewareAPI, Dispatch } from "redux";
 import type { Mock, MockInstance } from "vitest";
 import { vi } from "vitest";
@@ -19,6 +20,7 @@ import { auditEventFactory } from "testing/factories/juju/jimm";
 import {
   controllerFactory,
   jujuStateFactory,
+  modelDataInfoFactory,
   relationshipTupleFactory,
 } from "testing/factories/juju/juju";
 
@@ -1111,12 +1113,87 @@ describe("model poller", () => {
       ],
     });
     await middleware(next)(action);
-    await vi.advanceTimersByTimeAsync(10000);
     expect(fakeStore.dispatch).toHaveBeenCalledWith(
       jujuActions.destroyModelErrors({
         errors: [["123abc", "Error"]],
       }),
     );
+    expect(fetchModelInfo).toHaveBeenCalledWith(conn, ["456xyz"]);
+    expect(fakeStore.dispatch).toHaveBeenCalledWith(
+      jujuActions.updateModelsDestroyed({
+        modelUUIDs: ["456xyz"],
+        wsControllerURL: "wss://example.com",
+      }),
+    );
+  });
+
+  it("handles destroy model action for multiple models", async () => {
+    vi.spyOn(jujuModule, "loginWithBakery").mockImplementation(async () => ({
+      conn,
+      intervalId,
+      juju,
+    }));
+    conn.facades.modelManager.destroyModels.mockResolvedValue({
+      results: [{}, {}],
+    });
+    /** 
+     We expect fetchModelInfo to be called twice:
+     1. With both models.
+     2. With the one remaining model after 10s.
+     * */
+    const fetchModelInfo = vi
+      .spyOn(jujuModule, "fetchModelInfo")
+      .mockResolvedValueOnce({
+        results: [
+          { result: undefined },
+          { result: modelDataInfoFactory.build({ life: "dying" }) },
+        ],
+      })
+      .mockResolvedValue({
+        results: [{ result: undefined }],
+      });
+
+    const middleware = await runMiddleware();
+    const action = jujuActions.destroyModels({
+      wsControllerURL: "wss://example.com",
+      models: [
+        { "model-tag": "model-123abc", modelUUID: "123abc" },
+        { "model-tag": "model-456xyz", modelUUID: "456xyz" },
+      ],
+    });
+    middleware(next)(action);
+    await waitFor(() => {
+      expect(conn.facades.modelManager.destroyModels).toHaveBeenCalledWith({
+        models: [
+          { "model-tag": "model-123abc", modelUUID: "123abc" },
+          { "model-tag": "model-456xyz", modelUUID: "456xyz" },
+        ],
+      });
+      expect(fakeStore.dispatch).toHaveBeenCalledWith(
+        jujuActions.updateDestroyModelsLoading({
+          modelUUIDs: ["123abc", "456xyz"],
+          wsControllerURL: "wss://example.com",
+        }),
+      );
+    });
+
+    // Check that the first model was successfully destroyed in Poll 1
+    expect(fetchModelInfo).toHaveBeenCalledWith(conn, ["123abc", "456xyz"]);
+    expect(fakeStore.dispatch).toHaveBeenCalledWith(
+      jujuActions.updateModelsDestroyed({
+        modelUUIDs: ["123abc"],
+        wsControllerURL: "wss://example.com",
+      }),
+    );
+
+    // This resolves the 10s promise, triggers Poll 2
+    await vi.advanceTimersByTimeAsync(10000);
+
+    // The middleware should have called fetchModelInfo exactly twice:
+    // Once before the timer (Poll 1), once after the timer (Poll 2).
+    expect(fetchModelInfo).toHaveBeenCalledTimes(2);
+
+    // Check that the second fetch call (Poll 2) targeted the remaining model
     expect(fetchModelInfo).toHaveBeenCalledWith(conn, ["456xyz"]);
     expect(fakeStore.dispatch).toHaveBeenCalledWith(
       jujuActions.updateModelsDestroyed({


src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+                resolve(true);
+              }, 1000);
+            });
+          } while (remainingModels.length > 0 && !isDestructionComplete);
+        }
+      } catch (errors) {
+        let errorMessages: DestroyModelErrors = [];
+
+        if (isDestroyModelErrors(errors)) {
+          errorMessages = errors;
```
huwshimi, line 540: I'm probably missing it, but I can't see how these errors would match the `DestroyModelErrors` type as it looks like they're constructed and dispatched inside the `try` rather than thrown.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+          }
+        });
+
+        // Dispatch partial errors, if any, and continue with successful destructions
+        if (errors.length > 0) {
+          reduxStore.dispatch(jujuActions.destroyModelErrors({ errors }));
+        }
+
+        // Only proceed to check for completion if at least one model had no errors
+        if (errors.length < modelParams.length) {
```
huwshimi, line 485: I think there needs to be a test that it does not poll if there are no successful requests.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+          reduxStore.dispatch(
+            jujuActions.updateDestroyModelsLoading({
```
huwshimi, line 487: I think there needs to be a test that it begins polling.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+                if (info.result?.life === undefined) {
+                  acc.push(remainingModels[index]);
+                }
+                return acc;
+              }, [] as string[]) ?? [];
+
+            if (destroyedModels?.length === remainingModels.length) {
+              isDestructionComplete = true;
+              reduxStore.dispatch(
+                jujuActions.updateModelsDestroyed({
```
huwshimi, line 508: I think it needs a test for when all models have been destroyed

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+                  wsControllerURL,
+                }),
+              );
+              break;
+            }
+
+            if (destroyedModels && destroyedModels.length > 0) {
+              reduxStore.dispatch(
+                jujuActions.updateModelsDestroyed({
+                  modelTags: destroyedModels,
```
huwshimi, line 519: This needs a test for when some models get destroyed.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+                  wsControllerURL,
+                }),
+              );
+            }
+            remainingModels = remainingModels.filter(
```
huwshimi, line 524: There also needs to be a test that it polls again if not all models get destroyed.

huwshimi (commented):
This is looking good @Ninfa-Jeon, mostly just some small comments and a few additional tests.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+            ]);
+          } else {
+            remainingModels.push(modelParams[index]["model-tag"]);
+          }
+        });
+
+        // Dispatch partial errors, if any, and continue with successful destructions
+        if (errors.length > 0) {
+          reduxStore.dispatch(jujuActions.destroyModelErrors({ errors }));
+        }
```
andogq, line 482: Should we always set these, to clear any pre-existing errors if nothing goes wrong? If so, should we do this at the start of the handler before we do the `destroyModel` call?

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+                  modelTags: destroyedModels,
+                  wsControllerURL,
+                }),
+              );
+            }
+            remainingModels = remainingModels.filter(
+              (modelTag) => !destroyedModels?.includes(modelTag),
+            );
+
+            // Wait 1s then start again.
```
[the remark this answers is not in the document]
andogq: More generally, could we have poll frequency as a global option somewhere? Either as a constant or something in the `config.js`?

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+            await new Promise((resolve) => {
+              setTimeout(() => {
+                resolve(true);
+              }, 1000);
+            });
+          } while (remainingModels.length > 0 && !isDestructionComplete);
```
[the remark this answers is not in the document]
andogq: Could we maybe include some additional feedback to the user about what's happening? Eg. after 30s have an info toast stating that it's taking a while, with an option to stop/abort/whatever.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+        }
+      } catch (errors) {
+        let errorMessages: DestroyModelErrors = [];
+
+        if (isDestroyModelErrors(errors)) {
+          errorMessages = errors;
+        } else {
+          errorMessages = modelParams.map((modelParam) => [
```
andogq, line 542: It may be good to preserve the underlying `errors`, to assist with debugging. Especially if we don't know what's gone wrong.

andogq (commented):
A few small general suggestions. Nice work on this!

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+            if (destroyedModels && destroyedModels.length > 0) {
+              reduxStore.dispatch(
+                jujuActions.updateModelsDestroyed({
+                  modelTags: destroyedModels,
+                  wsControllerURL,
+                }),
+              );
+            }
+            remainingModels = remainingModels.filter(
+              (modelTag) => !destroyedModels?.includes(modelTag),
```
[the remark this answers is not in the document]
Ninfa-Jeon: sorry about that...i think i wrote that piece of code when there was a chance for it to be `undefined` but then i changed it and never reverted these. Thanks for noticing.

src/store/middleware/model-poller.ts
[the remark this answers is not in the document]
Ninfa-Jeon: This is already brought up with UX and we can work on it if/when they give us a green signal

src/store/middleware/model-poller.ts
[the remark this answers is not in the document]
Ninfa-Jeon: @andogq can be taken up in another ticket perhaps?

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +442,118 @@ export const modelPollerMiddleware: Middleware<
...
+
+            // Wait 1s then start again.
+            await new Promise((resolve) => {
+              setTimeout(() => {
+                resolve(true);
+              }, 1000);
+            });
+          } while (remainingModels.length > 0 && !isDestructionComplete);
+        }
+      } catch (errors) {
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done!...thanks that simplified a lot of things and also surfaced a mocking error in one of the tests

src/store/middleware/model-poller.ts
Ninfa-Jeon: Do you mean in case a use retries destroy operation?

Ninfa-Jeon: Sounds good. Will do

src/store/middleware/model-poller.ts
andogq: Maybe @huwshimi has a better idea, but yeah I think even logging would be good.

You're right in that we don't do this for other operations, but I'm thinking since this is something that _mutates_ the controller (rather than just reading from the controller), it may be good for us to be able to see what went wrong without having to go back and manually add in logging later.

huwshimi: I think logging would be helpful, you can see how this is done for some of the other catches in this file.

src/store/middleware/model-poller.ts
Ninfa-Jeon: No, you're right. I was throwing them previously but then that approach didn't work and now we have useless code lines like these. Thanks for flagging haha

src/juju/api.ts
```diff
@@ -303,11 +306,12 @@ export async function fetchAndStoreModelStatus(
...
 export async function fetchModelInfo(
   conn: ConnectionWithFacades,
-  modelUUID: string,
+  modelTags: string[],
 ): Promise<ModelInfoResults | undefined> {
-  const modelInfo = await conn.facades.modelManager?.modelInfo({
-    entities: [{ tag: `model-${modelUUID}` }],
-  });
+  const entities = modelTags.map((modelTag) => ({
+    tag: modelTag,
```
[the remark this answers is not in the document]
Ninfa-Jeon: Done. Thanks!

src/store/middleware/model-poller.ts
Ninfa-Jeon: Done.

src/store/middleware/model-poller.ts
```diff
@@ -440,6 +441,116 @@ export const modelPollerMiddleware: Middleware<
...
+              }),
+            );
+          }
+          remainingModels = remainingModels.filter(
+            (modelUUID) => !destroyedModels.includes(modelUUID),
+          );
+
+          // Wait 10s then start again.
+          await new Promise((resolve) => {
+            setTimeout(() => {
```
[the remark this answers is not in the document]
Ninfa-Jeon: Created a ticket for this here: https://warthogs.atlassian.net/browse/WD-29374

PR #2125, opened 2025-10-02
feat: enable rebac permissions checking

huwshimi:
## Done

- Check rebac permissions when using JIMM without needing to enable the flag (ReBAC Admin is still behind the flag).

## QA

- Connect to JIMM and make sure you don't have the rebac feature flag enabled.
- Check that the Permissions item is not in the side nav.
- Check that permissions are checked (i.e. you have audit logs in the side nav, have model access buttons etc.)
- Enable the rebac flag and the permissions item should appear in the side nav.

## Details

https://warthogs.atlassian.net/browse/WD-28065

### DIFF
--- a/src/components/PrimaryNav/PrimaryNav.test.tsx
+++ b/src/components/PrimaryNav/PrimaryNav.test.tsx
@@ -382,7 +382,48 @@ it("should not show Permissions navigation button if the controller doesn't supp
   ).not.toBeInTheDocument();
 });
 
-it("should show Permissions navigation button if the controller supports it", () => {
+it("should not show Permissions navigation button if the feature flag is not enabled", () => {
+  const state = rootStateFactory.build({
+    general: generalStateFactory.build({
+      config: configFactory.build({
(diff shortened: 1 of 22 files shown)


src/components/PrimaryNav/PrimaryNav.test.tsx
```diff
@@ -382,7 +382,48 @@ it("should not show Permissions navigation button if the controller doesn't supp
...
+    }),
+  });
+  renderComponent(<PrimaryNav />, { state });
+  expect(
+    screen.queryByRole("link", { name: Label.PERMISSIONS }),
+  ).not.toBeInTheDocument();
+});
+
+it("should show Permissions navigation button if the controller supports it and the feature flag is enabled", () => {
+  localStorage.setItem("flags", JSON.stringify(["rebac"]));
```
andogq, line 426: Is local storage automatically cleared between tests?

andogq (approved):
Tiny question about `localStorage` in tests, but otherwise looks good! I'll run it locally a little later.

PR #2135, opened 2025-10-19
feat: coordinate destroy model between CLI and UI

Ninfa-Jeon:
## Done

- Added the capability to pick up destroy-model action triggered from the CLI

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Add one or more models and deploy a few applications in them to prolong the destruction process. Alternately, you could also lower the poll interval in the dashboard.
- Trigger destroy on them from the CLI (try to pace them close to each other so that the dashboard picks up all of them in the next poll cycle).
- See that the dashboard reflects the destroy state with appropriate notifications.

## Details

https://warthogs.atlassian.net/browse/JUJU-9340

## Screenshots

https://github.com/user-attachments/assets/7306e3a1-974b-4c50-aa71-5c049ed665c9

### DIFF
--- a/src/store/middleware/model-poller.test.ts
+++ b/src/store/middleware/model-poller.test.ts
@@ -1172,6 +1172,123 @@ describe("model poller", () => {
     );
   });
 
+  it("batches models already destroying outside the dashboard by controller", async () => {
+    vi.spyOn(jujuModule, "loginWithBakery").mockImplementation(async () => ({
+      conn,
+      intervalId,
+      juju,
+    }));
+    vi.spyOn(jujuModule, "fetchAndStoreModelStatus").mockResolvedValue();
+    vi.spyOn(jujuModule, "fetchModelInfo")
+      .mockResolvedValueOnce({
+        results: [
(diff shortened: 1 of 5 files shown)


src/store/middleware/model-poller.test.ts
```diff
@@ -1172,6 +1172,123 @@ describe("model poller", () => {
...
+          },
+        ],
+      });
+
+    const middleware = await runMiddleware();
+    vi.spyOn(fakeStore, "getState").mockImplementation(() =>
+      rootStateFactory.build({
+        juju: jujuStateFactory.build({
+          models: {
+            abc123: {
```
huwshimi: You can use the `modelListInfoFactory` factory for these.

huwshimi (approved):
This seems to work well. Just a small comment about using factories.

src/store/middleware/model-poller.test.ts
Ninfa-Jeon: oops...done!

PR #2236, opened 2026-02-25
chore(ci): deploy iam model using terraform

huwshimi:
## Done

- Update IAM deployment to use a Terraform plan.
- Add new JIMM add-model permissions.
- Add support for /jimm-jimm path

## Details

https://warthogs.atlassian.net/browse/WD-34155

### DIFF
--- a/.github/actions/failure-log/action.yml
+++ b/.github/actions/failure-log/action.yml
@@ -51,6 +51,10 @@ runs:
         echo "-------------------------------------"
         juju status -m iam || true
         echo "-------------------------------------"
+        echo "IAM core status"
+        echo "-------------------------------------"
+        juju status -m core || true
+        echo "-------------------------------------"
         echo "JIMM status"
         echo "-------------------------------------"
         juju status -m ${{ inputs.controller }}:jimm || true
--- a/.github/actions/setup-jimm/action.yml
+++ b/.github/actions/setup-jimm/action.yml
@@ -21,9 +21,12 @@ outputs:
   jimm-controller-name:
     description: Name of the JIMM workloads controller
     value: ${{ steps.jimm-controller.outputs.name }}
-  workloads-controller-name:
+  workloads-controller-public-name:
     description: Name of the JIMM workloads controller
     value: ${{ steps.workloads-controller.outputs.public }}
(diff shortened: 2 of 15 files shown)


.github/actions/failure-log/action.yml
```diff
@@ -51,6 +51,10 @@ runs:
         echo "-------------------------------------"
         juju status -m iam || true
         echo "-------------------------------------"
+        echo "IAM core status"
```
huwshimi: The IAM deployment now has two models so this outputs the status of the second model.

scripts/wait-for
```diff
@@ -58,12 +58,5 @@ while true; do
     fi
   done
 
-  # Check for error states
-  if [ "$STATUS" == "error" ]; then
-    echo "ERROR: $TYPE $NAME is in error state"
-    juju status "$NAME"
-    exit 1
```
huwshimi, line 65: Sometimes an app goes into an error state before it goes into the state we're expecting, or the state we're expecting IS error so we don't want the script to fail in either of these cases.

.github/actions/setup-jimm/action.yml
```diff
@@ -130,33 +147,33 @@ runs:
...
       with:
-        timeout_minutes: 1
+        timeout_minutes: 5
         max_attempts: 5 # Retry the wait-for with a short timeout in case it gets stuck and times out. If it fails after multiple tries it's probably a real error.
-        command: |
+        on_retry_command: |
           STATUS=$(juju status --format json | yq .applications.openfga.application-status.current)
-          if [ "$STATUS" == "error" ]; then
+          if [ "$STATUS" == "blocked" ]; then
             # Sometimes the postgresql unit has a PrematureDataAccessError error so remove the relation and try again.
```
Ninfa-Jeon, line 155: Just asking, but does this comment need updating?

Ninfa-Jeon (approved):
Thanks for this, Huw!

.github/actions/setup-jimm/action.yml
huwshimi: I think it's still correct, just that error handling must have changed in the charm so now the status becomes `blocked` rather than `error`.

PR #2280, opened 2026-04-09
feat(middleware): add middleware and selectors for `modelMigrationTargets`

andogq:
## Done

- Create slice and reducers for `modelMigrationTargets`
- Create middleware for `listMigrationTrgets` JIMM call

## Details

JUJU-9577

### DIFF
--- a/src/store/middleware/source/migration-targets.ts
+++ b/src/store/middleware/source/migration-targets.ts
@@ -7,10 +7,10 @@ import { createSourceMiddleware } from "../source-middleware";
 
 export default createSourceMiddleware<
   string[],
-  { wsControllerURL: string; modelUuid: string }
+  { wsControllerURL: string; modelUUID: string }
 >(
   "migration-targets",
-  ({ wsControllerURL: _, modelUuid: modelTag, meta }) => {
+  ({ wsControllerURL: _, modelUUID: modelUUID, meta }) => {
     if (!hasConnection(meta)) {
       throw new Error("connection not provided");
     }
@@ -19,7 +19,10 @@ export default createSourceMiddleware<
 
     return createPollingSource(
       async () => {
-        const response = await listMigrationTargets(connection, modelTag);
+        const response = await listMigrationTargets(
+          connection,
+          `model-${modelUUID}`,
+        );
 
         const controllers = [];
         for (const { uuid } of response.controllers) {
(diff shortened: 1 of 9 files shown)


src/store/middleware/source/migration-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+      async () => {
+        const response = await listMigrationTargets(connection, modelTag);
+
+        const controllers = [];
+        for (const { uuid } of response.controllers) {
+          controllers.push(uuid);
+        }
+        return controllers;
+      },
+      { interval: { seconds: 30 } },
```
huwshimi, line 33: I don't know if we need to poll this, I think it might be better if we fetch/update the data when we display the upgrade form.

src/store/middleware/source/migration-targets.ts
```diff
@@ -0,0 +1,47 @@
+import { createPollingSource } from "data/pollingSource";
+import { listMigrationTargets } from "juju/jimm/api";
+import { actions as jujuActions } from "store/juju";
+
+import { hasConnection } from "../connection/middleware";
+import { createSourceMiddleware } from "../source-middleware";
+
+export default createSourceMiddleware<
```
huwshimi, line 8: I think we need to add this to the array if middleware in `src/store/middleware/source/index.ts` don't we?

src/store/middleware/source/migration-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+  ({ wsControllerURL: _, modelUuid: modelTag, meta }) => {
+    if (!hasConnection(meta)) {
+      throw new Error("connection not provided");
+    }
+
+    const { connection } = meta;
+
+    return createPollingSource(
+      async () => {
+        const response = await listMigrationTargets(connection, modelTag);
```
huwshimi: The action accepts a model UUID, but the API accepts a model tag, so this needs to append the tag prefix to turn it into a tag.

```suggestion
        const response = await listMigrationTargets(connection, `model-${modelUUID}`);
```

src/store/middleware/source/migration-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { hasConnection } from "../connection/middleware";
+import { createSourceMiddleware } from "../source-middleware";
+
+export default createSourceMiddleware<
+  string[],
+  { wsControllerURL: string; modelUuid: string }
+>(
+  "migration-targets",
+  ({ wsControllerURL: _, modelUuid: modelTag, meta }) => {
```
huwshimi: We try to use all uppercase or all lowercase for acronyms.

```suggestion
  ({ wsControllerURL: _, modelUUID: modelTag, meta }) => {
```

src/store/middleware/source/migration-targets.ts
```diff
@@ -0,0 +1,47 @@
...
+          controllers.push(uuid);
+        }
+        return controllers;
+      },
+      { interval: { seconds: 30 } },
+    );
+  },
+  {
+    setData: ({ modelUuid }, data) =>
+      jujuActions.updateModelMigrationTargets({ modelUuid, update: { data } }),
```
huwshimi: These need to have `wsControllerURL` or else be added to the allowed list in the check-auth middleware.

huwshimi (commented):
A few smallish things to get this working.

src/store/middleware/source/migration-targets.ts
andogq: I was thinking it may be nice to poll this whilst the form is open, so that if new targets come online, or one is no longer valid, the list will change. If you'd prefer it to be removed however, I'll convert it to a regular async thunk

huwshimi: Ah I see, I think that could work. There's a risk of the data changing after the user has selected a controller, but the other option would result in an error anyway so this is probably fine.

huwshimi (approved):
Thanks for the updates! Sorry I think I just made nearly every single file in this PR have a conflict.

PR #2281, opened 2026-04-10
feat: add state for a model upgrade

huwshimi:
## Done

- Added redux state, factory and selector for when an upgrade is in progress. It stores both the current version and the version in that it is upgrading to so that the upgrade can be displayed in the model list.

## Details

https://warthogs.atlassian.net/browse/JUJU-9590

### DIFF
--- a/src/store/juju/selectors.ts
+++ b/src/store/juju/selectors.ts
@@ -1380,3 +1380,12 @@ export const getSupportedJujuVersions = createSelector(
       data: null,
     },
 );
+
+/**
+  Get a model upgrade by UUID.
+*/
+export const getModelUpgrade = createSelector(
+  [slice, (_, uuid: null | string = null): null | string => uuid],
+  (state, uuid) =>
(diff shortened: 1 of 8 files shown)


src/store/juju/selectors.ts
```diff
@@ -1380,3 +1380,12 @@ export const getSupportedJujuVersions = createSelector(
       data: null,
     },
 );
+
+/**
+  Get a model upgrade by UUID.
+*/
+export const getModelUpgrade = createSelector(
+  [slice, (_, uuid: null | string = null): null | string => uuid],
```
andogq, line 1388: What's the intention of calling this with `null` is the UUID? I feel like we just shouldn't call this selector if that's the case

andogq (approved):
Small question about one of the selectors

src/store/juju/selectors.ts
huwshimi: It's because of how hooks work in function components, they always have to run, but it many cases required data isn't ready yet, so selectors normally have to accept null/undefined so they can be run without data.

PR #2285, opened 2026-04-10
feat: add access management form

Ninfa-Jeon:
## Done

- Added `AccessManagement` form to add model flow
- Chained this operation to occur only after a successful model creation
- Updated `react-components` to use the latest `MultiSelect`
- A model owner can share their model with other users
- A juju model owner can change their own access level if they are not a superuser on the controller. This doesn't work on JIMM yet due to a bug so the option to downgrade model owner's access stays disabled on JIMM.
- Model owner's access is reduced after granting all permissions to avoid potential bugs.
- Centralised `AccessType` constant

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Adding users from multiselect input works
- Added users are shown in the table
- Access for users other than active user can be changed freely
- Access for model owners can only be changed for non-JIMM scenarios and if they are not superusers on the controller
- Creating model works with or without interacting with this feature

## Details

https://warthogs.atlassian.net/browse/JUJU-9334

## Screenshots

Screenshot 2026-04-29 at 10 23 52 PM

https://github.com/user-attachments/assets/9396b03c-da0e-4cfa-b234-9e24098dc17d


src/store/middleware/model-poller.test.ts
```diff
@@ -1575,6 +1575,82 @@ describe("model poller", () => {
...
+      conn,
+      "new-user@external",
+      "read",
+      undefined,
+      "grant",
+      fakeStore.dispatch,
+    );
+  });
+
+  it("does not apply sharing permissions for owner or active-user entries", async () => {
```
Ninfa-Jeon: this is temporary as I must confirm this with someone on the JIMM/juju team. There is also some confusion regarding an optional flow around it. i will enable this after i have clarity.

src/pages/AddModel/AddModel.tsx
```diff
@@ -167,7 +169,7 @@ const AddModel: FC = () => {
           })}
         />
         <div className="add-model__step" {...testId(TestId.ADD_MODEL_CONTENT)}>
-          <Formik<AddModelFormState>
```
Ninfa-Jeon, line 170: i REALLY would like this back but i broke something and am no longer able to save an object within state if not for this hack

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -140,12 +140,15 @@ const MandatoryDetails = (): JSX.Element => {
...
-            credential: "",
-          }));
+          void setValues(
+            (prevValues) =>
+              ({
+                ...prevValues,
+                cloud: nextCloud,
+                region: "",
+                credential: "",
+              }) as AddModelFormState,
```
Ninfa-Jeon: im guessing this would get reverted once i found a fix for the typing issue i mentioned in `AddModel.tsx`

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+    observer.observe(input, {
+      attributes: true,
+      attributeFilter: ["aria-expanded"],
+    });
+
+    return (): void => {
+      observer.disconnect();
+    };
+  }, []);
```
Ninfa-Jeon: we would be able to get rid of this once my [PR in react-components](https://github.com/canonical/react-components/pull/1367) gets merged.

Ninfa-Jeon: Done

huwshimi (changes requested):
I started going through this and it looks like this could do with going over again if you don't mind and seeing if you can remove some of the redundant and confusing code. It's probably quicker and easier for you to do another pass than me leaving a lot of small comments.

src/pages/AddModel/AccessManagement/_access-management.scss
huwshimi, line 1: - Some of these classes are unscope.
- We generally try to apply a class that we can target rather than the element type.
- You could save a bunch of styles by changing the wrapping padding rather than the elements.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
huwshimi, line 1: There's quite a bit of default handling for things when they're undefined, I think it would be better to not show them than display something like "username".

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
huwshimi: It looks like you've fixed the issues so maybe this isn't required anymore?

src/store/middleware/model-poller.ts
```diff
@@ -540,24 +541,49 @@ function runModelPoller(
...
-                modelUUID: response.uuid,
-                modelURL,
-                wsControllerURL,
-                params: {
-                  type: disabledCommands,
-                },
-              }),
-            );
+            if (disabledCommands !== DisableType.NONE) {
+              const modelURL = wsControllerURL.replace(
```
huwshimi: Like I mentioned on the other PR it would be better to add a util for doing this.

huwshimi (changes requested):
I've left a couple of general comments and I saw the following two issues:

I couldn't get the button to appear when typing in a username (which probably also indicates a missing test) so I wasn't able to add any users to the table.

Clicking on the search and filter magnifying glass gives:
`Warning: You submitted a Formik form using a button with an unspecified `type` attribute.`

Ninfa-Jeon:
> I couldn't get the button to appear when typing in a username (which probably also indicates a missing test) so I wasn't able to add any users to the table.

Strange. The only time this should happen is if the username you are typing is already in the list. Just for my clarification, you are not able to see the button in the dropdown footer?

Ninfa-Jeon:
> Clicking on the search and filter magnifying glass gives:
> `Warning: You submitted a Formik form using a button with an unspecified `type` attribute.`

Seems like an internal issue with `SearchBox` component not specifying the type for this search button. I can raise another PR to fix this if you'd like? There is a hacky fix that can be done for this in the dashboard that prevents default form submission using `onClickCapture` on the parent wrapper. Let me know which one of these sounds reasonable!

src/pages/AddModel/AccessManagement/_access-management.scss
Ninfa-Jeon: the class `.multi-select__footer` is applied on the portal where the dropdown footer appears so while I can't scope it, I am moving it to `_vanilla-overrides.scss` to avoid confusions in the future.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: It still is required unfortunately </3 everywhere we read values from within formik state and try to put them back with some changes, formik refuses to recognise the nested object. due to the dynamic nature of the keys, I am not able to do anything better at this point.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: frankly, a user wouldn't even get this far in the journey if they are not recognised so I'm removing these fallbacks

src/pages/AddModel/AccessManagement/_access-management.scss
Ninfa-Jeon: Others, I have tried to simplify as much as possible. Thank you!

src/store/middleware/model-poller.ts

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -91,28 +91,30 @@ const AccessManagement = (): JSX.Element => {
...
                     <Icon name="plus" />
                     {trimmedSearchInput}
+                    <p className="p-form-help-text email-help-text">
+                      Make sure the email format is correct
+                    </p>
                   </Button>
                 ) : (
-                  <p className="p-form-help-text add-users">
+                  <p className="p-form-help-text u-no-margin--bottom u-sv1">
                     Add users by entering an email address
```
huwshimi: Just tried this with Juju and realised it also says "email address" there as well. I think the different username types are:

- Juju + username/password: `username`
- Juju + Candid: `username@domain`
- JIMM: `username@domain.tld` (email)

src/pages/AddModel/AccessManagement/_access-management.scss
huwshimi: Ah OK, looks like the multi select could do with some spreads to pass classes along. Thanks.

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,36 @@
...
+  width: 100%;
+
+  .email-help-text {
+    margin: 0 0 $spv--small $spv--x-large;
+  }
+
+  .p-icon--plus {
+    margin-right: $spv--small;
+  }
+}
```
huwshimi: I've noted in the component where you can use some classes so you don't need this custom CSS (it also needed a few tweaks to match the design).


```suggestion
```

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+      </Formik>,
+    );
+
+    const input = screen.getByRole("combobox", { name: "Add users" });
+
+    expect(screen.queryByText("No users found")).not.toBeInTheDocument();
+
+    await userEvent.type(input, "not-a-user");
+
+    expect(screen.getByText("No users found")).toBeInTheDocument();
```
huwshimi: I reckon this might be just a little TOO much whitespace :).

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+          <MultiSelect
+            variant="search"
+            placeholder="Search and add users"
+            label="Add users"
+            emptyMessage="No users found"
+            dropdownFooter={
+              <div className="access-management__dropdown-footer">
+                <h4 className="p-muted-heading u-no-padding--top p-text--small u-no-margin--bottom">
+                  YOU CAN ADD
+                </h4>
```
huwshimi:
```suggestion
              <div className="u-flex-grow">
                <h5 className="u-no-padding--top p-text--small-caps u-text--muted u-sh2 u-sv-1--top">
                  You can add
                </h5>
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                {trimmedSearchInput && !hasExactSelectedMatch ? (
+                  <Button
+                    appearance="base"
+                    type="button"
+                    className="u-align--left u-no-padding u-no-margin--bottom"
```
huwshimi:
```suggestion
                    className="u-align--left u-full-width u-no-margin--bottom u-sv-1--top"
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    onClick={handleAddUser}
+                  >
+                    <Icon name="plus" />
```
huwshimi:
```suggestion
                    <Icon name="plus" className="u-sh1--right" />
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    {trimmedSearchInput}
+                    <p className="p-form-help-text email-help-text">
```
huwshimi:
```suggestion
                    <p className="u-text--muted p-text--small u-sh3 u-sv-1">
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                      Make sure the email format is correct
+                    </p>
+                  </Button>
+                ) : (
+                  <p className="p-form-help-text u-no-margin--bottom u-sv1">
```
huwshimi:
```suggestion
                  <p className="u-text--muted p-text--small u-sh2 u-sv-1">
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+            onClose={() => {
+              setSearchInput("");
+            }}
+            onItemsUpdate={handleItemsUpdate}
+          />
+        </div>
+      </div>
+      <table className="u-no-margin--bottom">
+        <thead>
+          <tr className="u-flex u-sv2--top">
```
huwshimi: By making this table use flex it's likely that it'll break with any changes to Vanilla. I've noted some class changes here and in the CSS file for how we can do this without flex.

```suggestion
          <tr className="u-sv2--top">
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+            <th className="u-no-margin--bottom access-management__user-col">
```
huwshimi:
```suggestion
            <th className="u-no-margin--bottom">
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+              Access Level
+            </th>
+            <th className="u-no-margin--bottom u-sh2 u-flex-shrink p-table__cell--icon-placeholder" />
+          </tr>
+        </thead>
+        <tbody>
+          {selectedItems.map(
+            ({ label: userLabel, value: userValue }, index) => (
+              <tr key={userValue} className="u-flex user-row">
+                <td className="access-management__user-col">
```
huwshimi:
```suggestion
                <td>
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    aria-label="Delete"
+                  >
+                    <Icon name="delete" />
+                  </Button>
+                </td>
+              </tr>
+            ),
+          )}
+        </tbody>
+      </table>
```
huwshimi, line 226: I've you use a `<hr />` after the table you won't need the custom CSS.

```suggestion
      </table>
      <hr />
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    onChange={(accessLevel) => {
+                      void setFieldValue(
+                        `shareModelWith["${userValue}"]`,
+                        accessLevel,
+                      );
+                    }}
+                    options={ACCESS_LEVEL_OPTIONS}
+                  />
+                </td>
+                <td className="u-flex-shrink">
```
huwshimi:
```suggestion
                <td className="access-management__delete-col">
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+      <table className="u-no-margin--bottom">
+        <thead>
+          <tr className="u-flex u-sv2--top">
+            <th className="u-no-margin--bottom access-management__user-col">
+              User Name
+            </th>
+            <th className="u-no-margin--bottom access-management__access-col">
+              Access Level
+            </th>
+            <th className="u-no-margin--bottom u-sh2 u-flex-shrink p-table__cell--icon-placeholder" />
```
huwshimi:
```suggestion
            <th className="access-management__delete-col" />
```

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+          </tr>
+        </thead>
+        <tbody>
+          {selectedItems.map(
+            ({ label: userLabel, value: userValue }, index) => (
+              <tr key={userValue} className="u-flex user-row">
```
huwshimi:
```suggestion
              <tr key={userValue}>
```

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  .access-management__access-col {
+    flex: 1 0 0;
+    max-width: 22rem;
+    min-width: 22rem;
+  }
+
+  .user-row:last-child {
+    border-bottom: 1px solid $colors--theme--border-low-contrast;
+  }
```
huwshimi: You can add a `<hr />` after table instead of doing this (I've noted where it should go in the component.).

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,36 @@
...
+.access-management {
+  .multi-select__input {
+    margin-bottom: 0;
+    margin-top: $spv--medium;
+  }
+
+  .access-management__user-col {
+    flex: 1 1 auto;
+    min-width: 0;
+  }
```
huwshimi: To do this without needing flex we can set the width of the delete column instead:


```suggestion
  .access-management__delete-col {
    width: 4rem;
  }
```

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  .access-management__access-col {
+    flex: 1 0 0;
+    max-width: 22rem;
+    min-width: 22rem;
```
huwshimi: This will probably need some adjustment with media queries for smaller screen sizes as it's squashing the name column.

```suggestion
    width: 22rem;
```

src/utils/getModelURL.ts
```diff
@@ -0,0 +1,4 @@
+const getModelURL = (wsControllerURL: string, modelUUID: string): string =>
```
huwshimi, line 1: Thanks!

src/utils/getModelURL.test.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  it("should work with a URL that has path prefix and a port", () => {
+    const result = getModelURL(
+      "wss://controller.example.com/juju:17070/api",
+      "abc-123",
+    );
+    expect(result).toBe(
+      "wss://controller.example.com/juju:17070/model/abc-123/api",
+    );
+  });
```
huwshimi: These are testing things that the util isn't affecting. Is there anything here that you expect might fail or should we remove these tests?

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  it("shows no users found when the filter has no matches", async () => {
+    renderComponent(
+      <Formik initialValues={{ shareModelWith: {} }} onSubmit={vi.fn()}>
+        <AccessManagement />
+      </Formik>,
+    );
+
+    const input = screen.getByRole("combobox", { name: "Add users" });
+
+    expect(screen.queryByText("No users found")).not.toBeInTheDocument();
```
huwshimi: Seeing as we're testing for exact strings it would be good to put them in a labels enum to make the tests more reliable.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  it("shows active user in access table with disabled delete button", async () => {
+    renderComponent(
+      <Formik initialValues={{ shareModelWith: {} }} onSubmit={vi.fn()}>
+        <AccessManagement />
+      </Formik>,
+    );
+
+    expect(screen.getByText(/\(you\)/)).toBeInTheDocument();
```
huwshimi: We have a query: `getCellByHeader` so you could check that this text appears in the right column (it's probably also worth checking the full text so we see that the name actually appears in the cell).

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+const ACCESS_LEVEL_OPTIONS = [
+  { label: "Admin", value: AccessLevel.ADMIN },
+  { label: "Read", value: AccessLevel.READ },
+  { label: "Write", value: AccessLevel.WRITE },
+];
+
+const AccessManagement = (): JSX.Element => {
+  const wsControllerURL = useAppSelector(getWSControllerURL);
+  const userName =
+    useAppSelector((state) => getActiveUserTag(state, wsControllerURL)) ?? "";
```
huwshimi: Our old friend the default empty string is back. It would be more reliable to remove this and conditionally call `getUserName` below.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const activeUserName = getUserName(userName);
+  const { values, setFieldValue } = useFormikContext<AddModelFormState>();
+
+  const shareModelWith = values.shareModelWith ?? {};
+  const activeUserValue = activeUserName || "active-user";
```
huwshimi: It's a bit strange to have this default value. It would be more reliable to only construct the `ACTIVE_USER` item if the user is defined.

This will mean a bit more handling in a few places but will be more true to the types (and not have this backup value just in case).

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const ACTIVE_USER: AccessUserItem = {
+    label: activeUserName,
+    value: activeUserValue,
+    access: shareModelWith[activeUserName] ?? AccessLevel.ADMIN,
```
huwshimi: Doesn't the active user get filtered out of the `shareModelWith` object? If so I think we'd always end up using the nullish condition here.

I know this might change in the future if we add the ability to change the active users's permissions, but I think for now it might be better to simplify this.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+  };
+  const selectedItems: AccessUserItem[] = [
+    ACTIVE_USER,
+    ...Object.entries(shareModelWith)
+      .filter(([user]) => user !== activeUserName)
```
huwshimi: As above, isn't the active user filtered out before setting the shareModelWith value? If so we wouldn't need this filter.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+        access,
+      })),
+  ];
+  const [searchInput, setSearchInput] = useState("");
+  const trimmedSearchInput = searchInput.trim();
+  const hasExactSelectedMatch = selectedItems.some(
+    (item) => item.value === trimmedSearchInput,
+  );
+
+  const handleItemsUpdate = (nextSelectedItems: MultiSelectItem[]): void => {
```
huwshimi, line 62: The fact that the handler's items type is `MultiSelectItem[]` indicates that there's no guarantee that the values returned by the multi select component will contain any of the additional values (e.g. `access`) that are on the items that were given to the component (it currently works, but could be broken by updates to react components).

It would probably be safer to look up the item via the value that was returned and get the access from the original object.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+    const nextShareModelWith: Record<string, string> = {};
+
+    for (const item of nextSelectedItems as AccessUserItem[]) {
+      const { value: key, access } = item;
+
+      // Keep active user in state only when they are not admin.
+      if (key === ACTIVE_USER.value && access === AccessLevel.ADMIN) {
```
huwshimi: It might be safer to exclude this no matter what the access level is set to.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                      </>
+                    ) : (
+                      userLabel
+                    )}
+                  </span>
+                </td>
+                <td className="controller-select__cell access-management__access-col">
+                  <CustomSelect
+                    id={`access-level-${userValue}`}
+                    defaultToggleLabel="Admin"
```
huwshimi: Is this necessary? Shouldn't it default to the selected value?

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    toggleClassName="controller-select__toggle"
+                    dropdownClassName="controller-select__dropdown prevent-panel-close"
+                    value={
+                      userValue === ACTIVE_USER.value
+                        ? ACTIVE_USER.access
+                        : (shareModelWith[userValue] ?? AccessLevel.ADMIN)
```
huwshimi: I think we need some type narrowing at the start of this map as we shouldn't need to set a default of `AccessLevel.Admin`. It seems like that could be dangerous if we've made a mistake in our logic somewhere.

src/scss/_vanilla-overrides.scss
```diff
@@ -34,3 +34,8 @@
     }
   }
 }
+
+// The multi-select component has a border on the footer by default, but in our use case it doesn't fit the design, so we remove it here.
+.multi-select__footer {
+  border: none;
```
huwshimi: I think I'm probably saying the opposite of what I said yesterday, but seeing as we're overriding this anyway, we could adjust the padding here to simplify some other changes:

```suggestion
  border: none;
  padding-left: 0;
  padding-right: 0;
```

src/store/middleware/model-poller.ts
```diff
@@ -540,24 +542,46 @@ function runModelPoller(
...
-              disableCommand.run({
-                modelUUID: response.uuid,
-                modelURL,
-                wsControllerURL,
-                params: {
-                  type: disabledCommands,
-                },
-              }),
-            );
+            if (disabledCommands !== DisableType.NONE) {
```
huwshimi, line 549: It would probably be good to add a test for this condition.

huwshimi (commented):
I found why it wasn't showing the email... I hadn't yarn installed the latest react components. Sorry!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Already left comments for Lukas for clarification on this. Will switch `email` with `username` for non-jimm scenarios for the time being to avoid misleading prompts.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
Ninfa-Jeon: copilot doesnt do ONE thing right does it

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -140,12 +140,15 @@ const MandatoryDetails = (): JSX.Element => {
...
+          );
```
huwshimi: It looks like you can just do:

```suggestion
          const nextCloud = ev.target.value;
          void setFieldValue("cloud", nextCloud);
          void setFieldValue("region", "");
          void setFieldValue("credential", "");
```

src/utils/getModelURL.test.ts
Ninfa-Jeon: the file looked too empty </3 ill remove them XD

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: I'm thinking of just adding the functionality to change permissions for the active user. It's just a tiny bit more logic. That would be better than retracing this.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Sending this item ahead for permission change with my next commit!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Wait I don't get this. The active user can only add new users to the list and by default they show up with Admin access.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Do you think we should default to `Read` instead?

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: I thought you wouldn't like that XD

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Done

src/pages/AddModel/AccessManagement/_access-management.scss
Ninfa-Jeon: Done, thanks! I tried a lot myself but you know these classes better

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: done. thanks!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: thank you!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: done!

src/store/middleware/model-poller.ts
Ninfa-Jeon: done

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
Ninfa-Jeon: Done

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
Ninfa-Jeon: Done

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
Ninfa-Jeon: Done

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: You're right. Done!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: I have made the change but I'm not happy with the amount of conditionals that have now come up. Let me know if you have a better approach for them in mind!

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Defaulting to read as it makes more sense

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: While I have implemented passing the permission changes for active user along, it doesn't seem to be working. My guess is that a model owner can't reduce their access level. Once I confirm this with someone from JIMM/Juju team, I will either make a fix or remove sending active-user to the facade.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/_access-management.scss
Ninfa-Jeon: done
Screenshot 2026-05-01 at 11 17 32 AM

src/pages/AddModel/AccessManagement/AccessManagement.tsx
huwshimi: You set the initial permission when you add the user to the list: https://github.com/canonical/juju-dashboard/pull/2285/changes#diff-b428770a1e2bdc8b8ccd3f849bcaacd6cc39c4576d052988cf123b243ef1a484R89 so it seems like you shouldn't need to also set a default here.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
huwshimi: Same as my other comment, you set the value when you add the user: https://github.com/canonical/juju-dashboard/pull/2285/changes#diff-b428770a1e2bdc8b8ccd3f849bcaacd6cc39c4576d052988cf123b243ef1a484R89 so you shouldn't need to also set a default for here.

Ninfa-Jeon: Done

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: Done

Ninfa-Jeon:
Okay two major updates:
1. [Here](https://github.com/canonical/react-components/pull/1368) is the react-components PR for fixing the warning thrown from `SearchBox` and allowing custom styling of the dropdown.
2. While trying to reduce the model owner's access, I learnt that the operation wouldn't work if the model owner was also a superuser on the controller. This is why I have made it so that the access level dropdown remains disabled for model owners who are controller superusers. This doesn't take into account the manually added users and if they end up being superusers on the controller, it doesn't matter what choice the model owner makes, they will always be admin on it.

huwshimi:
A few QA things:
- When adding a model with Juju the `CreateModel` call contains `credential: ""` which seems to work fine for the admin user, but when using  different user the call returns `error: "no credential specified"`.
- If you add a second admin and then modify your permissions, and then remove the second user from the list your user should revert to admin (right now it's stuck at whatever permission you gave it while the field was enabled).
- Not sure if I've got something wrong with my setup but with both Juju and JIMM I tried adding a second non-controller-admin user, then as that user created a model, giving admin permissions to another user and reducing the logged in users permissions to read but both cases resulted in the logged in user having admin permissions.

Ninfa-Jeon:
> * When adding a model with Juju the `CreateModel` call contains `credential: ""` which seems to work fine for the admin user, but when using  different user the call returns `error: "no credential specified"`.

Seems like we were never sending the credential to the facade although we showed it on UI. Fixed it

Ninfa-Jeon:
> * If you add a second admin and then modify your permissions, and then remove the second user from the list your user should revert to admin (right now it's stuck at whatever permission you gave it while the field was enabled).

Done! Nice catch.

Ninfa-Jeon:
> * Not sure if I've got something wrong with my setup but with both Juju and JIMM I tried adding a second non-controller-admin user, then as that user created a model, giving admin permissions to another user and reducing the logged in users permissions to read but both cases resulted in the logged in user having admin permissions.

~No, you're right. I'm seeing it too. And not just on UI but also on the CLI. I might have to confirm this again with Kian because I'm starting to think reducing access of self might not be possible with the facade. Will change accordingly.~
Alright I have fixed this. For lowering the access, I had to send appropriate props to the function.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -43,7 +43,7 @@ const toCredentialOptions = (
     const credentialName = extractCredentialName(credential);
     return {
       label: credentialName,
-      value: credentialName,
+      value: credential,
```
Ninfa-Jeon, line 46: facade expects full tag value, otherwise it throws an `invalid tag`  error

huwshimi:
With JIMM if I try and share a model with the jimm-test user (the controller admin) it sends:
```
{
    "type": "ModelManager",
    "request": "ModifyModelAccess",
    "version": 9,
    "params": {
        "changes": [
            {
                "access": "write",
                "action": "revoke",
                "model-tag": "model-e8efcd58-070c-4efc-8815-ee1556568a05",
                "user-tag": "user-jimm-test@canonical.com"
            }
        ]
    },
    "request-id": 63
}
```

And the response is:

```
{
    "request-id": 63,
    "response": {
        "results": [
            {
                "error": {
                    "message": "failed to unset model access: failed to remove relations: cannot add or remove relations: POST validation error for Write POST with body {\"code\":\"invalid_write_input\",\"message\":\"Invalid input. Make sure you provide at least one write, or at least one delete\"}\n with error code invalid_write_input error message: Invalid input. Make sure you provide at least one write, or at least one delete",
                    "code": ""
                }
            }
        ]
    }
}
```

Also, if I try and add a user as an admin (test2@example.com), and change the logged in user's access (test@example.com) to read it sends:

```
{
    "type": "ModelManager",
    "request": "ModifyModelAccess",
    "version": 9,
    "params": {
        "changes": [
            {
                "access": "admin",
                "action": "revoke",
                "model-tag": "model-2819d514-9628-44e9-83dd-3f730109207d",
                "user-tag": "user-test2@example.com"
            }
        ]
    },
    "request-id": 33
}
```

And also sends:
```
{
    "type": "ModelManager",
    "request": "ModifyModelAccess",
    "version": 9,
    "params": {
        "changes": [
            {
                "access": "admin",
                "action": "grant",
                "model-tag": "model-2819d514-9628-44e9-83dd-3f730109207d",
                "user-tag": "user-test2@example.com"
            }
        ]
    },
    "request-id": 34
}
```

So it looks like it's both adding and revoking admin permissions.

Ninfa-Jeon:
> So it looks like it's both adding and revoking admin permissions.

my bad! I was too focussed on the active user. it should be fixed now

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,25 @@
+// This is rendered in a portal, so it cannot be scoped
+.access-management__dropdown-footer {
```
huwshimi: If you use `&__dropdown-footer {` inside `.access-management {` it won't get scoped under `.access-management`, it just concatenates the strings, but if you nest the full class name (`.access-management__dropdown-footer`) under `.access-management` it will get scoped.

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.access-management {
+  &__dropdown-input {
+    margin-bottom: 0;
+    margin-top: $spv--medium;
+  }
+
+  &__access-col {
+    width: 22rem;
```
huwshimi, line 7: This was still a bit squashed on medium screens so I've added medium breakpoint.

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,25 @@
...
+  border: none;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.access-management {
+  &__dropdown-input {
+    margin-bottom: 0;
+    margin-top: $spv--medium;
+  }
```
huwshimi: Now that we can apply the classes to the input I've made this use utilities instead of this custom CSS.

src/pages/AddModel/AccessManagement/_access-management.scss
```diff
@@ -0,0 +1,25 @@
+// This is rendered in a portal, so it cannot be scoped
+.access-management__dropdown-footer {
+  border: none;
+  padding-left: 0;
+  padding-right: 0;
```
huwshimi: I've moved these to utilities classes on the element.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      <Formik initialValues={{ shareModelWith: {} }} onSubmit={vi.fn()}>
+        <AccessManagement />
+      </Formik>,
+    );
+
+    const input = screen.getByRole("combobox", {
+      name: Label.MULTI_SELECT_LABEL,
+    });
+    expect(
+      screen.queryByRole("button", { name: "test@example.com" }),
```
huwshimi: I've updated this to use the regex.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    expect(screen.getByRole("button", { name: "Write" })).toBeInTheDocument();
+  });
+
+  it("shows active user in access table with disabled dropdown and remove button", async () => {
+    const state = rootStateFactory.build({
+      general: generalStateFactory.withConfig().build({
+        controllerConnections: {
+          "wss://controller.example.com": {
+            user: {
+              "display-name": "eggman",
```
huwshimi: I've updated this to use the factory.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    );
+    expect(screen.getByText(FormatHint.JIMM)).toBeInTheDocument();
+  });
+
+  it("enables active user dropdown when another user has admin access", async () => {
+    const state = rootStateFactory.build({
+      general: generalStateFactory.withConfig().build({
+        controllerConnections: {
+          "wss://controller.example.com": {
+            user: {
```
huwshimi: I've used the factory here too.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+        onSubmit={vi.fn()}
+      >
+        <AccessManagement />
+      </Formik>,
+      { state },
+    );
+
+    // Active user's dropdown should be enabled since test@example.com is admin
+    const adminButtons = screen.getAllByRole("button", { name: "Admin" });
+    expect(adminButtons[0]).not.toBeDisabled();
```
huwshimi: For accessibility reasons the buttons use the `aria-disabled` property instead of `disabled` so this was passing because it was checking something that would never be set (it was actually failing because isJuju was false).

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    expect(screen.getByText(AddUserHint.JIMM)).toBeInTheDocument();
+
+    await userEvent.type(
+      screen.getByRole("combobox", { name: Label.MULTI_SELECT_LABEL }),
+      "new@example.com",
+    );
+    expect(screen.getByText(FormatHint.JIMM)).toBeInTheDocument();
+  });
+
+  it("enables active user dropdown when another user has admin access", async () => {
```
huwshimi: I've added a second test to check that it's disabled for JIMM.

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+  activeUserControllerAccess: null | string,
+  isJuju?: boolean,
+  activeUserName?: string,
+): boolean => {
+  // Always enable for other users
+  if (!isActiveUser(userValue, activeUserName)) {
+    return false;
+  }
+  // Always disable for controller superusers
+  // Temporarily disable for JIMM as access reduction for active user needs correction
```
huwshimi: I've added a Jira task and a TODO here so we can keep track of this issue.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+      </Formik>,
+      { state },
+    );
+
+    expect(screen.getByRole("button", { name: "Admin" })).toHaveAttribute(
+      "aria-disabled",
+      "true",
+    );
+    const table = await screen.findByRole("table");
+    const userNameCell = customWithin(table).getCellByHeader("User Name");
```
huwshimi: I've updated this to use a label so we don't have to maintain these strings in multiple places.

src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    expect(userNameCell).toHaveTextContent("eggman@external (you)");
+    const deleteButtons = screen.getAllByRole("button", { name: "Delete" });
```
huwshimi: I've used a label here too.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,224 @@
...
+                  ) : (
+                    userLabel
+                  )}
+                </span>
+              </td>
+              <td className="controller-select__cell access-management__access-col">
+                <CustomSelect
+                  id={`access-level-${userValue}`}
+                  toggleClassName="controller-select__toggle"
+                  dropdownClassName="controller-select__dropdown prevent-panel-close"
```
huwshimi: I couldn't see anywhere that `prevent-panel-close` was being handled so I've removed it, but let me know if I've missed something.

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { type AccessUserItem, AddUserHint, FormatHint } from "./types";
+
+/**
+ * Check if a given user value is the active user
+ */
+export const isActiveUser = (
+  userValue: number | string,
+  activeUserName?: string,
+): boolean => userValue === activeUserName;
```
huwshimi: It doesn't seem like there's much benefit having this util and tests just for an `===`.

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+  );
+
+/**
+ * Select hint text based on whether it's Juju or JIMM environment
+ */
+export const selectHint = (
+  isJuju: boolean,
+  jujuHint: string,
+  jimmHint: string,
+): string => (isJuju ? jujuHint : jimmHint);
```
huwshimi: It doesn't seem like this and its tests are necessary when we could just have the ternary inline.

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Remove a user from the share model state
+ */
+export const removeUser = (
+  userValue: number | string,
+  shareModelWith: Record<string, string>,
+  activeUserName: string | undefined,
+): Record<string, string> => {
+  const { [userValue]: _removedUser, ...nextShareModelWith } = shareModelWith;
+
+  // If only one user remains after removal, set them to admin
```
huwshimi: This works if all other users are removed, but if all admins are removed and there are other users (or if the permissions are changed from admin to something else) then the model can still be created without an admin.

I haven't addressed this as I'm not sure what the fix should be.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -0,0 +1,224 @@
...
+                />
+              </td>
+              <td className="access-management__delete-col">
+                <Button
+                  hasIcon
+                  appearance="base"
+                  className="u-no-margin--bottom u-no-padding--top u-no-padding--bottom"
+                  disabled={userValue === activeUserName}
+                  onClick={() => {
+                    void setFieldValue(
```
huwshimi: I've added a test for resetting the admin's permission button from the form's perspective (I know there are already unit tests for the util).

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,119 @@
...
+ */
+export const getUserAccess = (
+  userValue: number | string,
+  activeUserName: string | undefined,
+  activeUserAccess: string | undefined,
+  shareModelWith: Record<string, string>,
+): string =>
+  userValue === activeUserName
+    ? (activeUserAccess ?? AccessLevel.READ)
+    : (shareModelWith[userValue] ?? AccessLevel.READ);
```
huwshimi: I think this can have a single nullish coalescing operator:


```suggestion
  (userValue === activeUserName
    ? activeUserAccess
    : shareModelWith[userValue]) ?? AccessLevel.READ);
```

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -85,6 +84,15 @@ const MandatoryDetails = (): JSX.Element => {
     }
   }, [values.cloud, defaultCloud, setFieldValue]);
 
+  useEffect(() => {
```
huwshimi, line 89: I think this could probably do with a comment and a test.

src/pages/AddModel/AccessManagement/utils.ts
Ninfa-Jeon: Thanks! I'll mention it over the tooltip as well and link the bug I create here too

src/pages/AddModel/AccessManagement/AccessManagement.tsx
Ninfa-Jeon: I don't even remember adding it

src/pages/AddModel/AccessManagement/utils.ts
Ninfa-Jeon: As discussed, I will disable the dropdown for the last remaining admin with a tooltip "please add another admin to modify access" or something

src/utils/getAccessLevel.ts
```diff
@@ -0,0 +1,21 @@
...
+  AccessLevel.WRITE,
+  AccessLevel.ADMIN,
+] as const;
+
+/**
+ * Increase the access level by 1.
+ */
+export const bumpAccessLevel = (accessLevel: AccessLevel): AccessLevel => {
+  const currentIndex = ACCESS_HIERARCHY.indexOf(
+    accessLevel as (typeof ACCESS_HIERARCHY)[number],
```
huwshimi: I think the type annotation is unnecessary here.


```suggestion
    accessLevel
```

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: ah yes

src/store/middleware/model-poller.ts
```diff
@@ -540,24 +543,70 @@ function runModelPoller(
...
+            if (shareModelWith) {
+              const activeUser = getUserName(userTag);
+              const usersToShare = Object.entries(shareModelWith);
+              let activeUserTargetAccessLevel = AccessLevel.ADMIN;
+
+              try {
+                for (const [user, targetAccessLevel] of usersToShare) {
+                  if (user === activeUser) {
+                    activeUserTargetAccessLevel =
+                      targetAccessLevel as AccessLevel;
```
huwshimi: You can remove this type annotation by setting the type elsewhere (I've added some comments).

src/pages/AddModel/types.ts
```diff
@@ -17,6 +17,7 @@ export type AddModelFormState = {
   cloud: string;
   region: string;
   credential: string;
+  shareModelWith?: Record<string, string>;
```
huwshimi:
```suggestion
  shareModelWith?: Record<string, AccessLevel>;
```

src/store/juju/types.ts
```diff
@@ -149,6 +149,7 @@ export type AddModel = {
   credential: string;
   cloudTag: string;
   userTag: string;
+  shareModelWith?: Record<string, string>;
```
huwshimi:
```suggestion
  shareModelWith?: Record<string, AccessLevel>;
```

src/store/middleware/model-poller.ts
Ninfa-Jeon: ~i will try to do as the following comments but we _might_  have to still cast this as we are getting these values out of formik~ I take that back...it works

src/pages/AddModel/AccessManagement/utils.ts
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/utils.ts
Ninfa-Jeon: done

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: done

src/utils/getAccessLevel.ts
Ninfa-Jeon: done

src/pages/AddModel/types.ts
Ninfa-Jeon: done

src/store/juju/types.ts
Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -92,6 +93,28 @@ const AccessManagement = (): JSX.Element => {
     );
   };
 
+  const accessLevelDropdown = (
+    userValue: number | string,
+    accessLevelDisabledReason: string | undefined,
+  ): JSX.Element => (
```
huwshimi: This has been effectively turned into a nested component that's redefined on every render. It needs to be moved to its own file instead.

Ninfa-Jeon: was gonna ask

Ninfa-Jeon: done

src/pages/AddModel/AccessManagement/AccessLeveLDropdown/AccessLevelDropdown.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useFormikContext } from "formik";
+import type { JSX } from "react";
+
+import type { AddModelFormState } from "pages/AddModel/types";
+import { AccessLevel } from "types";
+
+type Props = {
+  value: string;
+  userName: number | string;
+  accessLevelDisabledReason: string | undefined;
```
huwshimi:
```suggestion
  accessLevelDisabledReason?: string;
```

src/pages/AddModel/AccessManagement/AccessLeveLDropdown/AccessLevelDropdown.tsx
huwshimi, line 1: This file could do with a simple test that it can update the form state.

src/pages/AddModel/AccessManagement/utils.ts
```diff
@@ -0,0 +1,154 @@
...
+  );
+};
+
+/**
+ * Remove a user from the share model state
+ */
+export const removeUser = (
+  userValue: number | string,
+  shareModelWith: Record<string, string>,
+  activeUserName: string | undefined,
```
huwshimi:
```suggestion
  activeUserName?: string,
```

huwshimi (approved):
A few small things and then this can land!

PR #2287, opened 2026-04-14
feat: version and upgrade target support

huwshimi:
## Done

Added selectors, utils and factories for working with supported versions and upgrade targets.

## Details

Part of https://warthogs.atlassian.net/browse/JUJU-9589.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -127,13 +127,15 @@ export default function Breadcrumb(): JSX.Element {
         >
           <strong>{modelName}</strong>
         </Link>
-        <ModelActions
-          modelUUID={modelUUID}
-          modelName={modelName}
-          redirectOnDestroy
-          position="left"
-          qualifier={qualifier}
-        />
+        {modelUUID ? (
+          <ModelActions
+            modelUUID={modelUUID}
+            modelName={modelName}
+            redirectOnDestroy
+            position="left"
+            qualifier={qualifier}
+          />
+        ) : null}
       </li>
     );
   };
--- a/src/store/juju/selectors.ts
+++ b/src/store/juju/selectors.ts
@@ -1451,13 +1451,29 @@ export const getModelMigrationControllersByVersion = createSelector(
  * Get the version objects filtered by versions that have corresponding
  * model migration targets.
  */
-export const getModelMigrationTargetVersions = createSelector(
-  [getControllerData, getSupportedJujuVersions, getModelMigrationTargets],
-  (controllers, supportedVersions, migrationTargets) => {
+export const getModelUpgradeVersions = createSelector(
+  [
+    getControllerData,
+    getSupportedJujuVersions,
+    getModelMigrationTargets,
+    getModelDataByUUID,
+    (state: RootState, modelUUID?: null | string): Controller | null =>
+      getModelControllerDataByUUID(
+        state,
+        getModelDataByUUID(state, modelUUID)?.info?.["controller-uuid"],
+      ),
+  ],
+  (
+    controllers,
+    supportedVersions,
+    migrationTargets,
+    model,
+    modelController,
+  ) => {
     if (!controllers) {
       return [];
     }
-    return migrationTargets.data?.reduce<VersionElem[]>(
+    const upgradeVersions = (migrationTargets.data ?? []).reduce<VersionElem[]>(
       (versions, controllerUUID) => {
         const controller = getControllerByUUIDUtil(controllers, controllerUUID);
         const controllerVersion = controller
@@ -1482,6 +1498,24 @@ export const getModelMigrationTargetVersions = createSelector(
       },
       [],
     );
+    const controllerVersion = modelController
+      ? getControllerVersion(modelController)
+      : null;
+    const modelVersion = model?.model.version;
+    if (
+      modelVersion &&
+      controllerVersion &&
+      modelVersion !== controllerVersion &&
+      !upgradeVersions.find((version) => version.version === controllerVersion)
+    ) {
+      const controllerUpgradeVersion = supportedVersions.data?.find(
+        (version) => version.version === controllerVersion,
+      );
+      if (controllerUpgradeVersion) {
+        upgradeVersions.push(controllerUpgradeVersion);
+      }
+    }
+    return upgradeVersions;
   },
 );
 
@@ -1489,7 +1523,7 @@ export const getModelMigrationTargetVersions = createSelector(
  * Get available Juju versions for the latest LTS and the latest stable version, if there are controllers available.
  */
 export const getRecommendedVersions = createSelector(
-  [getModelMigrationTargetVersions],
+  [getModelUpgradeVersions],
   (migrationTargetVersions) => {
     let ltsVersion: null | VersionElem = null;
     let stableVersion: null | VersionElem = null;
--- a/src/store/juju/selectors.ts
+++ b/src/store/juju/selectors.ts
@@ -12,7 +12,7 @@ import cloneDeep from "clone-deep";
 import fastDeepEqual from "fast-deep-equal/es6";
 
 import type { AuditEvent } from "juju/jimm/JIMMV3";
-import type { RelationshipTuple } from "juju/jimm/JIMMV4";
+import type { RelationshipTuple, VersionElem } from "juju/jimm/JIMMV4";
 import type { FullStatusAnnotations, ModelInfo } from "juju/types";
 import {
   getActiveUserTag,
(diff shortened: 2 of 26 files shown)


src/store/juju/selectors.ts
```diff
@@ -562,7 +570,7 @@ export const getModelUUIDFromList = createSelector(
     ): null | string => qualifier,
   ],
   (modelList: ModelsList, modelName, qualifier) => {
-    let modelUUID: string = "";
```
huwshimi, line 566: This was causing bugs because the model UUID might not exist, but setting it to an empty string meant that we were missing some truthy checks (I've added them in this PR).

src/store/juju/slice.ts
```diff
@@ -152,7 +152,11 @@ const slice = createSlice({
       loading: false,
     },
     selectedApplications: {},
-    supportedJujuVersions: {},
+    supportedJujuVersions: {
```
huwshimi, line 155: I made this a single state rather than being keyed by controller as these versions are not controller specific.

src/store/middleware/source/jimm-supported-versions.ts
```diff
@@ -1,30 +1,13 @@
...
-  date: string;
-  /**
-   * URL to release information.
-   */
-  link: string;
-};
-
 export default createSourceMiddleware<
-  SupportedVersion[],
+  VersionElem[],
```
huwshimi, line 10: I changed this to just use the upstream type so that we don't need to convert between the two types and makes it easier to work with throughout the code.

src/store/juju/slice.ts
```diff
@@ -761,24 +765,16 @@ const slice = createSlice({
...
+      state.supportedJujuVersions.loading =
+        payload.update.loading ?? state.supportedJujuVersions.loading;
       if (payload.update.data !== undefined) {
-        value.data = payload.update.data;
+        state.supportedJujuVersions.data = payload.update.data;
       }
       if (payload.update.error !== undefined) {
-        value.error = payload.update.error;
+        state.supportedJujuVersions.error = payload.update.error;
       }
```
andogq, line 778: When I did this previously, I would get errors from redux about incorrectly mutating state, which is why I went with the other approach. Are these no longer happening?

src/store/juju/utils/upgrades.ts
```diff
@@ -0,0 +1,39 @@
+import { getMajorMinorVersion } from "utils";
+import getPatchVersion from "utils/getPatchVersion";
+
+const LTS_RELEASES = [3.6];
```
andogq, line 8: Might be worth leaving a comment describing where this information can be found

src/store/juju/utils/upgrades.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Check whether a Juju release is an LTS.
+ */
+export const isLTS = (version: string): boolean => {
+  const majorMinor = getMajorMinorVersion(version);
+  return majorMinor ? LTS_RELEASES.includes(majorMinor) : false;
+};
+
+/**
+ * Check whether an upgrade will require a controller migration. A model needs
```
andogq, line 19: Same here, if this is documented anywhere

src/store/juju/utils/upgrades.ts
```diff
@@ -0,0 +1,39 @@
...
+  upgradeVersion: string,
+): boolean => {
+  const upgradeMajorMinor = getMajorMinorVersion(upgradeVersion);
+  const controllerMajorMinor = getMajorMinorVersion(controllerVersion);
+  const upgradePatch = getPatchVersion(upgradeVersion);
+  const controllerPatch = getPatchVersion(controllerVersion);
+  if (
+    !upgradeMajorMinor ||
+    !controllerMajorMinor ||
+    !upgradePatch ||
```
andogq: These should be explicit null checks, as these functions may return 0

src/store/juju/utils/upgrades.ts
```diff
@@ -0,0 +1,39 @@
...
+export const isLTS = (version: string): boolean => {
+  const majorMinor = getMajorMinorVersion(version);
+  return majorMinor ? LTS_RELEASES.includes(majorMinor) : false;
+};
+
+/**
+ * Check whether an upgrade will require a controller migration. A model needs
+ * to migrate to a new controller if:
+ * - the major or minor version has changed, or
+ * - the patch version is higher than its current controller's patch version.
```
andogq, line 22: Maybe I'm misunderstanding, but it doesn't seem that this clause is checked here. If it's done elsewhere, it may be good to say where

src/store/juju/utils/upgrades.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  it("is an LTS if it is 3.6.x", () => {
+    expect(isLTS("3.6.36")).toBe(true);
+  });
+
+  it("is not an LTS if it is anything else", () => {
+    expect(isLTS("3.7.36")).toBe(false);
+  });
+});
+
+describe("requiresMigration", () => {
```
andogq, line 13: Add some cases where the version contains 0 (and maybe also prerelease versions?)
Also, `it.for` may make these cases less repetitive

src/components/Breadcrumb/Breadcrumb.tsx
andogq, line 1: I'm a little confused by this. Should there be added tests to ensure that `ModelActions` isn't shown if there is no model?

src/store/juju/utils/controllers.ts
```diff
@@ -20,3 +20,17 @@ export const getControllerByUUID = (
...
+
+/**
+    Get the Juju version for a controller.
+  */
+export const getControllerVersion = (controller: Controller): null | string => {
+  let version: null | string = null;
+  if ("agent-version" in controller && controller["agent-version"]) {
+    version = controller["agent-version"];
+  }
+  if ("version" in controller && controller.version) {
```
andogq, line 31: Would this make more sense as an `else if`, otherwise the previous branch will just be over written and is some what useless (I know this is just copied verbatim from above).

src/store/juju/selectors.ts
```diff
@@ -1416,3 +1418,130 @@ export const getModelMigrationTargets = createSelector(
...
+          : null;
+        const controllerVersion = controller
+          ? getControllerVersion(controller)
+          : null;
+        if (controller && controllerVersion && controllerVersion === version) {
+          controllerList.push(controller);
+        }
+        return controllerList;
+      },
+      [],
```
andogq, line 1445: This might be slightly clearer as a map and filter, if you're feeling like it

andogq (approved):
Looks good overall, some minor suggestions. What's happening with the changes to `modelUUID` everywhere?

huwshimi:
> Looks good overall, some minor suggestions. What's happening with the changes to `modelUUID` everywhere?

For some reason [the model UUID was being set to an empty string](https://github.com/canonical/juju-dashboard/pull/2287#discussion_r3077573149) which meant that we weren't getting hints from typescript that the value might be falsy and this was causing a few bugs, including one that I introduced in the followup to this branch that I pulled these changes from (dispatching the action to fetch the upgrade targets with a blank model uuid).

src/store/juju/slice.ts
huwshimi: I'm not getting those errors, and I'm not sure what might have caused them, as [`createSlice` uses Immer](https://redux-toolkit.js.org/usage/immer-reducers) so that you can modify state and not be concerned about mutations.

src/store/juju/utils/upgrades.ts
huwshimi: Done.

src/store/juju/utils/upgrades.ts
huwshimi: Done.

src/store/juju/utils/upgrades.ts
huwshimi: Thanks! Done.

src/store/juju/utils/upgrades.ts
huwshimi: It's checked by the last condition: `upgradePatch > controllerPatch`.

src/store/juju/utils/upgrades.test.ts
huwshimi: Done.

src/components/Breadcrumb/Breadcrumb.tsx
huwshimi: I personally think this is OK without a test as we [never show the model details](https://github.com/canonical/juju-dashboard/blob/e1553354d11777c80fdf8a11e40aa1e99f1c6b94/src/pages/EntityDetails/EntityDetails.tsx#L69) if it doesn't exist in the store, but if you think it would be better to add a test for completeness I can do that.

src/store/juju/utils/controllers.ts
huwshimi: Thanks, I've just made these return instead of assigning to a variable.

src/store/juju/slice.ts
andogq: Yep, it was coming from within Immer itself. If it works for you though, then go for it!

src/components/Breadcrumb/Breadcrumb.tsx
andogq: Happy for you to ignore this, I didn't consider your modelUUID comment

PR #2289, opened 2026-04-15
chore: move userCredentials and cloudInfo to Source

Ninfa-Jeon:
## Done

- Moved `userCredentials` and `cloudInfo` middleware from `model-poller` to Source
- Updated tests
- Added tests for sources:
  - `cloud-info`
  - `user-credentials`
- Moves `hasConnections` in `connection/middleware.ts` to `connection/utils.ts` to prevent cyclical imports

## QA

- Pull code
- Open http://0.0.0.0:8036/
- All functionality to fetch clouds and credentials along with the current add-model flow remains unchanged
- Also make sure that the clouds and credentials state remains isolated between different user sessions. I noticed some inherited state while dev and had to make additional changes.

## Details

https://warthogs.atlassian.net/browse/JUJU-9492


Ninfa-Jeon:
I am only a tiny bit unsure of what to do with the `interval` and when to trigger the `stop` on the middleware object.

huwshimi:
@andogq @Ninfa-Jeon should these calls use the new [`createProcess`](https://github.com/canonical/juju-dashboard/pull/2295/changes#diff-c5ced22e7878455de2f27ec7bb630e59b8b64b8aebb7c246e6517bc7dc59d513R43) middleware? Or should we poll while the forms are open?

Ninfa-Jeon:
> @andogq @Ninfa-Jeon should these calls use the new [`createProcess`](https://github.com/canonical/juju-dashboard/pull/2295/changes#diff-c5ced22e7878455de2f27ec7bb630e59b8b64b8aebb7c246e6517bc7dc59d513R43) middleware? Or should we poll while the forms are open?

To be frank, I originally implemented these methods to be non-polling so I am okay with either option.

andogq:
As per our chat in stand-up, I think it's good to keep these within a source, and use a large interval, and start/stop the source only when it's required.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -88,9 +88,8 @@ const MandatoryDetails = (): JSX.Element => {
       !userCredentials.credentials[initialCloud]
     ) {
       dispatch(
-        jujuActions.fetchUserCredentials({
+        userCredentialsMiddleware.actions.start({
```
huwshimi, line 95: In the cleanup method for this useEffect you can call `userCredentialsMiddleware.actions.stop(...)` to stop polling when this component gets unmounted.

As noted in model-poller.ts below, you can also call `cloudInfoMiddleware.actions.start` and `cloudInfoMiddleware.actions.stop` in this useEffect.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -118,9 +117,8 @@ const MandatoryDetails = (): JSX.Element => {
       nextCloud.length > 0
     ) {
       dispatch(
-        jujuActions.fetchUserCredentials({
+        userCredentialsMiddleware.actions.start({
```
huwshimi: You will also need to call `userCredentialsMiddleware.actions.stop()` with the old cloudTag (either that or you could set the cloudTag in state in this component that would case the useEffect above to update each time the cloudTag changes).

src/store/middleware/source/user-credentials.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        const response = (await connection.facades.cloud?.userCredentials({
+          "user-clouds": [
+            {
+              "cloud-tag": cloudTag,
+              "user-tag": connection.info.user.identity,
+            },
+          ],
+        })) ?? { results: [{ result: [] }] };
+        return response.results[0].result ?? [];
```
huwshimi: You can just use optional chaining instead of mimicking the response.

```suggestion
        }));
        return response?.results[0]?.result ?? [];
```

src/store/middleware/model-poller.ts
```diff
@@ -85,7 +86,11 @@ function runModelPoller(
           reduxStore.dispatch(generalActions.updateLoginLoading(false));
         }
 
-        reduxStore.dispatch(jujuActions.fetchClouds({ wsControllerURL }));
+        reduxStore.dispatch(
+          cloudInfoMiddleware.actions.start({
```
huwshimi, line 90: I've made a comment in MandatoryDetails.tsx, but instead of calling this here I think you can just do it when we need this data for the add model form, rather than starting polling when we make the connection.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: The latter would also trigger fetching clouds info unless I put more conditions in place. I'd prefer the former.

huwshimi: You could also split the cloud info call into its own useEffect.

Ninfa-Jeon: Looks like I'm gonna have to put the cloudTag in state either way. Putting the cloud middleware in its own useEffect.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: Done.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: Done

src/store/middleware/source/user-credentials.ts
Ninfa-Jeon: Done

src/store/middleware/model-poller.ts
Ninfa-Jeon: As discussed, I have kept this unchanged and added a call to `stop` inside the logout thunk. I have updated the QA steps above to add a special case that I noticed when I was working on this. Thanks!

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
huwshimi: It looks like this can be removed now.

Ninfa-Jeon: Done

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -80,53 +86,35 @@ const MandatoryDetails = (): JSX.Element => {
...
-    wsControllerURL,
-    activeUser,
-    defaultCloud,
-    userCredentials.credentials,
-    values.cloud,
-  ]);
+    return (): void => {
+      if (wsControllerURL && selectedCloud) {
+        dispatch(
+          userCredentialsMiddleware.actions.stop({
```
huwshimi, line 104: We should have a test for this.

src/store/app/thunks.ts
```diff
@@ -26,10 +27,18 @@ export const logOut = createAsyncThunk<
...
   const state = thunkAPI.getState();
   const pingerIntervalIds = getPingerIntervalIds(state);
+  const wsControllerURL = getWSControllerURL(state);
   bakery.storage.clear();
   Object.entries(pingerIntervalIds ?? {}).forEach((pingerIntervalId) => {
     clearInterval(pingerIntervalId[1]);
   });
+  if (wsControllerURL) {
+    thunkAPI.dispatch(
+      cloudInfoMiddleware.actions.stop({
```
huwshimi, line 37: This should also have a test.

src/store/middleware/source/cloud-info.ts
```diff
@@ -0,0 +1,47 @@
+import type { CloudsResult } from "@canonical/jujulib/dist/api/facades/cloud/CloudV7";
+
+import { createPollingSource } from "data/pollingSource";
+import { actions as jujuActions } from "store/juju";
+
+import { hasConnections } from "../connection/middleware";
+import { createSourceMiddleware } from "../source-middleware";
+
+export default createSourceMiddleware<
```
huwshimi, line 9: This file should have some tests.

src/store/middleware/source/user-credentials.ts
huwshimi, line 1: This file also needs some tests.

src/store/middleware/model-poller.ts
huwshimi, line 90: We have should have a test that this starts.

huwshimi (commented):
This is looking good, just need some tests.

src/store/middleware/source/cloud-info.ts
Ninfa-Jeon: We don't have any tests written for files in this directory. Let me try to write some

Ninfa-Jeon: it is harder than i figured. maybe we can add them later for these middleware files?

src/store/middleware/source/user-credentials.ts
Ninfa-Jeon: same as the cloud-info above

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
Ninfa-Jeon: done

src/store/app/thunks.ts
Ninfa-Jeon: done

src/store/middleware/model-poller.ts
Ninfa-Jeon: done

src/store/middleware/source/cloud-info.ts
huwshimi: @andogq not sure how we've missed these, but maybe you two can work together to figure out how to add tests to the sources.

andogq: https://github.com/canonical/juju-dashboard/pull/2301

I've started adding tests here, to give you an idea of how it should go. We can discuss it in stand-up if needed.

Ninfa-Jeon: i can add tests the same way if we're moving the `hasConnection` to utils from middleware for the sake

Ninfa-Jeon: done!

src/store/middleware/source/user-credentials.ts
Ninfa-Jeon: done

src/store/app/thunks.test.ts
```diff
@@ -83,6 +84,25 @@ describe("thunks", () => {
...
+    const dispatch = vi.fn();
+    const getState = vi.fn(() => state);
+    const logoutSpy = vi.spyOn(CandidAuth.prototype, "logout");
+    new CandidAuth(dispatch);
+    await action(dispatch, getState, null);
+    expect(dispatch).toHaveBeenCalledWith(
+      cloudInfoMiddleware.actions.stop({
+        wsControllerURL: "wss://controller.example.com",
+      }),
+    );
```
huwshimi, line 97: It looks like this is the only thing that's different between this and the above test. We can probably remove everything else from this test, so it's explicit about what it's testing.

src/store/middleware/source/user-credentials.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  beforeEach(() => {
+    mockUserCredentials = vi.fn();
+    connection = {
+      info: {
+        user: {
+          "display-name": "eggman",
+          identity: "user-eggman@external",
+          "controller-access": "",
+          "model-access": "",
+        },
```
huwshimi: I know we still have a bunch of places that include the full object, but you can just do `authUserInfoFactory.build()`.

huwshimi (approved):
Looks good! Two tiny comments.

PR #2294, opened 2026-04-16
feat(connection): use connectionList to specify multiple connections

andogq:
Previously, `meta.connection` would be set to the connection associated
with `payload.wsControllerURL` if `meta.withConnection` was `true`. This
limited all actions to at most a single connection.

Now, `meta.connectionList` can be provided, containing a list of keys
present within `payload`. Connections will be added to `meta.connections`,
an object keyed by each key in `meta.connectionList`. If not provided,
it will default to `["wsControllerURL"]`.

Related utils, such as `hasConnection` have been updated to accept
a similar list of connection keys, in order to verify that they have
been added.

Before:

```
{
    type: "some/action",
    payload: {
        controller1: "wss://controller1.example.com/"
        controller2: "wss://controller2.example.com/"
    },
    metadata: {
        withConnection: true,
        connectionList: ["controller1", "controller2"],
    }
}
```

After:

```
{
    type: "some/action",
    payload: {
        controller1: "wss://controller1.example.com/"
        controller2: "wss://controller2.example.com/"
    },
    metadata: {
        withConnection: true,
        connectionList: ["controller1", "controller2"],
        connections: {
            controller1: Connection,
            controller2: Connection,
        }
    }
}
```

### DIFF
--- a/src/store/middleware/connection/middleware.ts
+++ b/src/store/middleware/connection/middleware.ts
@@ -38,24 +38,25 @@ function hasConnectionURL<P extends Record<string, unknown>>(
 /**
  * Guard to ensure that the given object has a `connections` key with provided connection keys.
  */
-export function hasConnections<K extends readonly string[]>(
-  object: Record<string, unknown>,
-  connectionKeys: K,
-): object is { connections: Record<K[number], ConnectionWithFacades> } & Record<
+export function hasConnections<K extends string>(
+  meta: Record<string, unknown>,
+  connectionKeys: K[],
+): meta is { connections: Record<K, ConnectionWithFacades> } & Record<
   string,
   unknown
 > {
-  return (
-    "connections" in object &&
-    typeof object.connections === "object" &&
-    object.connections !== null &&
-    connectionKeys.every(
-      (key) =>
-        // @ts-expect-error - Doesn't pick up check due to closure
-        key in object.connections &&
-        // @ts-expect-error - Doesn't pick up check due to closure
-        object.connections[key] instanceof Connection,
+  if (
+    !(
+      "connections" in meta &&
+      typeof meta.connections === "object" &&
+      meta.connections !== null
     )
+  ) {
+    return false;
+  }
+  const connections = meta.connections as Record<string, unknown>;
+  return connectionKeys.every(
+    (key) => key in connections && connections[key] instanceof Connection,
   );
 }
 
--- a/src/store/middleware/connection/middleware.ts
+++ b/src/store/middleware/connection/middleware.ts
@@ -16,7 +16,7 @@ import {
 } from "store/general/selectors";
 import type { AuthCredential } from "store/general/types";
 import type { RootState, Store } from "store/store";
-import { isPayloadAction } from "types";
+import { isMetaAction, isPayloadAction } from "types";
 import analytics from "utils/analytics";
 import { logger } from "utils/logger";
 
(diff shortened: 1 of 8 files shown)


src/store/middleware/connection/middleware.ts
```diff
@@ -26,21 +26,37 @@ export const MISSING_WS_CONTROLLER_URL_ERROR =
...
+): action is PayloadAction<{ [key]: string } & P> {
+  return key in action.payload && typeof action.payload[key] === "string";
 }
 
-export function hasConnection(
+/**
+ * Guard to ensure that the given object has a `connections` key with provided connection keys.
+ */
+export function hasConnections<K extends readonly string[]>(
   object: Record<string, unknown>,
```
huwshimi, line 42: I didn't pick this up previously, but this should probably use a variable name that isn't a built-in type.

src/store/middleware/connection/middleware.ts
```diff
@@ -26,21 +26,37 @@ export const MISSING_WS_CONTROLLER_URL_ERROR =
...
+  unknown
+> {
+  return (
+    "connections" in object &&
+    typeof object.connections === "object" &&
+    object.connections !== null &&
+    connectionKeys.every(
+      (key) =>
+        // @ts-expect-error - Doesn't pick up check due to closure
+        key in object.connections &&
```
huwshimi, line 55: If you assign `object.connections` to a variable before this return statement then typescript will be able to narrow the type in each condition.

src/store/middleware/connection/middleware.ts
```diff
@@ -26,21 +26,37 @@ export const MISSING_WS_CONTROLLER_URL_ERROR =
...
+        // @ts-expect-error - Doesn't pick up check due to closure
```
huwshimi, line 56: It looks like this is also masking other error TS errors.

src/store/middleware/connection/middleware.ts
```diff
@@ -26,21 +26,37 @@ export const MISSING_WS_CONTROLLER_URL_ERROR =
...
+  key: string,
+): action is PayloadAction<{ [key]: string } & P> {
+  return key in action.payload && typeof action.payload[key] === "string";
 }
 
-export function hasConnection(
+/**
+ * Guard to ensure that the given object has a `connections` key with provided connection keys.
+ */
+export function hasConnections<K extends readonly string[]>(
```
huwshimi, line 41: It probably makes a bit more sense to make K a single string rather than having to get the type of an element in the return type e.g. 

```suggestion
export function hasConnections<K extends string>(
  object: Record<string, unknown>,
  connectionKeys: K[],
): object is { connections: Record<K, ConnectionWithFacades> } & Record<
```

src/store/middleware/connection/middleware.ts
```diff
@@ -145,29 +161,48 @@ export function createConnectionMiddleware(): {
...
+        > = {};
+
+        for (const connectionKey of connectionList) {
+          if (typeof connectionKey !== "string") {
+            continue;
+          }
+
+          if (hasConnectionURL(action, connectionKey)) {
+            // Action has the `withConnection` property, so attach the connection to it.
+            const connection = await connections.get(
```
huwshimi, line 193: Not sure if you want to do it in this PR, but this `get` method in the connection manager will need to be updated to know how to connect to a model.

huwshimi (commented):
A few very minor comments. I do have two questions but I'm not sure if they need to be solved in this PR or in your createProcess PR or a future one, but I'll leave them on this PR:
1. Because there can be a lot of models we don't leave model connections open (we used to while you're viewing a model for the model watcher, but not other calls like getting the full status, configuring an app etc.) so somehow we'll need to know if there are no calls in progress that need the model connection and then close it.
2. In #2295 the `upgradeModel` call can use the model connection that would be created when the action is dispatched, however for upgrading a model that requires a migration I think we will lose connection to the model while it switches to the new controller (though the ws address will remain the same). This will need testing to confirm, but we might need a way for in-progress middleware to reestablish connections if they're lost.

src/store/middleware/connection/middleware.ts
andogq: This is a nice change, thakns

andogq:
These are two very good points.

I'm thinking of two possible options for managing connection closing:

1. The [`using` keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using) would be ideal for this, as we could track as references are cleaned up. Unfortunately, [Safari does not support it](https://caniuse.com/mdn-javascript_statements_using), so it probably isn't the go for now.

2. We could build an assumption into the connection manager that if a promise is returned from `next(action)`, the connection counter should be decremented once it resolves (and clean up the connection when it reaches zero). This would require some additional changes and questions:
  
  -  Polling middleware should return a promise from the `start` action, which only resolves when the polling is completed (ie `stop` action is dispatched).
 
  - An escape-hatch is required for use in the model-poller middleware, with un-migrated actions (this should be very manageable, given the connection manager is shared, rather than the connection middleware).

  - We would need to decide how to handle actions which don't return a promise, I'd imagine we can immediately mark the usage of that connection as done, and log a warning.

We can undertake this in a [future ticket](https://warthogs.atlassian.net/browse/JUJU-9690).

---

As for switching the connection during the upgrade/migration process, I will investigate this as part of implementing that item.

src/store/middleware/connection/middleware.ts
```diff
@@ -38,24 +38,25 @@ function hasConnectionURL<P extends Record<string, unknown>>(
 /**
  * Guard to ensure that the given object has a `connections` key with provided connection keys.
  */
-export function hasConnections<K extends readonly string[]>(
-  object: Record<string, unknown>,
-  connectionKeys: K,
-): object is { connections: Record<K[number], ConnectionWithFacades> } & Record<
+export function hasConnections<K extends string>(
+  maybeObject: Record<string, unknown>,
```
huwshimi: This might be a bit pedantic, but don't we know that it's an object, but don't know whether it contains connections? Could even just call it `meta` as that's what gets passed to this function so it would be more descriptive about what this function accepts.

huwshimi (approved):
Looks good, with the other things being done in followups as discussed.

PR #2296, opened 2026-04-17
chore(ci): fix Juju workflows

huwshimi:
## Done

Last run before label change: https://github.com/canonical/juju-dashboard/actions/runs/26380890550?pr=2296.

- Fixed the contextual menu item roles to match the changes in react-components.
- Fixed Candid snap issue.
- Made it so that it cancels in-progress jobs when pushing changes.
- Fixed issues in e2e related to shifting content.
- Swapped to a faster charm.
- Fixed an issue where if there was an error when checking the login status it would display the "could not continue" error instead of the specific whoami error.

NOTE:
- the JIMM workflows are being addressed in this PR: https://github.com/canonical/juju-dashboard/pull/2236.
- Juju 4.0 with microk8s is currently failing due to a secrets issue that I will address separately if it's not a bug in the latest release.

### DIFF
--- a/.github/actions/setup-candid/action.yaml
+++ b/.github/actions/setup-candid/action.yaml
@@ -25,6 +25,13 @@ runs:
       uses: ./.github/actions/wait-for-snap
       with:
         snap: candid
+    - name: Wait for Candid config
+      shell: bash
+      run: |
+        until [ -f /var/snap/candid/current/config.yaml ]
+        do
+            sleep 1
+        done
     - name: Configure candid
       shell: bash
       run: |
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -46,6 +46,11 @@ on:
 
 permissions: read-all
 
+# Ensure only one e2e job runs at a time.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   juju-machine-local:
     name: Test Juju, machine-charm and local-auth
@@ -92,7 +97,7 @@ jobs:
     uses: ./.github/workflows/e2e-juju-machine-charm-local-auth.yml
     secrets: inherit
     with:
-      juju-channel: "4.0/stable"
+      juju-channel: "4.0/candidate"
       send-notification: false
     if: |
       github.event_name != 'pull_request' || (
@@ -103,7 +108,7 @@ jobs:
     uses: ./.github/workflows/e2e-juju-k8s-charm-local-auth.yml
     secrets: inherit
     with:
-      juju-channel: "4.0/stable"
+      juju-channel: "4.0/candidate"
       send-notification: false
     if: |
       github.event_name != 'pull_request' || (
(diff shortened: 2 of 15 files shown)


.github/actions/setup-candid/action.yaml
```diff
@@ -25,6 +25,13 @@ runs:
       uses: ./.github/actions/wait-for-snap
       with:
         snap: candid
+    - name: Wait for Candid config
```
huwshimi, line 28: Fixes an issue where the config wasn't ready yet.

e2e/helpers/objects/application.ts
```diff
@@ -9,26 +9,30 @@ import type { Model } from "../objects";
  */
 export enum CharmName {
   // spell
-  localhost = "anbox-cloud-dashboard",
-  microk8s = "pgbouncer-k8s",
+  localhost = "any-charm",
```
huwshimi, line 12: This charm is a lot more lightweight and can be used for both LXD and k8s.

src/auth/OIDCAuth.ts
```diff
@@ -44,14 +41,7 @@ export class OIDCAuth extends pollingMixin(Auth) {
...
     } catch (error) {
-      this.dispatch(
-        generalActions.storeLoginError({
-          wsControllerURL,
-          error: OIDCAuthLabel.WHOAMI,
-        }),
-      );
-      // Halt the connection attempt.
-      return false;
+      throw new Error(OIDCAuthLabel.WHOAMI);
```
huwshimi, line 44: The login error was being replaced by the parent's error handler when this returned false (the "could not continue" error). In this particular case we want to display the WHOAMI error instead of the generic error.

src/store/middleware/model-poller.ts
```diff
@@ -41,7 +41,6 @@ import { ModelsError } from "./types";
 export enum LoginError {
   LOG = "Unable to log into controller.",
   NO_INFO = "Unable to retrieve controller details.",
-  WHOAMI = "Unable to check authentication status. You can attempt to log in anyway.",
```
huwshimi, line 44: This error was moved elsewhere so this was no longer used.

e2e/fixtures/setup.ts
```diff
@@ -70,11 +70,13 @@ export const test = base.extend<Fixtures>({
     { option: true },
   ],
   jujuCLI: async ({ testOptions, browser }, use) => {
+    const versionOutput = await exec("juju --version");
```
huwshimi, line 73: There are some behaviour changes between Juju 3 and 4 so this now passes the Juju version to the CLI class.

.github/workflows/end-to-end.yml
```diff
@@ -92,7 +97,7 @@ jobs:
     uses: ./.github/workflows/e2e-juju-machine-charm-local-auth.yml
     secrets: inherit
     with:
-      juju-channel: "4.0/stable"
+      juju-channel: "4.0/candidate"
```
Ninfa-Jeon: Maybe there should be a comment explaining why we are using candidate instead of stable?

Ninfa-Jeon (approved):
Looks good!

.github/workflows/end-to-end.yml
huwshimi: Done.

PR #2297, opened 2026-04-17
feat: add a util for creating toast notifications

huwshimi:
## Done

- Added a util for creating toast notifications without the need for JSX.
- Updated toast notifications to use the util.

## QA

- Open the redux dev tools and in the dispatch box enter the following and a toast should appear:
```
{
  type: "app/createToast",
  payload: {
    message: "Boo!",
    severity: "caution"
  }
}
```

## Details

https://warthogs.atlassian.net/browse/JUJU-9674


src/components/LogIn/LogIn.tsx
```diff
@@ -75,7 +74,7 @@ export default function LogIn(): JSX.Element {
   useEffect(() => {
     visitURLs?.forEach((visitURL) => {
       if (!viewedAuthRequests.current.includes(visitURL)) {
-        reactHotToast.custom((toast: ToastInstance) => (
+        reactHotToast.custom((toast) => (
```
Ninfa-Jeon, line 77: why are we not using the `toastNotification` util here?

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -0,0 +1,22 @@
...
+const ToastNotification: FC<Props> = ({ toastInstance }: Props) => {
+  return (
+    <ToastCard
+      toastInstance={toastInstance}
+      type={toastInstance.type as NotificationSeverity}
+    >
+      <>{toastInstance.message}</>
+    </ToastCard>
+  );
+};
```
Ninfa-Jeon: Maybe I'm misunderstanding something but this just seems like a named wrapper around `ToastCard`. Is this really needed?

src/components/LogIn/LogIn.tsx
huwshimi: I should have mentioned this, but for more complex cases where extra props need to be set on `ToastCard` or the notification needs access to the toast props then we still need to use a custom toast.

In this case `AuthenticationButton` uses the toast `id` prop to do `reactHotToast.remove(toast.id);` to close the toast when clicked.

src/components/ToastNotification/ToastNotification.tsx
huwshimi: Nope you're right, it was mostly just a way to map the toast instance props to the card props and be consistent with the base layout and test setups. Because this needs to be the same for both cases I just realised it could also contain the `<Toaster>` as well, so I've moved in the Toaster bits and renamed this component and used it in both places.

src/store/middleware/notifications.ts
```diff
@@ -0,0 +1,40 @@
+/*
+  Redux middleware that gates every request on authentication unless an action
+  has been allowed.
+*/
+
```
andogq:
```suggestion
```

src/store/middleware/notifications.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+  Redux middleware to enable gating actions on the respective controller
+  authentication.
+  @param action The typical Redux action or thunk to execute
+  @param options Any options that this checker needs to perform an
+    appropriate auth check.
+      wsControllerURL: The full controller websocket url that the controller
+        is stored under in redux in order to determine it's logged in status.
+*/
```
andogq, line 9: this also looks out of date

src/store/middleware/notifications.ts
```diff
@@ -0,0 +1,40 @@
...
+  void,
+  RootState,
+  Store["dispatch"]
+> = () => (next) => (action) => {
+  if (!isAction(action)) {
+    return next(action);
+  }
+  if (
+    isSpecificAction<ReturnType<typeof createToast>>(action, createToast.type)
+  ) {
```
andogq: Should be able to use `createToast.matches(action)`

src/store/middleware/notifications.ts
```diff
@@ -0,0 +1,40 @@
...
+        is stored under in redux in order to determine it's logged in status.
+*/
+export const notificationsMiddleware: Middleware<
+  void,
+  RootState,
+  Store["dispatch"]
+> = () => (next) => (action) => {
+  if (!isAction(action)) {
+    return next(action);
+  }
```
andogq: If `createToast.matches` is used, this check won't be required.

src/store/middleware/notifications.test.ts
andogq, line 1: Seems to be a lint error in this file

src/pages/AddModel/AddModel.tsx
```diff
@@ -101,21 +100,21 @@ const AddModel: FC = () => {
     if (addModelState.loaded && !addModelState.loading && wsControllerURL) {
       if (addModelState.success) {
         // Handle a successful creation
-        reactHotToast.custom((toast: ToastInstance) => (
-          <ToastCard type="positive" toastInstance={toast}>
-            <b>Model "{modelName}" added successfully</b>
-          </ToastCard>
-        ));
+        toastNotification(
+          <b>Model "{modelName}" added successfully</b>,
```
andogq, line 104: Not now, but I feel like some day we should make these notifications a bit more consistent (with respect to bold, quotes, etc)

src/layout/BaseLayout/BaseLayout.tsx
```diff
@@ -75,15 +75,7 @@ const BaseLayout: FC = () => {
...
-      <Toaster
-        position="bottom-right"
-        containerClassName="toast-container"
-        toastOptions={{
-          duration: 5000,
-        }}
-        reverseOrder={true}
-      />
-      {import.meta.env.DEV && <DevBar />}
+      <ToastNotifications />){import.meta.env.DEV && <DevBar />}
```
andogq: This is formatted somewhat strangely, but I'm goint t oassume that it's fine

andogq (approved):
Just a few out-dated comments to be fixed, but is good overallq

src/layout/BaseLayout/BaseLayout.tsx
huwshimi: Oops, I don't know how I did that, I've fixed it.

src/store/middleware/notifications.test.ts
huwshimi: Fixed.

src/store/middleware/notifications.ts
huwshimi: Done

src/store/middleware/notifications.ts
huwshimi: Argh, I remember thinking I need to come back to fix up these comments, but I guess I forgot. Thanks.

src/store/middleware/notifications.ts
huwshimi: Nice, I didn't know about that util! Done.

src/store/middleware/notifications.ts
huwshimi: Done.

andogq (approved):

PR #2298, opened 2026-04-17
feat: page setup for configs and constraints

Ninfa-Jeon:
## Done

- Implemented configurations for configs&constraints tab
- Added link to docs
- Added toggle to view changed configs only
- Added radios to switch between list vs YAML view
- Created a stacklist builder for configs (can be extended to constraints)
- Made the configs list scrollable allowing the information under configs section to be always visible on various resolutions (check attachment below for reference)
- Implemented data porting from list to YAML view
- Implemented view preservation between tab navigations (view attachment)

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Navigate to the Configurations&Constraints tab on AddModel page
- You should see the configs table (constraints will be implemented next cycle as per UX decision and disabled commands will be implemented as a part of https://warthogs.atlassian.net/browse/JUJU-9681)
- The docs link should lead to correct documentation
- You should be able to switch between List and YAML view
- Changing anything in list view should flag it with an in-progress icon
- Switching the toggle `Changed configs only` on should show these flagged fields only. Switching it on with no fields changed should show nothing.
- Changing anything in list view and switching to YAML view should port this data to YAML with appropriate comments
- Making changes to this text field is possible
- Making changes to YAML view and switching back to List view DOES NOT preserve data (needs UX consideration so it will be dealt with in the future)
- All populated data is preserved between navigations, i.e., making changes to list view and navigating away does not wipe the changes and populating YAML field and navigating away not only preserves the YAML but also displays YAML view the next time user gets to this tab (refer attachment).

## Details

https://warthogs.atlassian.net/browse/JUJU-9679

## Screenshots

https://github.com/user-attachments/assets/bf4b74df-e445-420d-b5f6-542c842d0fbf

The following shows what this tab looks like with constraints below it, there is scrollview and resizing:

https://github.com/user-attachments/assets/e7319cd4-7def-42e7-94db-5c212b010928

https://github.com/user-attachments/assets/f05f0e57-acc1-41c7-8ace-7cd08da29762


src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import userEvent from "@testing-library/user-event";
+import { vi } from "vitest";
+
+import { renderComponent } from "testing/utils";
+
+import ContentSwitcher from "./ContentSwitcher";
+
+const onModeChange = vi.fn();
+
+const defaultProps = {
```
huwshimi: It would be better to assign this inside the `beforeEach` as there's nothing stopping this object being mutated.

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  it("renders properly", () => {
+    renderComponent(<ContentSwitcher {...defaultProps} />);
+
+    expect(
+      screen.getByRole("heading", { name: "Configuration (optional)" }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole("link", { name: "Learn about configuration" }),
+    ).toHaveAttribute("href", "https://juju.is/docs");
+    expect(screen.getByLabelText("List")).toBeInTheDocument();
```
huwshimi: These tests will be much easier to maintain if we use an enum of the strings so we don't need to update the tests if the labels change.

It's also probably a good idea to use getByRole for these so we don't get false positives from other elements, or so these don't pass if the actual input gets removed.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+      });
+    });
+
+  const handleConfigModeChange = (isListMode: boolean): void => {
+    const existingConfigYAML =
+      typeof values.configYAML === "string" ? values.configYAML.trim() : "";
+
+    if (!isListMode && !existingConfigYAML) {
+      void setFieldValue(
+        "configYAML",
```
huwshimi: If you use an enum of the field names and use that here and also pass it to the generic for useFormikContext then these will get validated so if the fields changes we don't accidentally forget to update all the places the names are usued.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { testId } from "testing/utils";
+import { externalURLs } from "urls";
+
+import ContentSwitcher from "./ContentSwitcher/ContentSwitcher";
+import { type CategoryDefinition, CONFIG_CATEGORIES } from "./configCatalog";
+import { Label, TestId } from "./types";
+import { buildConfigYAML, getChangedFields, isConfigChanged } from "./utils";
+
+const ConfigsConstraints = (): JSX.Element => {
+  const { values, setFieldValue } = useFormikContext<Record<string, unknown>>();
```
huwshimi: If you supply a generic to useFormContext that has a mapping of field names to field types then typescript will be able to follow that type for values and setFieldValue so it will validate we're using those correctly. You'd also need to allow for the generated config fields but they'll always be strings so you can do something like `<FormFields & Record<string, string>>`.

If you search for `FormFields` you'll see how that gets done elsewhere.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const [isConfigListMode, setIsConfigListMode] = useState(
+    values.configInputMode !== "yaml",
```
huwshimi: As this only works with an exact string it should use an enum.

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -0,0 +1,60 @@
+.configs__section-heading {
+  line-height: 1rem;
+  margin-bottom: 0.75rem;
```
huwshimi: This is getting overridden by the `u-no-margin--bottom` on the element.

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -0,0 +1,60 @@
+.configs__section-heading {
+  line-height: 1rem;
```
huwshimi: It looks like you could achieve the same thing by using `u-no-padding--top` and will have the benefit of not squashing the lines if they wrap on small screens.

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -0,0 +1,60 @@
...
+  margin-bottom: 0.75rem;
+}
+
+.configs__link {
+  font-size: #{map-get($font-sizes, small)}rem;
+}
```
huwshimi: It looks like the design is using small text so you can use the `p-text--small` class on the element instead this.

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -0,0 +1,60 @@
...
+
+    .p-form-validation__input {
+      margin-bottom: 0;
+    }
+  }
+}
+
+.configs__yaml-input {
+  min-height: 10rem;
+  width: 40%;
```
huwshimi: I suspect this should be using a value based on the cols sizes rather than a percent. You might be able to set it to 100% and wrap it in a col (you might need to check with Lukas what size he intended it to be as I can't see it in the figma).

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -0,0 +1,60 @@
...
+    div {
+      display: flex;
+      flex-direction: column;
+      gap: 5px;
+    }
+  }
+
+  .configs__input {
+    padding-bottom: 3px;
+    padding-top: 3px;
```
huwshimi: These should probably use spacing variables (I think `$sp-x-small`).

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+              checked={changedOnly}
+              onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                setChangedOnly(event.target.checked);
+              }}
+            />
+            <div className="configs__table-scroll">
+              <MainTable
+                className="p-main-table configs__table"
+                rows={buildRows(CONFIG_CATEGORIES)}
+              />
```
huwshimi: I'm not sure if makes semantic sense for these to be in a table and it will be difficult to make this work at small screen sizes. It looks more like a modified version of the stacked form pattern: https://vanillaframework.io/docs/base/forms#stacked (with help text under the label and borders between rows).

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+              className: "configs__input",
+            },
+          ],
+        };
+      });
+    });
+
+  const handleConfigModeChange = (isListMode: boolean): void => {
+    const existingConfigYAML =
+      typeof values.configYAML === "string" ? values.configYAML.trim() : "";
```
huwshimi: If you pass the type of the fields to `useFormikContext` (noted above) you won't need to check the type here.

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <h5 className="configs__section-heading u-no-margin--bottom">{title}</h5>
+      <a href={docsLink} className="configs__link">
+        {docsLabel}
+      </a>
+      <div className="u-flex u-flex--gap">
+        <div>
+          <RadioInput
+            checked={showPrimary}
+            label="List"
+            name={name}
```
huwshimi: A very small optimisation, but unless you're expecting to need to access these fields outside of this component you could just use `useId` to generate a unique name.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+                    className: "configs__category",
+                  },
+                ]
+              : []),
+            {
+              content: (
+                <div>
+                  <span>
+                    {changed ? (
+                      <i className="p-icon--status-in-progress-small" />
```
huwshimi:
```suggestion
                      <Icon name="status-in-progress-small" />
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,136 @@
...
+import ContentSwitcher from "./ContentSwitcher/ContentSwitcher";
+import { type CategoryDefinition, CONFIG_CATEGORIES } from "./configCatalog";
+import { Label, TestId } from "./types";
+import { buildConfigYAML, getChangedFields, isConfigChanged } from "./utils";
+
+const ConfigsConstraints = (): JSX.Element => {
+  const { values, setFieldValue } = useFormikContext<Record<string, unknown>>();
+  const [isConfigListMode, setIsConfigListMode] = useState(
+    values.configInputMode !== "yaml",
+  );
```
huwshimi: Is there any need to use state for this? Could you just do `const isConfigListMode = values.configInputMode === "list";`? (Also reads a bit better if it uses `=== "list"` as the name mentions list mode).

src/pages/AddModel/AddModel.tsx
```diff
@@ -25,6 +25,7 @@ import { testId } from "testing/utils";
 import urls from "urls";
 import { toErrorString } from "utils";
 
+import ConfigsConstraints from "./ConfigsConstraints/ConfigsConstraints";
```
huwshimi: For consistency if you add `./ConfigsConstraints/index.ts` with `export { default } from "./ConfigsConstraints";` then you can do `import ConfigsConstraints from "./ConfigsConstraints";`

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
```diff
@@ -0,0 +1,308 @@
...
+          ],
+        },
+      },
+      {
+        label: "net-bond-reconfigure-delay",
+        description:
+          "The amount of time in seconds to sleep between ifdown and ifup when bridging",
+        input: {
+          type: "number",
+          defaultValue: 17,
```
huwshimi: HTML input fields are always strings (and sometimes Formik can get a bit confused when working with numbers) so it's probably easier to make this a string and then convert to a number before you submit it to the API.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -0,0 +1,48 @@
+import type { CategoryDefinition } from "./configCatalog";
+
+export const isConfigChanged = (
+  label: string,
+  defaultValue: unknown,
+  values: Record<string, unknown>,
```
huwshimi: With my suggestion about making the `net-bond-reconfigure-delay` default a string then you'll always be working with strings, so you can simplify the logic in this util.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+    : category.fields;
+
+export const buildConfigYAML = (
+  categories: CategoryDefinition[],
+  values: Record<string, unknown>,
+): string => {
+  const yamlSections = categories
+    .map((category) => {
+      const changedFields = getChangedFields(category, true, values).map(
+        (field) => `${field.label}: ${String(values[field.label])}`,
```
huwshimi: When this gets inserted into the template literal it'll get converted into a string, so the `String()` is redundant (also with my suggestion to always make everything strings you wouldn't need this either way).

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      );
+
+      if (changedFields.length === 0) {
+        return null;
+      }
+
+      return [`# ${category.category}`, ...changedFields].join("\n");
+    })
+    .filter((section): section is string => Boolean(section));
```
huwshimi: `Boolean` can just be passed as the function (or you could use a reduce here instead of map then filter).

```suggestion
    .filter(Boolean);
```

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: that margin isnt needed...stupid copilot

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: After trying this out, it still seems to allow a lot of space between the heading and the link below it. Maybe I can set the `line-height` for large screens only?

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
Ninfa-Jeon: How would you suggest we validate the input in such a case? As this data is large and the form validation sits within `AddModel.tsx`, I'm finding it hard to picture that part.

Ninfa-Jeon: Frankly, using strings would be easier on the logic because sending these values as a YAML sounds easier than mapping them individually and so, we wouldnt have to convert numbers to strings.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: I tried it your way but wouldnt we need this state to switch between List and YAML mode?

Ninfa-Jeon: Nvm the previous comment. It needed a couple of changes to ensure everything worked but i have removed the state now.

src/pages/AddModel/ConfigsConstraints/utils.ts

src/pages/AddModel/ConfigsConstraints/utils.ts

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: I've made the change but the validation is weak currently. Maybe Ill address is in another PR

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
Ninfa-Jeon: Done

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: Done for both implemented tabs

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
Ninfa-Jeon: Done.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: I have asked Lukas, will make the changes when I hear from him

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: Removed

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Done.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.test.tsx
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.test.tsx
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -0,0 +1,51 @@
...
+      <div className="u-flex u-flex--gap">
+        <div>
+          <RadioInput
+            checked={showPrimary}
+            label={InputMode.LIST}
+            name={id}
+            onChange={() => {
+              onModeChange(true);
+            }}
+            value="list"
```
huwshimi: This should use the enum.

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <div>
+          <RadioInput
+            checked={!showPrimary}
+            label={InputMode.YAML}
+            name={id}
+            onChange={() => {
+              onModeChange(false);
+            }}
+            value="yaml"
```
huwshimi: This should use the enum as well.

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
```diff
@@ -0,0 +1,51 @@
...
+  onModeChange,
+  title,
+}: Props): JSX.Element => {
+  const id = useId();
+  return (
+    <>
+      <h5 className="configs__section-heading u-no-margin--bottom">{title}</h5>
+      <a href={docsLink} className="p-text--small">
+        {docsLabel}
+      </a>
```
huwshimi: It looks like this gives the correct spacing, without the need for custom CSS:

```suggestion
      <h5 className="u-no-padding--top u-no-margin--bottom">{title}</h5>
      <p className="u-no-margin--bottom p-text--small">
        <a href={docsLink}>
          {docsLabel}
        </a>
      </p>
```

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
huwshimi: I've added a comment in the ContentSwitcher component with a suggested fix.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: You're 100% right about this. My bad. Will push with corrected code in a bit.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
Ninfa-Jeon: confirmed and set cols in place instead, thanks!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Done! Thank you

src/pages/AddModel/ConfigsConstraints/ContentSwitcher/ContentSwitcher.tsx
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/StackList/StackList.tsx
```diff
@@ -0,0 +1,62 @@
...
+          values,
+          config.defaultValue,
+        );
+        return (
+          <>
+            <FormikField
+              {...(config.input?.type === "select"
+                ? { component: Select }
+                : { type: "text", placeholder: config.placeholder })}
+              key={config.label}
```
huwshimi: My mistake, I put the key in the wrong place so there's a console error:

```suggestion
          <Fragment key={config.label}>
            <FormikField
              {...(config.input?.type === "select"
                ? { component: Select }
                : { type: "text", placeholder: config.placeholder })}
```

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
```diff
@@ -0,0 +1,256 @@
...
+            { label: "False", value: "false" },
+          ],
+        },
+      },
+      {
+        label: "net-bond-reconfigure-delay",
+        description:
+          "The amount of time in seconds to sleep between ifdown and ifup when bridging",
+        defaultValue: "17",
+        placeholder: "17",
```
huwshimi: It looks like wherever you set `placeholder` you also have a `defaultValue` of the same value, does it need both?

src/pages/AddModel/ConfigsConstraints/StackList/StackList.tsx
```diff
@@ -0,0 +1,62 @@
...
+              label={
+                <>
+                  {changed ? (
+                    <Icon name="status-in-progress-small u-sh-3 u-sh1--right" />
+                  ) : null}
+                  {config.label}
+                </>
+              }
+              name={config.label}
+              defaultValue={config.defaultValue}
```
huwshimi, line 39: When changing these text fields it's giving an `A component is changing an uncontrolled input to be controlled` error.

I think the solution is to loop through the fields and add `[config.label]: config.defaultValue ?? ""` to the `initialValues` object in AddModel.tsx.

huwshimi (approved):
A couple of tiny things and then this can land!

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
Ninfa-Jeon: Ah, you're right. No it doesn't need both. I'll correct it right away, thanks

PR #2300, opened 2026-04-22
feat(process): create process middleware

andogq:
Create process middleware to handle the lifecycle of one-shot functions. Includes a `createProcess` function to handle the plumbing.

### DIFF
--- a/src/store/middleware/process/createProcess.ts
+++ b/src/store/middleware/process/createProcess.ts
@@ -1,0 +1,91 @@
+import type { PayloadActionCreator } from "@reduxjs/toolkit";
+import { createAction } from "@reduxjs/toolkit";
+
+import type { Store } from "store/store";
+import { toErrorString } from "utils";
+
+import type { Hooks, Process, ProcessActions } from "./types";
+
+export function createProcess<Payload, Status, Result>(
+  identifier: string,
+  runProcess: (
+    payload: { meta: Record<string, unknown> } & Payload,
+  ) => AsyncIterator<Status, Result, void>,
+  processActions: ProcessActions<Payload, Status, Result>,
+  hooks?: Hooks<Payload>,
+): Process<Payload> {
+  const actions = {
+    run: createAction(
+      `process/${identifier}/run`,
+      (
+        payload: Payload,
+      ): { payload: Payload; meta?: Record<string, unknown> } =>
+        Object.assign(
+          {
+            payload,
+          },
+          hooks?.addActionMeta
+            ? { meta: hooks.addActionMeta(payload) }
+            : undefined,
+        ),
+    ) as unknown as PayloadActionCreator<Payload>,
+  };
+
+  const start = async (
+    payload: { meta: Record<string, unknown> } & Payload,
+    dispatch: Store["dispatch"],
+  ): Promise<void> => {
+    // Mark as running.
+    dispatch(processActions.setRunning(payload, true));
+
+    // Begin running.
+    const process = runProcess(payload);
+
+    /**
+     * For the given iterator result, dispatch into the store as a status update or result, using
+     * the provided action.
+     *
+     * @returns `true` if this is the last result from the iterator.
+     */
+    function processIteratorResult(
+      result: IteratorResult<Status, Result>,
+    ): boolean {
+      if (result.done) {
+        // Final item from this iterator, set the outcome.
+        dispatch(processActions.setOutcome(payload, { result: result.value }));
+        return true;
+      } else {
+        // Status update.
+        dispatch(processActions.setStatus(payload, result.value));
+        return false;
+      }
+    }
+
+    // Continually poll from process iterator.
+    while (true) {
+      try {
+        const result = await process.next();
+        if (processIteratorResult(result)) {
+          // Process completed.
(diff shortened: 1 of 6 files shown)


src/store/middleware/process/types.ts
```diff
@@ -0,0 +1,30 @@
...
+  setRunning: (payload: Payload, running: boolean) => PayloadAction<unknown>;
+  setOutcome: (
+    payload: Payload,
+    outcome: ProcessOutcome<Result>,
+  ) => PayloadAction<unknown>;
+};
+
+export type Process<Payload> = {
+  actions: {
+    run: PayloadActionCreator<Payload>;
```
huwshimi: This type is what is not allowing the createAction with a prepared payload as it's defaulting the prepare action to `void`.

```suggestion
    run: PayloadActionCreator<Payload, string, PrepareAction<Payload>>;
```

src/store/middleware/process/createProcess.ts
```diff
@@ -0,0 +1,91 @@
...
+      ): { payload: Payload; meta?: Record<string, unknown> } =>
+        Object.assign(
+          {
+            payload,
+          },
+          hooks?.addActionMeta
+            ? { meta: hooks.addActionMeta(payload) }
+            : undefined,
+        ),
+    ) as unknown as PayloadActionCreator<Payload>,
```
huwshimi: Having to use `as unknown as something` is an indication we've got our types wrong somewhere and will allow `createAction` to return anything and still be valid.

I've noted in types.ts what the correct type is and then you can remove this `as`.

src/store/middleware/process/createProcess.ts
```diff
@@ -0,0 +1,91 @@
...
+  ) => AsyncIterator<Status, Result, void>,
+  processActions: ProcessActions<Payload, Status, Result>,
+  hooks?: Hooks<Payload>,
+): Process<Payload> {
+  const actions = {
+    run: createAction(
+      `process/${identifier}/run`,
+      (
+        payload: Payload,
+      ): { payload: Payload; meta?: Record<string, unknown> } =>
```
huwshimi: It looks like it can infer this return type, but there's probably no harm in having it.

src/store/middleware/process/createProcess.ts
```diff
@@ -0,0 +1,91 @@
...
+        Object.assign(
+          {
+            payload,
+          },
+          hooks?.addActionMeta
+            ? { meta: hooks.addActionMeta(payload) }
+            : undefined,
+        ),
```
huwshimi: Is there any harm in having `meta: undefined` in the object, or does redux not like that? If it allows it we could just have:
 
```suggestion
          {
            payload,
            meta: hooks?.addActionMeta?.(payload),
          }
```

src/store/middleware/process/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      const process = processes.find(({ actions }) =>
+        actions.run.match(action),
+      );
+      if (!process) {
+        return next(action);
+      }
+
+      if (process.actions.run.match(action)) {
```
huwshimi, line 25: Don't we already know that the action matches from the `.find` above?

src/store/middleware/process/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+        const actionPayload = action.payload as Payload;
+        const actionMeta = (
+          action as PayloadAction<Payload, string, Record<string, unknown>>
+        ).meta as Record<string, unknown> | undefined;
```
huwshimi: It might be nicer to use the `isPayloadAction` and `isMetaAction` utils here rather than these type annotations.

src/store/middleware/process/middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+        const payload = Object.assign(
+          {},
+          actionPayload,
+          actionMeta ? { meta: actionMeta } : undefined,
+        );
```
huwshimi: Same question here about whether it's OK to have `payload: undefined` or `meta: undefined` in the object.

huwshimi (commented):
Some small comments, mostly around types.

src/store/middleware/process/types.ts
andogq: That's perfect, thank you very much!

src/store/middleware/process/middleware.ts
andogq: Yes, however I was keeping in mind having additional actions in the future. For example, I originally had a `cancel` action too, although I've ripped it out to keep things simple for now. If you want this check reverted too, I'm happy to do that.

src/store/middleware/process/createProcess.ts
andogq: It's nicer without it. I was getting an eslint warning previously that there was no signature, but seems fine now.

src/store/middleware/process/middleware.ts
huwshimi: Ah OK, I see, that's fine.

huwshimi (approved):
Looks good!

PR #2301, opened 2026-04-22
feat(source): add tests

andogq:
- Added tests for sources:
  - `jimm-supported-versions`
  - `migration-targets`
  - `model-list`
- Move some exports to prevent cyclical imports
  - `ModelsError` in `model-poller.ts` to `types.ts`
  - `hasConnections` in `connection/middleware.ts` to `connection/utils.ts`

### DIFF
--- a/src/store/middleware/source/model-list.ts
+++ b/src/store/middleware/source/model-list.ts
@@ -9,6 +9,8 @@ import { createSourceMiddleware } from "../source-middleware";
 import { ModelsError } from "../types";
 
 export const NOT_AUTHENTICATED_ERROR = "not authenticated with controller";
+export const NO_MODEL_MANAGER_FACADE =
+  "ModelManager facade is not available on the connection";
 
 export async function getModelList(
   connection: ConnectionWithFacades,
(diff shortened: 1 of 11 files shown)


src/store/middleware/source/model-list.ts
```diff
@@ -1,44 +1,49 @@
...
+
+export const NOT_AUTHENTICATED_ERROR = "not authenticated with controller";
+
+export async function getModelList(
+  connection: ConnectionWithFacades,
+): Promise<UserModelList> {
+  if (!connection.info.user?.identity) {
+    throw new Error(NOT_AUTHENTICATED_ERROR);
+  }
+
```
Ninfa-Jeon, line 24: I am also adding another check right under this one for the presence of the facade like so:
```suggestion
  if (!connection.facades.modelManager) {
    throw new Error("Unsupported facade: modelManager");
  }
```

src/store/middleware/source/model-list.ts
```diff
@@ -1,44 +1,54 @@
...
 import { logger } from "utils/logger";
 
 import { hasConnections } from "../connection/util";
-import { ModelsError } from "../model-poller";
 import { createSourceMiddleware } from "../source-middleware";
+import { ModelsError } from "../types";
+
+export const NOT_AUTHENTICATED_ERROR = "not authenticated with controller";
+export const NO_MODEL_MANAGER_FACADE =
+  "ModelManager facade is not available on the connection";
```
Ninfa-Jeon, line 13: these strings will be used in almost every middleware we make, do you think we should extract them to a global const?...

Ninfa-Jeon (approved):
All good, one question. Doesn't need to be addressed if its no use

src/store/middleware/source/model-list.ts
andogq: I think you're right, but I'll leave it here for now. Sometime we should go through everywhere and have consistent error messages

PR #2302, opened 2026-04-22
feat: move target controller to the version table

huwshimi:
## Done

- Add instructions for creating model migration controllers.
- Make the upgrade model panels wider.
- Update the target controller select to use a CustomSelect.
- Move the target controller select into the table of versions.

## QA

- If you don't already have controllers with different versions then follow the new instructions in the HACKING doc to set them up.
- Open the model upgrade panel and choose a version that is higher than the model's current controller (if using the hacking doc instructions this will be 3.6.21).
- On the second screen you should see the target controller select in the table of versions.

## Details

https://warthogs.atlassian.net/browse/JUJU-9697
https://warthogs.atlassian.net/browse/JUJU-9347

## Screenshots

Screenshot 2026-04-22 at 12 44 00 pm

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -38,7 +38,7 @@ contribute and what kinds of contributions are welcome.
       - [Vanilla Framework](#vanilla-framework)
       - [Vanilla React Components](#vanilla-react-components)
   - [Deployed JIMM controller](#deployed-jimm-controller)
-    - [Adding users](#adding-users)
+      - [Adding users](#adding-users)
     - [Self signed certificates](#self-signed-certificates)
     - [Juju on M1 Macs](#juju-on-m1-macs)
   - [Building the Docker image](#building-the-docker-image)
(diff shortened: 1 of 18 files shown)


HACKING.md
```diff
@@ -38,14 +38,16 @@ contribute and what kinds of contributions are welcome.
       - [Vanilla Framework](#vanilla-framework)
       - [Vanilla React Components](#vanilla-react-components)
   - [Deployed JIMM controller](#deployed-jimm-controller)
-    - [Adding users](#adding-users)
+      - [Adding users](#adding-users)
```
andogq: Is this accidental?

andogq (approved):

HACKING.md
huwshimi: Looks like the plugin I use picked up that the heading was another level lower, which it shouldn't have been. I've fixed it.

PR #2303, opened 2026-04-22
feat: yaml to list view data porting and confirmation dialog

Ninfa-Jeon:
Depends on https://github.com/canonical/juju-dashboard/pull/2384 https://github.com/canonical/juju-dashboard/pull/2385

## Done

- Added `YAMLErrorsModal` in `ConfigsConstraints` to show different types of errors that show up during YAML validation and parsing.
- Added a util to validate and parse YAML text input when the triggers a switch to list mode. The util uses `parseDocument` function from `yaml` library
- This util also makes note of various errors that show up during parsing and segregates them based on "Invalid key", "Invalid value" and "Other" errors.
- The modal uses these three error types and displays them accordingly, along with the line number
- The util is also equipped to drop any key-value pairs that were added in list mode but wiped in yaml mode. Those keys are reverted to default making them unchanged.
- Forcing a switch without fixing erroneous values drops them. Any values that were parsed without errors are set in the formik context.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Go to ConfigsConstraints tab on AddModel page and play a bit with the List<->YAML mode
  - Set a value in list mode, switch to YAML, reset it in YAML, switch to list.
  - Set incorrect values and value types
  - Use incorrect format for setting "key- value" instead of "key: value"
  - Try setting an unknown key-value pair
  - Set a field expecting numerical value to a string
- All the above cases should be handled gracefully. [Figma](https://www.figma.com/design/Rs1v3IS0oY2RK4CU4yo4Ly/25.10-JAAS-model-management---Juju?node-id=2068-148122&m=dev)

## Details

https://warthogs.atlassian.net/browse/JUJU-9833

## Screenshots
Screenshot 2026-06-23 at 7 34 55 PM

### DIFF
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
@@ -28,6 +28,8 @@ describe("CategoriesListing", () => {
       searchPlaceholder: "Search configurations",
       yamlPlaceholder: Label.MODEL_CONFIG_PLACEHOLDER,
       searchName: "configSearch",
+      setYAMLErrors: vi.fn(),
+      yamlErrorLabel: Label.INCORRECT_YAML_CONFIGURATION_ERROR,
     };
   });
 
@@ -178,6 +180,28 @@ describe("CategoriesListing", () => {
     expect(textarea.value).toContain("default-space: my-space");
     expect(textarea.value).toContain("# Proxy & Mirror");
     expect(textarea.value).toContain("apt-http-proxy: http://proxy.example");
+  });
+
+  it("calls setYAMLErrors and stays in YAML mode when YAML has invalid keys", async () => {
+    render(
+      <Formik
+        initialValues={{ [FieldName.CONFIG_INPUT_MODE]: InputMode.YAML }}
+        onSubmit={vi.fn()}
+      >
+        <CategoriesListing {...defaultProps} />
+      </Formik>,
+    );
+
+    const textarea = screen.getByPlaceholderText(
+      Label.MODEL_CONFIG_PLACEHOLDER,
+    ) as HTMLTextAreaElement;
+    await userEvent.clear(textarea);
+    await userEvent.type(textarea, "unknown-field: value");
+
+    await userEvent.click(screen.getByRole("radio", { name: InputMode.LIST }));
+
+    expect(defaultProps.setYAMLErrors).toHaveBeenCalledOnce();
+    expect(screen.getByRole("radio", { name: InputMode.YAML })).toBeChecked();
   });
 
   it("filters categories by field name", () => {
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
@@ -16,6 +16,7 @@ import useDebounce from "hooks/useDebounce";
 
 import { InputMode } from "../../types";
 import StackList from "../StackList";
+import { YAMLErrorType } from "../YAMLErrorsModal/types";
 import type { ConfigFieldValue, FormFields } from "../types";
 import {
   buildYAML,
@@ -62,9 +63,9 @@ const CategoriesListing = ({
       }
 
       if (
-        errors.invalidKeys.length > 0 ||
-        errors.invalidValues.length > 0 ||
-        errors.otherErrors.length > 0
+        errors[YAMLErrorType.UNKNOWN_KEYS].length > 0 ||
+        errors[YAMLErrorType.INVALID_VALUES].length > 0 ||
+        errors[YAMLErrorType.OTHERS].length > 0
       ) {
         void setFieldTouched(yamlKey, true, false);
         // setFieldError must be called after setFieldTouched to prevent
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
@@ -21,6 +21,7 @@ import {
   buildYAML,
   filterCategoriesBySearch,
   getCategoriesWithVisibleFields,
+  validateAndParseYAML,
 } from "../utils";
 
 import type { Props } from "./types";
@@ -37,6 +38,8 @@ const CategoriesListing = ({
   searchPlaceholder,
   yamlPlaceholder,
   searchName,
+  setYAMLErrors,
+  yamlErrorLabel,
 }: Props): JSX.Element => {
   const id = useId();
   const { values, setFieldValue } = useFormikContext<
@@ -49,6 +52,32 @@ const CategoriesListing = ({
   const handleModeChange = (isListModeSelected: boolean): void => {
     if (!isListModeSelected) {
       void setFieldValue(yamlKey, buildYAML(categoriesList, values));
+    } else {
+      const { validValues, errors } = validateAndParseYAML(
+        values[yamlKey] ?? "",
+        categoriesList,
+        values,
+      );
+      Object.entries(validValues).forEach(([label, value]) => {
+        void setFieldValue(label, value);
+      });
+
+      if (
+        errors.invalidKeys.length > 0 ||
+        errors.invalidValues.length > 0 ||
+        errors.otherErrors.length > 0
+      ) {
+        void setFieldTouched(yamlKey, true, false);
+        // setFieldError must be called after setFieldTouched to prevent
+        // Formik's internal validation run from wiping the error message.
+        setTimeout(() => {
+          setFieldError(yamlKey, yamlErrorLabel);
+        }, 0);
+        setYAMLErrors({ errors, inputMode, yamlKey });
+        return;
+      }
+
+      setFieldError(yamlKey, undefined);
     }
 
     void setFieldValue(
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
@@ -42,9 +42,8 @@ const CategoriesListing = ({
   yamlErrorLabel,
 }: Props): JSX.Element => {
   const id = useId();
-  const { values, setFieldValue } = useFormikContext<
-    FormFields & Record<string, ConfigFieldValue>
-  >();
+  const { values, setFieldError, setFieldTouched, setFieldValue } =
+    useFormikContext<FormFields & Record<string, ConfigFieldValue>>();
   const [searchQuery, setSearchQuery] = useDebounce("", 250);
   const [changedOnly, setChangedOnly] = useState(false);
 
@@ -58,9 +57,9 @@ const CategoriesListing = ({
         categoriesList,
         values,
       );
-      Object.entries(validValues).forEach(([label, value]) => {
+      for (const [label, value] of Object.entries(validValues)) {
         void setFieldValue(label, value);
-      });
+      }
 
       if (
         errors.invalidKeys.length > 0 ||
@@ -122,7 +121,9 @@ const CategoriesListing = ({
         {title}
       </h5>
       <p className="u-no-margin--bottom p-text--small">
-        <a href={docsLink}>{docsLabel}</a>
+        <a href={docsLink} target="_blank">
+          {docsLabel}
+        </a>
       </p>
       <div className="u-flex u-flex--gap">
         <div>
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
@@ -53,8 +53,13 @@ const CategoriesListing = ({
     if (!isListModeSelected) {
       void setFieldValue(yamlKey, buildYAML(categoriesList, values));
     } else {
+      const yamlString = values[yamlKey];
+      if (!yamlString) {
+        void setFieldValue(inputMode, InputMode.LIST);
+        return;
+      }
       const { validValues, errors } = validateAndParseYAML(
-        values[yamlKey] ?? "",
+        yamlString,
         categoriesList,
         values,
       );
--- a/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
@@ -169,4 +169,73 @@ describe("ConfigsConstraints", () => {
       screen.getByRole("radio", { name: Label.DISABLE_ALL_COMMANDS }),
     ).toBeChecked();
   });
+
+  it("shows the error modal when switching to list mode with invalid YAML", async () => {
+    renderComponent(
+      <Formik
+        initialValues={{
+          [FieldName.CONFIG_INPUT_MODE]: InputMode.YAML,
+          [FieldName.CONSTRAINT_INPUT_MODE]: InputMode.LIST,
+          [FieldName.CONFIG_YAML]: "unknown-field: value",
+        }}
+        onSubmit={vi.fn()}
+      >
+        <ConfigsConstraints />
+      </Formik>,
+    );
+
+    const configsSection = screen.getByRole("region", {
+      name: Label.CONFIGS_TITLE,
+    });
+    await userEvent.click(
+      within(configsSection).getByRole("radio", { name: InputMode.LIST }),
+    );
+
+    const dialog = screen.getByRole("dialog", {
+      name: "Invalid values will be lost",
+    });
+    expect(dialog).toBeInTheDocument();
+    expect(
+      within(dialog).getByText(/invalid configuration values/i),
+    ).toBeInTheDocument();
+  });
+
+  it("wipes invalid YAML entries when switching to list mode forcefully", async () => {
+    renderComponent(
+      <Formik
+        initialValues={{
+          [FieldName.CONFIG_INPUT_MODE]: InputMode.YAML,
+          [FieldName.CONSTRAINT_INPUT_MODE]: InputMode.LIST,
+          [FieldName.CONFIG_YAML]:
+            "default-space: my-space\nunknown-field: value",
+        }}
+        onSubmit={vi.fn()}
+      >
+        <ConfigsConstraints />
+      </Formik>,
+    );
+
+    const configsSection = screen.getByRole("region", {
+      name: Label.CONFIGS_TITLE,
+    });
+    await userEvent.click(
+      within(configsSection).getByRole("radio", { name: InputMode.LIST }),
+    );
+    await userEvent.click(
+      screen.getByRole("button", { name: "Switch to list view" }),
+    );
+
+    expect(
+      screen.queryByRole("dialog", { name: "Invalid values will be lost" }),
+    ).not.toBeInTheDocument();
+    expect(
+      within(configsSection).getByRole("radio", { name: InputMode.LIST }),
+    ).toBeChecked();
+
+    // The valid field value was ported over
+    const defaultSpaceInput = within(configsSection).getByLabelText(
+      "default-space",
+    ) as HTMLInputElement;
+    expect(defaultSpaceInput.value).toBe("my-space");
+  });
 });
--- a/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
@@ -11,8 +11,12 @@ const ErrorList = ({
 }: {
   label: string;
   errors: YAMLValidationError[];
-}): JSX.Element | null =>
-  errors.length > 0 ? (
+}): JSX.Element | null => {
+  if (errors.length === 0) {
+    return null;
+  }
+
+  return (
     <ul className="u-no-padding--left u-no-margin--left">
       {label}:
       {errors.map(({ line, message }, index) => {
@@ -23,7 +27,8 @@ const ErrorList = ({
         );
       })}
     </ul>
-  ) : null;
+  );
+};
 
 const YAMLErrorsModal = ({
   errors,
--- a/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
@@ -3,23 +3,25 @@ import type { JSX } from "react";
 
 import { FieldName } from "../types";
 
-import type { YAMLErrors, YAMLErrorsModalProps } from "./types";
+import type { YAMLErrorsModalProps, YAMLValidationError } from "./types";
 
 const ErrorList = ({
   label,
   errors,
 }: {
   label: string;
-  errors: YAMLErrors["invalidKeys" | "invalidValues" | "otherErrors"];
+  errors: YAMLValidationError[];
 }): JSX.Element | null =>
   errors.length > 0 ? (
     <ul className="u-no-padding--left u-no-margin--left">
       {label}:
-      {errors.map((error, index) => (
-        <li className="u-sh3" key={`${label}-${index}`}>
-          Line {error.line}: {error.message}
-        </li>
-      ))}
+      {errors.map(({ line, message }, index) => {
+        return (
+          <li className="u-sh3" key={`${label}-${index}`}>
+            {line ? `Line ${line}: ${message}` : message}
+          </li>
+        );
+      })}
     </ul>
   ) : null;
 
--- a/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
@@ -3,6 +3,7 @@ import type { JSX } from "react";
 
 import {
   ENTITY_LABELS,
+  YAMLErrorType,
   type YAMLErrorsModalProps,
   type YAMLValidationError,
 } from "./types";
@@ -55,9 +56,15 @@ const YAMLErrorsModal = ({
           You have one or more invalid {ENTITY_LABELS[yamlKey]} values.{" "}
           <b>If you switch to list view, those values will be lost.</b>
         </p>
-        <ErrorList label="Invalid keys" errors={errors.invalidKeys} />
-        <ErrorList label="Invalid values" errors={errors.invalidValues} />
-        <ErrorList label="Other errors" errors={errors.otherErrors} />
+        <ErrorList
+          label="Invalid keys"
+          errors={errors[YAMLErrorType.UNKNOWN_KEYS]}
+        />
+        <ErrorList
+          label="Invalid values"
+          errors={errors[YAMLErrorType.INVALID_VALUES]}
+        />
+        <ErrorList label="Other errors" errors={errors[YAMLErrorType.OTHERS]} />
       </ConfirmationModal>
     </Portal>
   );
--- a/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
@@ -1,0 +1,58 @@
+import { ConfirmationModal, usePortal } from "@canonical/react-components";
+import type { JSX } from "react";
+
+import { FieldName } from "../types";
+
+import type { YAMLErrors, YAMLErrorsModalProps } from "./types";
+
+const ErrorList = ({
+  label,
+  errors,
+}: {
+  label: string;
+  errors: YAMLErrors["invalidKeys" | "invalidValues" | "otherErrors"];
+}): JSX.Element | null =>
+  errors.length > 0 ? (
+    <ul className="u-no-padding--left u-no-margin--left">
+      {label}:
+      {errors.map((error, index) => (
+        <li className="u-sh3" key={`${label}-${index}`}>
+          Line {error.line}: {error.message}
+        </li>
+      ))}
+    </ul>
+  ) : null;
+
+const YAMLErrorsModal = ({
+  errors,
+  yamlKey,
+  onConfirm,
+  onClose,
+}: YAMLErrorsModalProps): JSX.Element => {
+  const { Portal } = usePortal();
+
+  return (
+    <Portal>
+      <ConfirmationModal
+        title="Invalid values will be lost"
+        className="configs__error-modal"
+        cancelButtonLabel="Cancel"
+        confirmButtonLabel="Switch to list view"
+        confirmButtonAppearance="primary"
+        onConfirm={onConfirm}
+        close={onClose}
+      >
+        <p>
+          You have one or more invalid{" "}
+          {yamlKey === FieldName.CONFIG_YAML ? "configuration" : "constraint"}{" "}
+          values. <b>If you switch to list view, those values will be lost.</b>
+        </p>
+        <ErrorList label="Invalid keys" errors={errors.invalidKeys} />
+        <ErrorList label="Invalid values" errors={errors.invalidValues} />
+        <ErrorList label="Other errors" errors={errors.otherErrors} />
+      </ConfirmationModal>
+    </Portal>
+  );
+};
+
+export default YAMLErrorsModal;
--- a/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
@@ -1,9 +1,11 @@
 import { ConfirmationModal, usePortal } from "@canonical/react-components";
 import type { JSX } from "react";
 
-import { FieldName } from "../types";
-
-import type { YAMLErrorsModalProps, YAMLValidationError } from "./types";
+import {
+  ENTITY_LABELS,
+  type YAMLErrorsModalProps,
+  type YAMLValidationError,
+} from "./types";
 
 const ErrorList = ({
   label,
@@ -50,9 +52,8 @@ const YAMLErrorsModal = ({
         close={onClose}
       >
         <p>
-          You have one or more invalid{" "}
-          {yamlKey === FieldName.CONFIG_YAML ? "configuration" : "constraint"}{" "}
-          values. <b>If you switch to list view, those values will be lost.</b>
+          You have one or more invalid {ENTITY_LABELS[yamlKey]} values.{" "}
+          <b>If you switch to list view, those values will be lost.</b>
         </p>
         <ErrorList label="Invalid keys" errors={errors.invalidKeys} />
         <ErrorList label="Invalid values" errors={errors.invalidValues} />
--- a/src/pages/AddModel/ConfigsConstraints/constraintsCatalog.ts
+++ b/src/pages/AddModel/ConfigsConstraints/constraintsCatalog.ts
@@ -4,7 +4,7 @@ import { BOOLEAN_OPTIONS } from "consts";
 
 import type { CategoryDefinition } from "./types";
 
-const EMPTY_OPTION = { label: "Choose an option", value: "" };
+const EMPTY_OPTION = { label: "Unset", value: "" };
 
 export const CONSTRAINT_CATEGORIES: CategoryDefinition[] = [
   {
--- a/src/pages/AddModel/ConfigsConstraints/utils.test.ts
+++ b/src/pages/AddModel/ConfigsConstraints/utils.test.ts
@@ -408,13 +408,14 @@ describe("utils", () => {
       expect(errors.invalidKeys[0].line).toBe(1);
     });
 
-    it("reports other error for lines missing a colon", () => {
-      const { errors } = validateAndParseYAML("no-colon-here", categories);
-
-      expect(errors.otherErrors).toHaveLength(1);
-      expect(errors.otherErrors[0].message).toContain(
-        "Invalid format. Expected <key>: <value>",
+    it("reports other error for malformed YAML syntax", () => {
+      const { errors } = validateAndParseYAML(
+        "key: value\n  bad-indent: oops",
+        categories,
       );
+
+      expect(errors.otherErrors.length).toBeGreaterThanOrEqual(1);
+      expect(errors.otherErrors[0].line).toBeGreaterThan(0);
     });
 
     it("reports an invalid value for select fields with disallowed values", () => {
--- a/src/pages/AddModel/ConfigsConstraints/utils.test.ts
+++ b/src/pages/AddModel/ConfigsConstraints/utils.test.ts
@@ -36,7 +36,7 @@ describe("utils", () => {
           {
             label: "cores",
             description: "Number of cores",
-            isNumeric: true,
+            valueType: "number",
           },
         ],
       },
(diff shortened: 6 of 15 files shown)


src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+  yaml.split("\n").forEach((line, index) => {
+    const lineNumber = index + 1;
+    const trimmedLine = line.trim();
+
+    // Ignore empty lines and comments
+    if (!trimmedLine || trimmedLine.startsWith("#")) {
+      return;
+    }
+
+    const separatorIndex = trimmedLine.indexOf(":");
```
andogq: This does not handle colons in keys.

```yaml
"my:key": value
```

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    if (separatorIndex === -1) {
+      errors.otherErrors.push({
+        line: lineNumber,
+        message: "Invalid format. Expected <key>: <value>",
+      });
+      return;
+    }
+
+    const key = trimmedLine.slice(0, separatorIndex).trim();
```
andogq: Keys can also be surrounded in quotes.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    const yamlValue = trimmedLine.slice(separatorIndex + 1).trim();
+    // Handle quoted empty string values correctly. This is done for double quotes only as buildYAML always emits "".
```
andogq: Should also handle single quotes, as there's no guarantee that double quotes will be output in the future.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    const normalizedValue = yamlValue === '""' ? "" : yamlValue;
```
andogq: Should surrounding quotes also be stripped from non-empty keys and values?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+        });
+        return;
+      }
+    }
+
+    // If the field is numeric, validate the value
+    if (
+      field.isNumeric &&
+      normalizedValue &&
+      !/^-?\d+(\.\d+)?$/.test(normalizedValue)
```
andogq: Do not use a regex. Use `new Number` with `isNaN`.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    ) {
+      errors.invalidValues.push({
+        line: lineNumber,
+        message: `Invalid type for ${key}. Expected a number`,
+      });
+      return;
+    }
+
+    validValues[key] = normalizedValue;
```
andogq: Parsed value (eg number) is not saved, only the strings.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+  );
+
+  const validValues: Record<string, string> = {};
+  const errors: YAMLErrors = {
+    invalidKeys: [],
+    invalidValues: [],
+    otherErrors: [],
+  };
+
+  yaml.split("\n").forEach((line, index) => {
```
andogq: Don't use `forEach` for iteration. Prefer to use a normal `for` loop with a manual counter.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+      });
+      return;
+    }
+
+    validValues[key] = normalizedValue;
+  });
+
+  // For fields that were previously changed but absent from YAML, include
+  // their catalog default so callers can reset them in a single pass.
+  fieldsByLabel.forEach((field) => {
```
andogq: Same here, do not use `forEach` for iteration.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    if (
+      !(field.label in validValues) &&
+      isConfigChanged(field.label, currentValues, field.defaultValue)
+    ) {
+      validValues[field.label] = field.defaultValue?.toString() ?? "";
```
andogq: I don't believe this will correctly handle different types.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
     }))
     .filter((category) => category.fields.length > 0);
 };
+
+export function validateAndParseYAML(
```
andogq, line 166: I don't think this YAML parsing is a good idea, it does not account for many YAML features which could feasibly be entered by a user including:

- multiline strings
- boolean values (`false`/`off`/`no`/etc)
- nubmers (`1_234`, `1.234e4`, `0x1234`)

Similarly, restoring the default value does not seem very robust (eg. If the default value is `"off"`, it will be inserted as `key: off`, which will parsed as `key: false`).

Given all of these edge cases, I think it would be more appropriate to use an off-the-shelf YAML parser.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+  yaml: string,
+  categories: CategoryDefinition[],
+  currentValues: Record<string, string> = {},
+): {
+  validValues: Record<string, string>;
+  errors: YAMLErrors;
+} {
+  const fieldsByLabel = new Map(
```
andogq: Why is a Map used here?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +110,109 @@ export const filterCategoriesBySearch = (
...
+    invalidValues: [],
+    otherErrors: [],
+  };
+
+  yaml.split("\n").forEach((line, index) => {
+    const lineNumber = index + 1;
+    const trimmedLine = line.trim();
+
+    // Ignore empty lines and comments
+    if (!trimmedLine || trimmedLine.startsWith("#")) {
```
andogq: This does not handle comments which appear within the line. Also, it won't work if the `#` appears in a multi-line string.

src/pages/AddModel/ConfigsConstraints/utils.test.ts
```diff
@@ -324,4 +336,104 @@ describe("utils", () => {
...
+    it("parses the YAML into valid key-value pairs", () => {
+      const { validValues, errors } = validateAndParseYAML(
+        "# a comment\n\ndefault-space: custom\nlogging-config: debug\ncores: 4",
+        categories,
+      );
+
+      expect(validValues).toEqual({
+        "default-space": "custom",
+        "logging-config": "debug",
+        cores: "4",
```
andogq: If this field is numeric, should it be returned as a number?

andogq (changes requested):
I've only left comments on the YAML handling for now. I think a pre-built solution would be more robust for this use case.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: there are no config keys in juju's catalog with colons in them.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Ah, I considered it but then I had an edge case in mind `" "` that it wouldn't be able to handle so I reverted, completely ignoring the `trim` at the top. Replacing this.

Ninfa-Jeon: one other edge case would be: `Number("1e5")` returns `100000` and is not `NaN`, so scientific notation would pass. just saying

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: the restoring defaults is only a problem if we migrate to a real YAML parser. My current custom line parser reads everything as raw strings regardless, so `key: off` comes back as the string "off" correctly. The vulnerability only materialises at the point a real YAML library is introduced. similarly for numbers, with my regex, I was specifically ignoring values like `1_234, 1.234e4, 0x1234` because we are not looking at values that are numbers at runtime.

Ninfa-Jeon: however, I am planning an improvement of the following sort:

1. We quote all non-numeric values with double quotes, i.e., `key: off` becomes `key: "off"`, `key: true` becomes `key: "true"`, `key: null` becomes `key: "null"`, etc. Bare numbers like `17` are left unquoted since they're unambiguous.
2. We strip surrounding double quotes from any quoted value (not just `""` for empty), so `"off"` correctly parses back to the string `off`.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: I didn't presume this would happen because we are the ones adding comments on switch from list to yaml

src/pages/AddModel/ConfigsConstraints/utils.test.ts
Ninfa-Jeon: no, all formik values are stored as strings and juju expects them as strings too

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: why so? this seems correct for our case to me

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: This is intended, as yaml supports scientific notation, hex, thousands separators, etc all for numbers

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: From the [top-level comment](https://github.com/canonical/juju-dashboard/pull/2303#discussion_r3471502614):

> Similarly, restoring the default value does not seem very robust (eg. If the default value is "off", it will be inserted as key: off, which will parsed as key: false).

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: Users may enter comments in the free-form yaml

src/pages/AddModel/ConfigsConstraints/utils.test.ts
andogq: Seems like Juju stores it as an integer, and the ambiguity between integers and strings in yaml is what allows it work

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: I understand there isn't now, but there's no guarantee for the future, and also nothing preventing the user from typing it themselves.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: > so key: off comes back as the string "off" correctly

This is not not correct for YAML.

```yaml
myKey: off
anotherKey: "off"
```

is equivalent to

```json
{"myKey":false,"anotherKey":"off"}
```

> because we are not looking at values that are numbers at runtime

Given that Juju does store these values as integers, shouldn't we also have validation that they're valid? `1.234_567e10` is a perfectly valid number in YAML.

---

I still think you should work towards integrating a standard YAML parser. Juju uses standard YAML, users are expecting standard YAML, it will only cause bugs, confusion, and frustration if we have our own sets of rules and behaviours for interacting with YAML.

Ninfa-Jeon:
> I think a pre-built solution would be more robust for this use case.

As discussed, I have introduced `yaml` library into our codebase for parsing yaml text input, allowing a better handling of all edge cases.
The parsed yaml is then validated for our case. Closing comments that were about the older logic.

src/pages/AddModel/ConfigsConstraints/utils.ts

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq, line 59: This should be replaced with the yaml stringify feature from the package.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+  const invalidValues: YAMLValidationError[] = [];
+
+  const { errors: docErrors, contents } = parseDocument(yamlString);
+
+  // Walk the top-level key-value pairs and validate each against the catalog.
+  if (isMap(contents)) {
+    for (const pair of contents.items) {
+      if (!isPair(pair) || !isScalar(pair.key)) {
+        continue;
+      }
```
andogq: Should there be an error here?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+
+      const key = String(pair.key.value);
```
andogq: What else could `key` be, if not a string?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+      if (!field) {
+        invalidKeys.push({ line, message: `Unknown key: ${key}` });
+        continue;
+      }
+
+      // Coerce parsed value to string as the library already handled booleans,
+      // nulls, numbers, quoted strings etc. correctly.
+      const rawValue = isScalar(pair.value) ? pair.value.value : null;
+      const value =
+        rawValue === null || rawValue === undefined ? "" : rawValue.toString();
```
andogq: This seems a bit strange. This seems to just replace non-scalar values with an empty string, I feel like the user should be notified of this with an error. Additionally, why do non-string scalar values need to be turned into a string? This feels needlessly lossy, especially given that the Juju API doesn't use strings for all values.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+
+      // If the field has a predefined set of allowed values, check if valid.
+      if (field.input?.type === "select") {
+        const allowedValues = (field.input.options ?? []).map(
+          ({ value: optionValue }) => optionValue,
+        );
+        if (!allowedValues.includes(value)) {
```
andogq: Can condense this with `!Array.some`, preventing excessive iterations.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+          invalidValues.push({
+            line,
+            message: `Invalid value for ${key}. Expected one of: ${allowedValues.join(", ")}`,
+          });
+          continue;
+        }
+      }
+
+      // If the field is numeric, validate the value.
+      if (field.isNumeric && value && isNaN(Number(value))) {
```
andogq: This wouldn't be required if everything wasn't turned into a string before.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+  );
+
+  const validValues: Record<string, string> = {};
+  const invalidKeys: YAMLValidationError[] = [];
+  const invalidValues: YAMLValidationError[] = [];
+
+  const { errors: docErrors, contents } = parseDocument(yamlString);
+
+  // Walk the top-level key-value pairs and validate each against the catalog.
+  if (isMap(contents)) {
```
andogq: I feel like an error should be provided if it's not a top-level map

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -108,3 +112,103 @@ export const filterCategoriesBySearch = (
...
+  // their catalog default so callers can reset them in a single pass.
+  for (const label in fieldsByLabel) {
+    const { defaultValue } = fieldsByLabel[label];
+    if (
+      !(label in validValues) &&
+      isConfigChanged(label, currentValues, defaultValue)
+    ) {
+      validValues[label] = defaultValue?.toString() ?? "";
+    }
+  }
```
andogq, line 312: If I'm understanding this correctly, the default value will be inserted whenever an invalid value is parsed. Is this actually what's desired? I feel like we'd want to warn the user, so they don't potentially loose their work.

src/pages/AddModel/ConfigsConstraints/utils.test.ts
andogq: I still think this should be a number, given the Juju API does parse them as numbers.

src/pages/AddModel/ConfigsConstraints/utils.test.ts
```diff
@@ -324,4 +336,105 @@ describe("utils", () => {
...
+    });
+
+    it("normalizes double-quoted empty string to actual empty string", () => {
+      const { validValues } = validateAndParseYAML(
+        'default-space: ""',
+        categories,
+      );
+
+      expect(validValues["default-space"]).toBe("");
+    });
```
andogq: Not sure if this test is required, as it's just testing the yaml package.

src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {label}:
+      {errors.map(({ line, message }, index) => {
+        return (
+          <li className="u-sh3" key={`${label}-${index}`}>
+            {line ? `Line ${line}: ${message}` : message}
+          </li>
+        );
+      })}
+    </ul>
+  ) : null;
```
andogq: This ternary makes it a bit difficult to immediately see.


```suggestion
}): JSX.Element | null => {
  if (errors.length === 0) {
    return null;
  }
  
  return (
    <ul className="u-no-padding--left u-no-margin--left">
      {label}:
      {errors.map(({ line, message }, index) => {
        return (
          <li className="u-sh3" key={`${label}-${index}`}>
            {line ? `Line ${line}: ${message}` : message}
          </li>
        );
      })}
    </ul>
  );
}
```

src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+        className="configs__error-modal"
+        cancelButtonLabel="Cancel"
+        confirmButtonLabel="Switch to list view"
+        confirmButtonAppearance="primary"
+        onConfirm={onConfirm}
+        close={onClose}
+      >
+        <p>
+          You have one or more invalid{" "}
+          {yamlKey === FieldName.CONFIG_YAML ? "configuration" : "constraint"}{" "}
```
andogq: Should the labels be used here?

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
```diff
@@ -37,18 +38,45 @@ const CategoriesListing = ({
...
       void setFieldValue(yamlKey, buildYAML(categoriesList, values));
+    } else {
+      const { validValues, errors } = validateAndParseYAML(
+        values[yamlKey] ?? "",
+        categoriesList,
+        values,
+      );
+      Object.entries(validValues).forEach(([label, value]) => {
+        void setFieldValue(label, value);
+      });
```
andogq: Use a `for..of` loop instead

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
```diff
@@ -37,18 +38,45 @@ const CategoriesListing = ({
...
+        errors.invalidKeys.length > 0 ||
+        errors.invalidValues.length > 0 ||
+        errors.otherErrors.length > 0
+      ) {
+        void setFieldTouched(yamlKey, true, false);
+        // setFieldError must be called after setFieldTouched to prevent
+        // Formik's internal validation run from wiping the error message.
+        setTimeout(() => {
+          setFieldError(yamlKey, yamlErrorLabel);
+        }, 0);
```
andogq, line 80: Just await the returned promise.

```suggestion
        await setFieldTouched(yamlKey, true, false);
        setFieldError(yamlKey, yamlErrorLabel);
```

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
andogq, line 1: Would be good to see a test that ensures YAML errors are reset when toggling YAML mode.

1. Form with invalid value
2. Switch to YAML mode (should show errors)
3. Switch to form mode
4. Fix error value
5. Switch to YAML mode (should not have errors)

andogq (commented):
Nice work with the `yaml` package, looks much more robust now.

A bunch of more minor changes that'd be good to see, particularly:
- Using the correct types for constraint values
- Use `yaml` package for serialising data to YAML
- Consistent error reporting (whenever a user's value is changed/replaced, there should be something that notifies them)

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: The function from the package will not add category comments which means they'll have to be somehow added manually. Additionally, `stringify({ key: value })` adds a newline and produces `key: value\n` requiring `.trimEnd()`. And it's more expensive than the simple ternary for no real benefit here. The only case it handles better is values containing special characters like `:`, `#`, `[`, `{` , which in our catalog are unlikely but possible. I can however make the following change if you are okay with the added complexity:

```
// current
const quotedValue = value === "" ? '""' : value;
return `${field.label}: ${quotedValue}`;

// with yaml.stringify
return stringify({ [field.label]: value }).trimEnd();
```

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: But doesn't `includes` short circuit on first match making this somehow a similarly performing code?

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: No, the default value is inserted when a previously changed key-value pair is removed from the yaml. This means the user isn't explicitly setting the config which is why we need to revert it to default. Yes, it is desired.

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
Ninfa-Jeon: No, this doesn't work because without the `setTimeout`, the error state is overwritten and so the field is not marked with red on the first render.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: will have to add `as string` cast but maybe thats a better choice anyway

src/pages/AddModel/ConfigsConstraints/utils.test.ts
Ninfa-Jeon: its not the juju API, let me correct myself, but the formik state that expects it all as strings

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Considering the check `!isScalar()` that we are doing later, both are defensive guards for malformed or unusual YAML that the library itself would have already flagged in `docErrors`. Silently skipping them is fine because:

1. If the YAML has structural issues, `docErrors` already reports them.
2. If it somehow has a non-scalar key without a parse error, it's an unsupported edge case we don't need to surface separately.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Check [this](https://github.com/canonical/juju-dashboard/pull/2298#discussion_r3121901900) comment on an older PR for why we are sticking to string values for all types. Since sending strings does not break the API, we never tried to switch the values back to numbers

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
Ninfa-Jeon: The  proposed test flow isn't achievable with this implementation. Invalid YAML can only be introduced in YAML mode, list mode inputs are constrained by form fields and cannot produce invalid states. Switching from list to YAML never triggers errors; only the reverse (YAML → list) does.

The existing test "calls setYAMLErrors and stays in YAML mode when YAML has invalid keys" already covers the error case: invalid YAML → attempt switch → error shown, mode stays YAML. The error is cleared when the user switches back to YAML mode to edit, which is the only meaningful reset path given the UX design.

Happy to add a test specifically asserting setYAMLErrors is not called when YAML is valid (covering the "no error" path explicitly) if that would help close this comment.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/utils.test.ts
Ninfa-Jeon: removed

src/pages/AddModel/ConfigsConstraints/YAMLErrorsModal/YAMLErrorsModal.tsx
Ninfa-Jeon: alright

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: The library does support adding comments to a document: https://eemeli.org/yaml/#comments-and-blank-lines
Trailing new lines in documents is pretty standard, and shouldn't be an issue for anything parsing the yaml (ie. you shouldn't need to use `.trimEnd` anywhere).

The benefit of using the library is that we're protected against all of the issues we've previously discussed when using YAML.

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
andogq: I just tried it locally, it works fine if you remove the two trailing parameters of `setFieldTouched` (which will default to `true` for `isTouched`, and `true` for `shouldValidate`, which is fine as we're not using a validator with Formik.

```suggestion
        await setFieldTouched(yamlKey);
        setFieldError(yamlKey, yamlErrorLabel);
```

src/pages/AddModel/ConfigsConstraints/utils.test.ts
andogq: I'm not sure if I'm misunderstanding here, but Formik seems to accept numbers and booleans fine, if you use the correct `type` attribute on the `Field`.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: The second `isScalar` call is checking a different value, so it's not relevant. The yaml library only returns valid items, and stores malformed items in the `docErrors` you mentioned. So a non-`Pair` item could be present here, but would be silently ignored. Given we already have a mechanism for reporting errors, it would be good to include it here.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: `pair.key.toString()` should suffice, as `Node` appears to have a suitable implementation for this.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: `includes` does, but `map` doesn't. So you're dealing with all the extra allocations plus a full iteration.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: Happy to help you with Formik if that's the problem here, but from my testing it seems to handle the conversion to/from numbers for us. Additionally, it lets us lean on browser features for validation/input methods, which makes it better for accessibility.

andogq (commented):
Still a few more comments, but overall it's in a good place.

- In regards to manually converting all values into a string, my main concern here is that it's an extra step (which may be lossy) which we're needlessly introducing. Formik supports numbers, the API supports numbers, so it just seems like we're adding a new potential point for bugs being introduced. Even if we're super concerned about funny things in the form, we can use `Yup` for validation, which may be preferable to hand-rolling validation.

- The `yaml` library does support attaching comments to nodes, so we should be using that instead of manually serialising the data ourselves.

Some notes from QA:

- Default field values are not present in the YAML view, but are filled in the form view. The default values can be used as a placeholder (as you already have), which indicates that if the user doesn't put anything in, that's what the API will insert. However, if there is a value in the field, I would expect it to show in the YAML.
  - Strings/numbers: display default in placeholder
  - Boolean/select: add a third 'unspecified' option which is the default selected (eg. pic below)

Image

- (related to above) If the user enters a value in YAML which is identical to the default value, it is removed from the input after toggling to/from field input mode. I would expect that if a user has interacted with a field (even just to revert it back to the default value), it should be retained, and they can delete the line themselves.

- When validating yaml upon returning to field view, JS values are used for validation, instead of YAML values (this partially seems to be an issue with the config catalogue representing booleans as selects, and continuing to rely on JS number parsing for validation, instead of relying on the `yaml` parser).

  - booleans: true/on/yes/etc. should all be accepted
 
  - numbers: yaml numbers like `1_230.15e+10` should be accepted
  
  - (I know this one's a bit annoying, but I'm imagining a user copying/pasting their YAML config from elsewhere to pre-fill the form. If we tell them that it's YAML, then we've got to support YAML.)
  
- If an invalid value is inserted into a field, then it is switched to YAML mode, it will be replaced with an empty string. Returning back to field view deletes the value completely (even if there was a default value initially). I think there's two ways to fix this:

  - Perform the same validation before switching to the YAML view, highlighting value errors.
  
  - Perform validation checks individually on each field, constantly, in both YAML and field view. Allow user to freely swap between the two without modifying their values, but prevent them from progressing to the next stage of model creation if there's errors.
  
- And finally two small things which aren't even caused by this PR:
  
  - It'd be great to have a `target` attribute on the 'Learn about model creation' link, to prevent it from loading in the current tab. I accidentally lost my configuration a few times when clicking between the views :')

  - The `vpc-id` config field doesn't seem correct. It's currently in the catalogue as a boolean, however in the [docs](https://canonical.com/juju/docs/juju-cli/latest/reference/cloud/list-of-supported-clouds/the-amazon-ec2-cloud-and-juju/#vpc-id) it's a string. Not sure if other values are a problem, but I noticed this since it's at the top of the list.

---

To sum all of this up, I would try advocate for a separation of concerns:

- `yaml` package: handle parsing/stringifying to/from YAML. Produces errors for malformed YAML, and nothing else.
- `yup`: handles validation of the JSON representation of the data. Dynamically build a schema based on the constraints catalogue you already have.
- `formik`: handles all of the form interactions/change detection/error visibility as you already have, however it relies on the same `yup` schema for consistency.

Image

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: oh wait you meant the the entire block..my bad i read that wrong

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Let's discuss this with Huw tomorrow as he was the one that requested the change

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Will still need to map over the options for the error message. it will look something like:
```
if (field.input?.type === "select") {
        const options = field.input.options ?? [];
        if (
          !options.some(({ value: allowedValue }) => allowedValue === value)
        ) {
          invalidValues.push({
            line,
            message: `Invalid value for ${key}. Expected one of: ${options.map(({ value: allowedValue }) => allowedValue).join(", ")}`,
          });
          continue;
        }
      }
```
let me know if this suits better because it is still making two passes

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
Ninfa-Jeon: it does not work


https://github.com/user-attachments/assets/76be1c1f-bb2e-4b3a-9c6f-43d9e2c60e86

Ninfa-Jeon:
> The yaml library does support attaching comments to nodes, so we should be using that instead of manually serialising the data ourselves.

Done.

> * Default field values are not present in the YAML view, but are filled in the form view. The default values can be used as a placeholder (as you already have), which indicates that if the user doesn't put anything in, that's what the API will insert. However, if there is a value in the field, I would expect it to show in the YAML.
>   
>   * Strings/numbers: display default in placeholder
>   * Boolean/select: add a third 'unspecified' option which is the default selected (eg. pic below)
>  If the user enters a value in YAML which is identical to the default value, it is removed from the input after toggling to/from field input mode. I would expect that if a user has interacted with a field (even just to revert it back to the default value), it should be retained, and they can delete the line themselves.

For the above, I cannot make the change as required because we are not aiming for that UX. -> Could be a follow-up piece of work if UX-approved.

> And finally two small things which aren't even caused by this PR:
> 
> It'd be great to have a target attribute on the 'Learn about model creation' link, to prevent it from loading in the current tab. I accidentally lost my configuration a few times when clicking between the views :')
> 
> The vpc-id config field doesn't seem correct. It's currently in the catalogue as a boolean, however in the [docs](https://canonical.com/juju/docs/juju-cli/latest/reference/cloud/list-of-supported-clouds/the-amazon-ec2-cloud-and-juju/#vpc-id) it's a string. Not sure if other values are a problem, but I noticed this since it's at the top of the list.

Fixed

> booleans: true/on/yes/etc. should all be accepted

Depends on the version of the library we are using. -> can be addressed in another PR if it is still an issue.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -110,3 +124,174 @@ export const filterCategoriesBySearch = (
...
+  switch (type) {
+    case YAMLErrorType.OTHERS:
+      return `Invalid format. ${options?.context ?? ""}`;
+    case YAMLErrorType.UNKNOWN_KEYS:
+      return `Unknown key: ${options?.key}`;
+    case YAMLErrorType.INVALID_VALUES:
+      return options?.expectedValue
+        ? `Invalid value for ${options?.key}. Expected ${options?.expectedValue}`
+        : `Invalid value for ${options?.key}`;
+  }
```
andogq, line 163: No point having that ternary

```suggestion
    case YAMLErrorType.INVALID_VALUES:
      const message = `Invalid value for ${options?.key}`;
      if (options?.expectedValue) {
        return `${message}. Expected ${options?.expectedValue}`;
      }
      return message;
  }
```

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.tsx
```diff
@@ -37,18 +39,45 @@ const CategoriesListing = ({
...
   const [searchQuery, setSearchQuery] = useDebounce("", 250);
   const [changedOnly, setChangedOnly] = useState(false);
 
   const isListMode = values[inputMode] === InputMode.LIST;
   const handleModeChange = (isListModeSelected: boolean): void => {
     if (!isListModeSelected) {
       void setFieldValue(yamlKey, buildYAML(categoriesList, values));
+    } else {
+      const { validValues, errors } = validateAndParseYAML(
+        values[yamlKey] ?? "",
```
andogq: I feel like we can just early return if this is not present. Otherwise I think it would produce an error since it's not a top level map?

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
andogq, line 1: Would be good to see some tests for yaml-specific values

```yaml
my-boolean: off
big-number: 1_234.56e10
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
```diff
@@ -169,4 +169,73 @@ describe("ConfigsConstraints", () => {
       screen.getByRole("radio", { name: Label.DISABLE_ALL_COMMANDS }),
     ).toBeChecked();
   });
+
+  it("shows the error modal when switching to list mode with invalid YAML", async () => {
```
andogq, line 173: Nice test

src/pages/AddModel/ConfigsConstraints/constraintsCatalog.ts
```diff
@@ -4,7 +4,7 @@ import { BOOLEAN_OPTIONS } from "consts";
 
 import type { CategoryDefinition } from "./types";
 
-const EMPTY_OPTION = { label: "Choose an option", value: "" };
```
andogq, line 7: 👌

src/pages/AddModel/ConfigsConstraints/utils.test.ts
```diff
@@ -359,4 +372,210 @@ describe("utils", () => {
       });
     });
   });
+
+  describe("validateAndParseYAML", () => {
+    it("parses the YAML into valid key-value pairs", () => {
+      const { validValues, errors } = validateAndParseYAML(
+        "# a comment\n\ndefault-space: custom\nlogging-config: debug\ncores: 4",
```
andogq: Would also be good to see some weird yaml boolean and numbers here

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -46,23 +53,30 @@ export const buildYAML = (
...
+      // outputs them as unquoted true/false rather than quoted strings.
+      fieldObject[field.label] =
+        field.valueType === "boolean"
+          ? value === "true" || value === true
+          : value;
+    }
 
-  return yamlSections.join("\n\n");
+    // Prepend the category name as a YAML comment, then the stringified fields.
+    sections.push(`# ${category.category}\n${stringify(fieldObject)}`);
```
andogq: It'd be better to see modification of the document AST to add comments. That'll save you from having to manually stitch together document fragments too, which'll be more robust. See [the docs](https://eemeli.org/yaml/#identifying-node-types)

andogq (approved):
I think we finally got there!

Two small changes for you to do. First set is just to add tests to make sure we're truely able to handle all of what YAML can throw at us. Second thing is to use the YAML library to create comments on the document nodes directly.

PR #2305, opened 2026-04-24
feat(connection-manager): different connection logic

andogq:
Depending on connection URL path, use different connection logic, allowing for different behaviour for models vs controllers.

### DIFF
--- a/src/store/middleware/connection/connection-manager.ts
+++ b/src/store/middleware/connection/connection-manager.ts
@@ -4,6 +4,7 @@ import { Auth } from "auth";
 import {
   CLIENT_VERSION,
   generateConnectionOptions,
+  LOGIN_TIMEOUT,
   loginWithBakery,
 } from "juju/api";
 import { Label, type ConnectionWithFacades } from "juju/types";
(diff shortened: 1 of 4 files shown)


src/store/middleware/connection/connection-manager.ts
```diff
@@ -14,16 +18,114 @@ type Hooks = {
...
+ * Map of URL paths to a connection handler, with `DefaultHandler` suitable for any connection kind.
+ */
+export const CONNECTION_HANDLERS_BY_PATH: Record<
+  string | typeof DefaultHandler,
+  ConnectionHandler
+> = {
+  /**
+   * Handler used by default
+   */
+  [DefaultHandler]: async (wsControllerURL, credentials) => {
```
huwshimi: It might be helpful to rename the ws url to something generic as it could be a controller or model e.g. `wsURL`.

src/store/middleware/connection/connection-manager.ts
```diff
@@ -14,16 +18,114 @@ type Hooks = {
...
+    const options = generateConnectionOptions(true);
```
huwshimi, line 62: Do we still need the pinger for models? My guess is we could still have responses that could take a long time to be returned from Juju.

src/store/middleware/connection/connection-manager.ts
```diff
@@ -14,16 +18,114 @@ type Hooks = {
...
+    const instanceCredentials = Auth.instance.determineCredentials(credentials);
+    const { conn, logout } = await connectAndLogin(
```
huwshimi: Do we want to use a timeout like we do in `connectAndLoginWithTimeout` for these connections?

huwshimi (commented):
A couple of questions about differences between this and the current implementation.

src/store/middleware/connection/connection-manager.ts
andogq: In [`connectToModel`](https://github.com/canonical/juju-dashboard/blob/e3c04c42a38d730fec8cc8eb8e3c8185f1465072/src/juju/api.ts#L393) (the existing method that's used), it uses the pinger.

andogq: Just pushed up the pinger loop connection

huwshimi (approved):

PR #2306, opened 2026-04-24
feat(connection-manager): add `reconnect` callback to managed connections

andogq:
This allows a connection user to request the connection to be reconnected. This does not do anything to assist when a connection is closed elsewhere.

### DIFF
--- a/src/store/middleware/connection/connection-manager.ts
+++ b/src/store/middleware/connection/connection-manager.ts
@@ -13,19 +13,23 @@ import { Label, type ConnectionWithFacades } from "juju/types";
 import type { AuthCredential } from "store/general/types";
 import { logger } from "utils/logger";
 
+export type ManagedConnection = {
+  reconnect: () => Promise<ManagedConnection>;
+} & ConnectionWithFacades;
+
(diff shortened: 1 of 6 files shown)


src/store/middleware/connection/connection-manager.ts
```diff
@@ -210,17 +211,38 @@ export class ConnectionManager {
...
       wsControllerURL,
       credentials,
     );
-    const { connection } = connectionResult;
 
-    this.connections.set(wsControllerURL, connectionResult);
-    connectionPromise.resolve(connection);
+    // Force cast the connection type, then add the extra fields.
+    const managedConnection = connection as unknown as ManagedConnection;
+    managedConnection.reconnect = async (): Promise<ManagedConnection> => {
```
huwshimi: You can avoid the `as unknown` (which will mask issues) by using a spread or object assign:

```suggestion
    const managedConnection = Object.assign(connection, {
      reconnect: async (): Promise<ManagedConnection> => 
```

huwshimi (approved):
Looks good, thanks.

PR #2307, opened 2026-04-24
feat: disable upgrade menu item if there are no upgrades

huwshimi:
## Done

- Created a hook for fetching upgrades data and a selector for the loading state.
- Disable upgrade menu item if there are no upgrades.

## QA

- Create a model that can be upgraded and a model that has no upgrades.
- In the model list open the action menu next to the model that can't be upgraded and the upgrade menu item should be disabled.
- For the model that can be upgraded the menu item should be enabled.

## Details

https://warthogs.atlassian.net/browse/JUJU-9696

### DIFF
--- a/src/store/juju/selectors.test.ts
+++ b/src/store/juju/selectors.test.ts
@@ -154,6 +154,7 @@ import {
   getModelUpgradeVersions,
   getRecommendedVersions,
   getModelUpgrades,
+  getModelUpgradeDataLoaded,
 } from "./selectors";
 
 describe("selectors", () => {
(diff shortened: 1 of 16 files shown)


src/store/juju/selectors.test.ts
```diff
@@ -3630,6 +3683,89 @@ describe("getModelUpgradeVersions", () => {
...
+      },
+    });
+    expect(getModelUpgradeDataLoaded(state, "abc123")).toBe(true);
+  });
+
+  it("is not loaded when there is no version data", () => {
+    const state = rootStateFactory.build({
+      juju: {
+        supportedJujuVersions: supportedJujuVersionsStateFactory.build({
+          data: undefined,
```
andogq: To be pedantic, it will be `null` if it's not loaded.

src/store/juju/selectors.ts
```diff
@@ -1427,6 +1427,44 @@ export const getModelMigrationTargets = createSelector(
...
+        highest = version;
+      }
+      return highest;
+    }, null) ?? null,
+);
+
+export const getNextSupportedVersion = createSelector(
+  [
+    getSupportedJujuVersions,
+    (_state, version?: null | string): null | string | undefined => version,
```
andogq, line 1444: Where does `undefined` come from? Can we keep it as just `null | string`?

andogq (approved):

src/store/juju/selectors.test.ts
huwshimi: Thanks, done.

src/store/juju/selectors.ts
huwshimi: For this specific usage it's because a model not be in the store, in which case [we won't have a version](
https://github.com/canonical/juju-dashboard/pull/2307/changes#diff-77da4e2d5db38cc088055cb909c594de3c634a55544731a15c6702903a3a5d46R160), but because selectors are hooks they don't conditionally run so will be called even if the model doesn't get called.

It could pass `modelVersion ?? null` if you prefer, but I generally like to keep hooks/selectors as flexible as possible so they are simple to call.

andogq: That's fine with me then!

PR #2295, opened 2026-04-16
feat: model migration calls

andogq:
> [!important]
> Relies on #2300 
> [Relevant review range](https://github.com/canonical/juju-dashboard/pull/2295/changes/49b312c4a1f002b60c90cb6a97bfab95aca4b2cb)

PR #2308, opened 2026-04-24
feat: add disable-command chained operation to add-model

Ninfa-Jeon:
## Done

- Added UI for disabling commands on Configs&Constraints tab
- Added a process using `createProcess` to trigger  the `switchBlockOn` method for disabling commands
- Triggering this process from within the model-poller right after the addmodel passes
- Added required tests

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Navigate to AddModel page
- Fill the mandatory details and navigate to the next tab
- Here, try selecting one of the given options and create a model
- The model should get created and the listed commands should now be blocked on this model
- Easiest way to test this is trying to trigger a destroy-model on this (via dashboard or CLI). This operation should fail for all three block options.
- To verify that the destroy-model is being blocked due to the disable operation, you could shell into your multipass instance, switch to your model and run `juju enable-command {block-type}`. Trying to destroy the model after this should succeed.
- The addmodel passes regardless of your interaction with this feature
Note: No success or failure notifications are put in place at the moment as they are under UX consideration.

## Details

https://warthogs.atlassian.net/browse/JUJU-9681

## Screenshots

Screenshot 2026-04-24 at 6 50 36 PM


src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <div className="disabled-commands__option" key={value}>
+              <RadioInput
+                checked={values.disabledCommands === value}
+                name="disableCommands"
+                label={label}
+                onChange={() => {
+                  void setFieldValue("disabledCommands", value);
+                }}
+                value={value}
+              />
```
huwshimi: Rather than managing the state yourself, you can just use a FormikField (and also using a `FieldName` enum like you did in your other PR will help prevent errors):

```suggestion
              <FormikField
                type="radio"
                name={FieldName.DISABLED_COMMANDS}
                label={label}
                value={value}
              />
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -0,0 +1,54 @@
...
+              <div className="p-text--small u-no-margin--bottom">
+                {description}
+                {disabledCommands && disabledCommands.length > 0 ? (
```
huwshimi: `disabledCommands` and `disabledCommands.length > 0` mean the same thing in javascript (an empty string is false).

src/store/middleware/process/index.ts
```diff
@@ -1,6 +1,8 @@
+import disableCommandProcess from "./disableCommand";
 import createMiddleware from "./middleware";
+import type { Process } from "./types";
 
 export type { ProcessOutcome, ProcessActions, Process } from "./types";
 
 // Construct the middleware with process handlers.
-export default createMiddleware([]);
+export default createMiddleware([disableCommandProcess as Process<unknown>]);
```
huwshimi, line 11: @andogq I think we might have something wrong in our types if we need to use `as Process<unknown>` here.

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +537,37 @@ function runModelPoller(
...
+              reduxStore.getState(),
+              wsControllerURL,
+            );
+            const modelConnection = await connectToModel(
+              response.uuid,
+              wsControllerURL,
+              credentials,
+            );
+            if (modelConnection) {
+              console.log("here");
```
huwshimi: What's here?????

src/store/middleware/process/disableCommand.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export type Payload = {
+  connection: ConnectionWithFacades;
+  modelUUID: string;
+  wsControllerURL: string;
+  params: BlockSwitchParams;
+};
+
+// eslint-disable-next-line require-yield
+async function* runBlock(
```
huwshimi: @andogq it looks like this is being used a bit differently from how you're using createProcess. Maybe you can comment on this.

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +537,37 @@ function runModelPoller(
...
+              await disableCommandProcess.start(
+                {
+                  connection: modelConnection,
```
huwshimi: I think if you set up the process to use a connecting you won't need to create this connection yourself: e.g. https://github.com/canonical/juju-dashboard/pull/2295/changes#diff-c5ced22e7878455de2f27ec7bb630e59b8b64b8aebb7c246e6517bc7dc59d513R130

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +537,37 @@ function runModelPoller(
...
+              wsControllerURL,
+            );
+            const modelConnection = await connectToModel(
+              response.uuid,
+              wsControllerURL,
+              credentials,
+            );
+            if (modelConnection) {
+              console.log("here");
+              await disableCommandProcess.start(
```
huwshimi: I wonder if this sub-process is necessary if all it's doing is `await payload.connection.facades.block.switchBlockOn(payload.params)`

@andogq what do you think? Should it just be part of this parent call to create the model?

src/store/middleware/process/index.ts
andogq: This is intended for now. `createMiddleware` takes an array of `Process`, but doesn't care what the inner type is. `Process<any>` would also work, but we do not allow explicit `any`.

src/store/middleware/process/disableCommand.ts
andogq: It just comes down to how much feedback we'd like to give to the user. For this method, it's admittedly quite simple, but we could still provide a notification that it's been initiated or that it's pending.

Additionally, if there's nothing useful to return from the process, just return nothing and use `void` as the third generic in `Process` (`Process<Payload, void, void>`). If `switchBlockOn` can resolve to an object that has an error, then check it within the process and throw as necessary, rather than returning an error.

```ts
  if (!payload.connection.facades.block) {
    throw new Error(Label.NO_BLOCK_FACADE_ERROR);
  }
  const request = payload.connection.facades.block.switchBlockOn(payload.params);
  yield { status: "pending" };
  await request;
  return;
```

src/store/middleware/model-poller.ts
andogq: If it's truely only a single call, then it could be just called directly. But if it requires additional logic (error checking, verifying parameters/return value, etc), then I'd say put it in a process to keep it contained.

Additionally, do not invoke processes directly like this. You should dispatch the `run` action ([example](https://github.com/andogq/juju-dashboard/blob/feat/juju-9503-upgrade-to/src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx#L95-L104)):

```ts
import { disableCommand } from "store/middleware/process";

// Middleware returns a promise that will resolve when the process is complete.
await dispatch(disableCommand.run(...));
```

src/store/middleware/model-poller.ts
Ninfa-Jeon: oops

src/store/middleware/model-poller.ts
Ninfa-Jeon: We will definitely be dealing with the error case in future. Lukas needs to come up with something for that.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/store/middleware/model-poller.ts

src/store/middleware/model-poller.ts

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: oh right...thanks for this suggestion!

src/store/middleware/process/disableCommand.ts
Ninfa-Jeon: We will not be providing the user with feedbacks for initiation or pending states of the block. That is because we are making this to look seamless with the add-model. We will be handling the error states in the future.

Ninfa-Jeon: keeping the above in mind, is there anything that needs to be done here?

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Done.

src/store/middleware/connection/middleware.ts
```diff
@@ -53,8 +53,21 @@ export function createConnectionMiddleware(): {
...
+      }
+
+      const modelURLMatch = connectionURL.match(/^(.*)\/model\/[^/]+\/api\/?$/);
+      if (!modelURLMatch) {
+        return undefined;
+      }
+
+      const controllerURL = `${modelURLMatch[1]}/api`;
+      return getUserPass(state, controllerURL);
+    },
```
Ninfa-Jeon, line 63: The process connection middleware asks credentials by connection key and for `modelURL` that key has no stored credentials, so model connection auth was failing intermittently with the block never running. So, I've patched credential lookup to derive controller URL from model URL and reuse controller creds.

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +536,34 @@ function runModelPoller(
...
+                    type: disabledCommands,
+                  },
+                }),
+              ) as unknown as Promise<void>);
+            } catch (blockError) {
+              logger.error(
+                "Could not apply block commands to model.",
+                blockError,
+              );
+            }
```
Ninfa-Jeon: simply logging the error until UX comes up with a graceful way to handle it.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,69 @@
...
+): AsyncGenerator<void, ErrorResult, void> {
+  if (!hasConnections(payload.meta, REQUIRED_CONNECTIONS)) {
+    throw new Error(Label.MISSING_CONNECTION_ERROR);
+  }
+
+  const modelConnection = payload.meta.connections.modelURL;
+
+  if (!modelConnection.facades.block) {
+    throw new Error(Label.NO_BLOCK_FACADE_ERROR);
+  }
```
andogq: Prefer to place the metadata and connection checks in the `createProcess` callback, and save this `runBlock` function for the distilled logic. See [an example](https://github.com/andogq/juju-dashboard/blob/feat/juju-9503-upgrade-to/src/store/middleware/process/model-upgrade/upgrade-to.ts#L127-L140).

In this case, `runBlock` should take `params` and `modelConnection` as two parameters.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,69 @@
...
+  return await modelConnection.facades.block.switchBlockOn(payload.params);
```
andogq: Please see my [previous comment](https://github.com/canonical/juju-dashboard/pull/2308#discussion_r3144635747) about correctly handling the error from the facade, and not directly returning the result.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,69 @@
...
+}
+
+export type Payload = {
+  modelUUID: string;
+  modelURL: string;
+  wsControllerURL: string;
+  params: BlockSwitchParams;
+};
+
+// eslint-disable-next-line require-yield
```
andogq: I'd still recommend including some kind of basic status to this process, even if it's just a simple `initiated` before making the call. Keeps everything in line in case this grows more complex in the future. See [an example](https://github.com/andogq/juju-dashboard/blob/a01f5d07ae83ff6bb253ce975feefad11f71702d/src/store/middleware/process/model-upgrade/upgrade-to.ts#L84-L90).

src/store/middleware/model-poller.ts
andogq: This is where correctly handling the error within the process (dispatching it to a store, [sending a toast](https://github.com/andogq/juju-dashboard/blob/a01f5d07ae83ff6bb253ce975feefad11f71702d/src/store/middleware/process/model-upgrade/upgrade-to.ts#L153-L156), etc).

Notably, the promise returned by the dispatched `run` action will never reject (if it does, this is a bug, lmk and I'll fix it). Instead, it's just a promise that will resolve with `undefined` when the process is complete, so it's only useful for tracking when a process is done, so there's no need for the `try`/`catch`. If you don't need to do this, then don't `await` the dispatch.

andogq (commented):
Not a full review, just looking at the `process`-adjacent things.

src/store/middleware/process/disableCommand.ts
andogq: Yes, see https://github.com/canonical/juju-dashboard/pull/2308#pullrequestreview-4178546870

src/store/middleware/model-poller.ts
Ninfa-Jeon: ah yes, this was throwing an error without the middleware fix that I made above and we were seeing the error toast while the model was still being created. After the fix that I've made, we see success toast for model creation with the logged error. I don't think there will be an issue like this with my fix in?

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: Done

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: Done

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: Done

src/store/middleware/model-poller.ts
Ninfa-Jeon: Done!

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export type Payload = {
+  modelUUID: string;
+  modelURL: string;
+  wsControllerURL: string;
+  params: BlockSwitchParams;
+};
+
+export type BlockStatus = { status: "initiated" } | { status: "pending" };
```
huwshimi:
```suggestion
export type BlockStatus = { status: "initiated" | "pending" };
```

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +536,27 @@ function runModelPoller(
         if (response) {
           if ("error" in response) {
             throw response.error;
+          } else if (
+            disabledCommands !== "none" &&
```
huwshimi:
```suggestion
            disabledCommands !== DisableType.NONE &&
```

src/pages/AddModel/AddModel.test.tsx
```diff
@@ -244,6 +247,7 @@ describe("AddModel page", () => {
       modelName: "my-model",
       userTag: "user-eggman@external",
       wsControllerURL: "wss://controller.example.com",
+      disabledCommands: "none",
```
huwshimi: These tests will be easier to maintain if they use the enums for these values.

src/store/juju/types.ts
```diff
@@ -147,6 +149,7 @@ export type AddModel = {
   credential: string;
   cloudTag: string;
   userTag: string;
+  disabledCommands: string;
```
huwshimi: Is there any reason not to use the enum here?

src/store/middleware/connection/middleware.ts
```diff
@@ -53,8 +53,21 @@ export function createConnectionMiddleware(): {
...
+        return directCredentials;
+      }
+
+      const modelURLMatch = connectionURL.match(/^(.*)\/model\/[^/]+\/api\/?$/);
+      if (!modelURLMatch) {
+        return undefined;
+      }
+
+      const controllerURL = `${modelURLMatch[1]}/api`;
+      return getUserPass(state, controllerURL);
```
huwshimi: We know the credentials will only be available for the controller URL so we could swap the logic around and check for `modelURLMatch` first which will simplify this a lot. Something like this (untested):

```suggestion
      const modelURLMatch = connectionURL.match(/^(.*)\/model\/[^/]+\/api\/?$/);
      return getUserPass(state, modelURLMatch ? `${modelURLMatch[1]}/api` : connectionURL);
```

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        withConnection: true,
+        connectionList: ["wsControllerURL", "modelURL"],
+      },
+    });
+  });
+
+  it("calls switchBlockOn and dispatches success outcome", async () => {
+    const result = {
+      error: null,
+    } as unknown as ErrorResult;
```
huwshimi: It looks like you can avoid the as unknown by setting `error: undefined` or just use an empty object.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!hasConnections(meta, REQUIRED_CONNECTIONS)) {
+      throw new Error(Label.MISSING_CONNECTION_ERROR);
+    }
+
+    const modelConnection = meta.connections.modelURL;
+
+    return yield* runBlock(params, modelConnection);
+  },
+  {
+    setStatus: (payload) => noOp({ wsControllerURL: payload.wsControllerURL }),
```
huwshimi: Don't we need a way to keep track of this call so that the add model action in model-poller knows if this was successful or not?

src/store/middleware/model-poller.ts
```diff
@@ -534,7 +536,27 @@ function runModelPoller(
...
+                params: {
+                  type: disabledCommands,
+                },
+              }),
+            );
           }
+
           reduxStore.dispatch(
             jujuActions.setAddModelResult({
               success: true,
```
huwshimi, line 563: To go with my comment in the disable-commands middleware, how do we know if the above call to disable the commands was successful. I imagine we would want to display an error here instead of success if disabling the commands failed.

It would be a bit of a strange situation because the model would have been created but not successfully set up (same as if the permissions call failed) so we'd need an error message that somehow explained that.

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: No we do not. For two reasons mainly:

1. As said previously, we are not planning to show any feedback of this call to the user currently. The UX will decide about this in the next cycle.
2. This is an independent call which is only chained to addmodel. This should have no affect on the toasts and notifications of addmodel.

src/store/middleware/model-poller.ts
Ninfa-Jeon: Next cycle :)

src/store/juju/types.ts
Ninfa-Jeon: You're right. Switched it out!

src/pages/AddModel/AddModel.tsx
```diff
@@ -168,6 +170,7 @@ const AddModel: FC = () => {
               [ConfigFieldName.CONFIG_INPUT_MODE]: InputMode.LIST,
               [ConfigFieldName.CONFIG_YAML]: "",
               ...getConfigInitialValues(CONFIG_CATEGORIES),
+              disabledCommands: DisableType.NONE,
```
andogq: Should this use `ConfigFieldName`?

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -94,6 +96,35 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                type="radio"
+                name={FieldName.DISABLED_COMMANDS}
+                label={label}
+                value={value}
+              />
+              <div className="p-text--small u-no-margin--bottom">
+                {description}
+                {disabledCommands ? (
+                  <div>
+                    <b>Disables commands:</b> {disabledCommands.join(", ")}
```
andogq, line 119: Should the disabled commands be placed in a code lock/monospace?

src/pages/AddModel/ConfigsConstraints/disabledCommandOptions.ts
```diff
@@ -0,0 +1,84 @@
...
+      "remove-user",
+      "resolved",
+      "retry-provisioning",
+      "run",
+      "scale-application",
+      "set-application-base",
+      "set-credential",
+      "set-constraints",
+      "sync-agents",
+      // eslint-disable-next-line @cspell/spellchecker
```
andogq: It's fine to add this to the cspell config file, since it's a domain specific term

src/pages/AddModel/ConfigsConstraints/types.ts
```diff
@@ -8,14 +8,31 @@ export enum Label {
...
 export enum FieldName {
   CONFIG_INPUT_MODE = "configInputMode",
   CONFIG_YAML = "configYAML",
+  DISABLED_COMMANDS = "disabledCommands",
 }
 
 export type FormFields = {
   [FieldName.CONFIG_INPUT_MODE]: InputMode;
   [FieldName.CONFIG_YAML]: string;
+  disabledCommands: DisableType;
```
andogq: Should this use the `FieldName`?

src/store/juju/slice.ts
```diff
@@ -835,6 +837,31 @@ const slice = createSlice({
       state.addModelState.loading = false;
       state.addModelState.loaded = true;
     },
+    setBlockState: (
```
andogq: Is this needed as a separate action just to add the key with defaults? Would it be better to set it in other reducers when it's needed?

src/store/middleware/connection/middleware.ts
andogq: Interesting, I haven't come across this issue working with upgradeTo and checking fullStatus. does it consistently fail?

src/store/middleware/model-poller.test.ts
```diff
@@ -1524,4 +1531,90 @@ describe("model poller", () => {
...
+    );
+    expect(fakeStore.dispatch).toHaveBeenCalledWith(
+      jujuActions.setAddModelResult({
+        success: true,
+        wsControllerURL: "wss://example.com/api",
+      }),
+    );
+  });
+
+  it("skips disable command process when model connection fails", async () => {
```
andogq: Doesn't look like this test has been implemented

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,85 @@
...
+  params: BlockSwitchParams,
+  modelConnection: ConnectionWithFacades,
+): AsyncGenerator<BlockStatus, ErrorResult, void> {
+  if (!modelConnection.facades.block) {
+    throw new Error(Label.NO_BLOCK_FACADE_ERROR);
+  }
+  const request = modelConnection.facades.block.switchBlockOn(params);
+  yield { status: "pending" };
+  const result = await request;
+  yield { status: "initiated" };
```
andogq: I'd move the initiated after the error, as presumably if there's an error thrown from the facade, then it's not actually initiated.

Additionally, I heard you talking about in the standup about how this is processed in the background. If there was some way to fetch a list of disabled commands from JIMM, this'd be a good place to poll that endpoint, and wait for the commands to be added in order to verify they're done.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (result.error) {
+    throw result.error;
+  }
+  return result;
```
andogq: Don't just return things for the sake of it, if it's not useful. It's fine to just `return`, and update the return type to be `AsyncGenerator<BlockStatus, void, void>`.

src/store/middleware/process/block/disable-command.ts
andogq: Given you already have a slice for this, it wouldn't hurt to chuck the state in there anyway, so we can access it easier in the future when we do get around to it.

src/store/middleware/process/block/disable-command.ts
```diff
@@ -0,0 +1,85 @@
...
+      throw new Error(Label.MISSING_CONNECTION_ERROR);
+    }
+
+    const modelConnection = meta.connections.modelURL;
+
+    return yield* runBlock(params, modelConnection);
+  },
+  {
+    setStatus: (payload) => noOp({ wsControllerURL: payload.wsControllerURL }),
+    setRunning: (payload) =>
```
andogq: `setRunning` takes a second parameter, a boolean which indicates that it's in progress. So this action is called twice, once when it starts (with true), and again when it stops (with false). This means that this part of the slice will instantly be cleared when the result comes in.

I'd update the slice to include a `running` field, and set it to the boolean from this hook.

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,185 @@
...
+  let dispatch: Store["dispatch"];
+  let params: BlockSwitchParams;
+  let payload: { meta: Record<string, unknown> } & Payload;
+  let modelConnection: Connection;
+
+  beforeEach(() => {
+    dispatch = vi.fn();
+    params = { type: "BlockChange" };
+    // @ts-expect-error - Connection mocked in `src/testing/setup.ts`.
+    modelConnection = new Connection();
```
andogq: `{} as ManagedConnection` should suffice here.

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,185 @@
...
+    modelConnection.facades = {
+      block: {
+        switchBlockOn: vi.fn(),
+      },
+    };
+    modelConnection.transport = {} as Transport;
```
andogq: This also shouldn't be needed.

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    payload = {
+      modelUUID: "model-uuid-123",
+      modelURL: "wss://example.com/model/model-uuid-123/api",
+      wsControllerURL: "wss://example.com",
+      params,
+      meta: {
+        connections: {
+          modelURL: modelConnection,
+          wsControllerURL: modelConnection,
```
andogq: I'd set up a separate connection mock for the controller.

andogq (changes requested):
Looks good overall. Some changes in the tests are still needed, but given you get that implemented I'm happy with it!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: We don't want the commands list to look any different from the text. Do you suggest I use code anyway and apply custom styles?

src/store/juju/slice.ts
Ninfa-Jeon: Not needed. Good catch.

src/store/middleware/connection/middleware.ts
Ninfa-Jeon: Yes it was happening every time. I am guessing thats because a new model was being created.

src/store/middleware/model-poller.test.ts
Ninfa-Jeon: It's impressive how you figured that out. Must've gotten lost in the endless merge and rebase cycles.

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: The latter cannot be done because the command disabling is not done as a list but as a group. Which means for example, that when I disable commands in group A, a whole set gets blocked. I don't think there is a utility for fetching this list at the moment. Could be requested in the future

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: umm...wouldn't we need the ErrorResult to populate the `outcome` in case this fails? I know we are not handling the failure scenario today but it is in near future that we would have to

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: done. it took me a while to figure out why it wouldnt work

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: done

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: done

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: done

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: nvm that previous comment, i got it

src/store/middleware/process/block/disable-command.ts
Ninfa-Jeon: done

src/store/middleware/model-poller.test.ts
Ninfa-Jeon: done

src/store/juju/slice.ts
Ninfa-Jeon: removed.

src/pages/AddModel/ConfigsConstraints/types.ts
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/disabledCommandOptions.ts
Ninfa-Jeon: done

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: done.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
andogq: If it's meant to be non-monospace, then that's fine then!

src/store/middleware/model-poller.ts
```diff
@@ -546,16 +546,21 @@ function runModelPoller(
...
-                wsControllerURL,
-                params: {
-                  type: disabledCommands,
-                },
-              }),
-            );
+            const modelConnection = await connections
+              .get(modelURL)
+              .catch((_err) => undefined);
+            if (modelConnection) {
```
andogq: What's the purpose of creating the connection and using this if statement? It shouldn't be required.

src/store/middleware/process/block/disable-command.test.ts
andogq, line 1: Apologies that I didn't fully pick up on this initially, but similar to the `source` implementations, there isn't any need to create tests to ensure that `createProcess` works (that the actions it creates are correct, that it calls the implementation, etc).

Instead, I'd only test the `runBlock` function that you've created, since that contains the important logic. Everything else is more-so boilerplate code that's tested elsewhere.

I'd have the following test scenarios:

- Mock `switchBlockOn` function return successfully, and call `runBlock`. Ensure that the generator yields `pending` and `initiated`.
- Mock `switchBlockOn` to return the `error` field, and call `runBlock`. Ensure that the generator yields `pending`, then throws the mocked error.
- Call `runBlock` with a connection that doesn't have the `block` facade. Ensure it throws an error.
- Call `disableCommand.actions.run` and ensure that the action has the meta fields (similar to what you have here).

I've left more specific comments below.

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,202 @@
...
+import type { ManagedConnection } from "store/middleware/connection/connection-manager";
+import type { Store } from "store/store";
+
+import disableCommand, { Label, type Payload } from "./disable-command";
+
+describe("disableCommand process", () => {
+  let dispatch: Store["dispatch"];
+  let params: BlockSwitchParams;
+  let payload: { meta: Record<string, unknown> } & Payload;
+  let modelConnection: Connection;
```
andogq: Mocked values should be placed in a scope they can be used, with the correct types.

```suggestion
  let modelConnection: ManagedConnection;
  let controllerConnection: ManagedConnection;
```

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    params = { type: "BlockChange" };
+    modelConnection = {} as ManagedConnection;
+    const controllerConnection = {} as ManagedConnection;
+    Object.setPrototypeOf(modelConnection, Connection.prototype);
+    Object.setPrototypeOf(controllerConnection, Connection.prototype);
+    modelConnection.facades = {
+      block: {
+        switchBlockOn: vi.fn(),
+      },
+    };
```
andogq: Don't use prototype hacks here, it's not required.

```suggestion
    modelConnection = {
      facades: {
        block: {
          switchBlockOn: vi.fn()
        }
      }
    } as ManagedConnection;
    controllerConnection = {} as ManagedConnection;
```

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      modelURL: "wss://example.com/model/model-uuid-123/api",
+      wsControllerURL: "wss://example.com",
+      params,
+      meta: {
+        connections: {
+          modelURL: modelConnection,
+          wsControllerURL: controllerConnection,
+        },
+      },
+    };
```
andogq: I'm unsure what the purpose of this is, but it's not required.

src/store/middleware/process/block/disable-command.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      error: undefined,
+    };
+    const switchBlockOn = vi.fn().mockResolvedValue(result);
+    modelConnection.facades = {
+      block: {
+        switchBlockOn,
+      },
+    };
+
+    await disableCommand.start(payload, dispatch);
```
andogq: Instead of this test, just call `runBlock` directly and assert what's yielded out of it, instead of mocking the `dispatch` function and checking what was passed to it.

andogq (commented):
This is getting very close, just some simplification in the tests to make, and a question about the model connection. Happy to have a chat about it when you get online if you need a hand with it.

src/store/middleware/model-poller.ts
Ninfa-Jeon: it was easier to test this way but you're right..ill find another way to assert

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: I suppose this wouldn't be needed with the test file rewrite

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: Done

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: Removed

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: Done

src/store/middleware/process/block/disable-command.test.ts
Ninfa-Jeon: Thank you for checking! I have made the necessary changes.

src/store/middleware/model-poller.ts

andogq (approved):
Nice job with those changes, looks great now! :shipit:

andogq:
image
Seems to be some spacing issues on text in firefox, the label text has `-0.5rem` applied to it.

andogq:
The issue with the form fields is also present in `main`, so I'll +1 this for QA

PR #2311, opened 2026-04-27
feat: submit changed configs to facade method

Ninfa-Jeon:
## Done

- Added search bar to configs list allowing filtering based on config name and description
- Added `useDebounce` hook to optimise search
- Added tooltip to `changed-configs-only` toggle for when no configs have been changed from default
- Search works for all as well as changed configs
- Added `config` key to `createModel` payload

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Try changing the `changed-configs-only` toggle without any changes, it should stay disabled with a tooltip message
- Try doing the same after changing something, it should get enabled and work as expected
- Try searching for configs based on config name and description
- Search should show a no results found message when no results match the query
- Change some configs and create a model
- Verify that the model was created with this changed value by either monitoring the websocket or simply running `juju model-config` within the model context

## Details

https://warthogs.atlassian.net/browse/JUJU-9680

## Screenshots

Screenshot 2026-04-27 at 4 43 52 PM

### DIFF
--- a/src/hooks/useDebounce.test.ts
+++ b/src/hooks/useDebounce.test.ts
@@ -1,0 +1,74 @@
+import { act, renderHook } from "@testing-library/react";
+import { vi } from "vitest";
+
+import useDebounce from "./useDebounce";
+
+describe("useDebounce", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("returns the initial value immediately", () => {
+    const { result } = renderHook(() => useDebounce("hello", 250));
+    expect(result.current).toBe("hello");
+  });
+
+  it("debounces updates", () => {
+    const { result, rerender } = renderHook(
+      ({ value, delay }: { value: string; delay: number }) =>
+        useDebounce(value, delay),
+      { initialProps: { value: "hello", delay: 250 } },
+    );
+
+    expect(result.current).toBe("hello");
+
+    rerender({ value: "world", delay: 250 });
+    expect(result.current).toBe("hello");
+
+    act(() => {
+      vi.advanceTimersByTime(250);
+    });
+    expect(result.current).toBe("world");
+  });
+
+  it("cancels pending updates on unmount", () => {
+    const { result, rerender, unmount } = renderHook(
+      ({ value, delay }: { value: string; delay: number }) =>
+        useDebounce(value, delay),
+      { initialProps: { value: "hello", delay: 250 } },
+    );
+
+    rerender({ value: "world", delay: 250 });
+    unmount();
+
+    act(() => {
+      vi.advanceTimersByTime(250);
+    });
+    expect(result.current).toBe("hello");
+  });
+
+  it("respects custom delay", () => {
+    const { result, rerender } = renderHook(
+      ({ value, delay }: { value: string; delay: number }) =>
+        useDebounce(value, delay),
+      { initialProps: { value: "hello", delay: 100 } },
+    );
+
+    rerender({ value: "world", delay: 100 });
+    expect(result.current).toBe("hello");
+
+    act(() => {
+      vi.advanceTimersByTime(50);
+    });
+    expect(result.current).toBe("hello");
+
+    act(() => {
+      vi.advanceTimersByTime(50);
+    });
+    expect(result.current).toBe("world");
+  });
+});
--- a/src/hooks/useDebounce.test.ts
+++ b/src/hooks/useDebounce.test.ts
@@ -14,61 +14,119 @@ describe("useDebounce", () => {
 
   it("returns the initial value immediately", () => {
     const { result } = renderHook(() => useDebounce("hello", 250));
-    expect(result.current).toBe("hello");
+    expect(result.current[0]).toBe("hello");
   });
 
   it("debounces updates", () => {
-    const { result, rerender } = renderHook(
-      ({ value, delay }: { value: string; delay: number }) =>
-        useDebounce(value, delay),
-      { initialProps: { value: "hello", delay: 250 } },
-    );
+    const { result } = renderHook(() => useDebounce("hello", 250));
 
-    expect(result.current).toBe("hello");
+    expect(result.current[0]).toBe("hello");
 
-    rerender({ value: "world", delay: 250 });
-    expect(result.current).toBe("hello");
+    act(() => {
+      result.current[1]("world");
+    });
+    expect(result.current[0]).toBe("hello");
 
     act(() => {
       vi.advanceTimersByTime(250);
     });
-    expect(result.current).toBe("world");
+    expect(result.current[0]).toBe("world");
+  });
+
+  it("only applies the last value when updates happen during the delay", () => {
+    const { result } = renderHook(() => useDebounce("hello", 250));
+
+    act(() => {
+      result.current[1]("first");
+    });
+
+    act(() => {
+      vi.advanceTimersByTime(100);
+    });
+
+    act(() => {
+      result.current[1]("second");
+    });
+
+    act(() => {
+      vi.advanceTimersByTime(249);
+    });
+    expect(result.current[0]).toBe("hello");
+
+    act(() => {
+      vi.advanceTimersByTime(1);
+    });
+    expect(result.current[0]).toBe("second");
   });
 
   it("cancels pending updates on unmount", () => {
-    const { result, rerender, unmount } = renderHook(
-      ({ value, delay }: { value: string; delay: number }) =>
-        useDebounce(value, delay),
-      { initialProps: { value: "hello", delay: 250 } },
-    );
+    const { result, unmount } = renderHook(() => useDebounce("hello", 250));
 
-    rerender({ value: "world", delay: 250 });
+    act(() => {
+      result.current[1]("world");
+    });
     unmount();
 
     act(() => {
       vi.advanceTimersByTime(250);
     });
-    expect(result.current).toBe("hello");
+    expect(result.current[0]).toBe("hello");
   });
 
   it("respects custom delay", () => {
-    const { result, rerender } = renderHook(
-      ({ value, delay }: { value: string; delay: number }) =>
-        useDebounce(value, delay),
-      { initialProps: { value: "hello", delay: 100 } },
-    );
+    const { result } = renderHook(() => useDebounce("hello", 100));
 
-    rerender({ value: "world", delay: 100 });
-    expect(result.current).toBe("hello");
+    act(() => {
+      result.current[1]("world");
+    });
+    expect(result.current[0]).toBe("hello");
 
     act(() => {
       vi.advanceTimersByTime(50);
     });
-    expect(result.current).toBe("hello");
+    expect(result.current[0]).toBe("hello");
 
     act(() => {
       vi.advanceTimersByTime(50);
     });
-    expect(result.current).toBe("world");
+    expect(result.current[0]).toBe("world");
+  });
+
+  it("updates immediately and clears pending debounced updates", () => {
+    const { result } = renderHook(() => useDebounce("hello", 250));
+
+    act(() => {
+      result.current[1]("delayed");
+    });
+
+    act(() => {
+      result.current[1]("", { immediate: true });
+    });
+    expect(result.current[0]).toBe("");
+
+    act(() => {
+      vi.advanceTimersByTime(250);
+    });
+    expect(result.current[0]).toBe("");
+  });
+
+  it("supports function values without invoking them", () => {
+    const initialFn = vi.fn(() => "initial");
+    const nextFn = vi.fn(() => "next");
+    const { result } = renderHook(() => useDebounce(initialFn, 250));
+
+    expect(result.current[0]).toBe(initialFn);
+
+    act(() => {
+      result.current[1](nextFn);
+    });
+    expect(result.current[0]).toBe(initialFn);
+
+    act(() => {
+      vi.advanceTimersByTime(250);
+    });
+    expect(result.current[0]).toBe(nextFn);
+    expect(initialFn).not.toHaveBeenCalled();
+    expect(nextFn).not.toHaveBeenCalled();
   });
 });
--- a/src/hooks/useDebounce.ts
+++ b/src/hooks/useDebounce.ts
@@ -1,22 +1,50 @@
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
 
 /**
- * Hook that debounces a value.
- * @param value - The value to debounce
+ * Hook that debounces state updates.
+ * @param initialValue - The initial value
  * @param delay - The debounce delay in milliseconds
- * @returns The debounced value
+ * @returns Tuple of debounced value and setter
  */
-export default function useDebounce<T>(value: T, delay: number): T {
-  const [debouncedValue, setDebouncedValue] = useState(value);
+export default function useDebounce<T>(
+  initialValue: T,
+  delay: number,
+): [T, (value: T, options?: { immediate?: boolean }) => void] {
+  const [debouncedValue, setDebouncedValue] = useState(() => initialValue);
+  const pendingValueRef = useRef(initialValue);
+  const timeoutRef = useRef<null | ReturnType<typeof setTimeout>>(null);
+
+  const clearPendingTimeout = useCallback((): void => {
+    if (timeoutRef.current !== null && timeoutRef.current !== undefined) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
+  }, []);
+
+  const setValue = useCallback(
+    (value: T, options?: { immediate?: boolean }): void => {
+      pendingValueRef.current = value;
+
+      clearPendingTimeout();
+
+      if (options?.immediate) {
+        setDebouncedValue(() => value);
+        return;
+      }
+
+      timeoutRef.current = setTimeout(() => {
+        setDebouncedValue(() => pendingValueRef.current);
+        timeoutRef.current = null;
+      }, delay);
+    },
+    [clearPendingTimeout, delay],
+  );
 
   useEffect(() => {
-    const id = setTimeout(() => {
-      setDebouncedValue(value);
-    }, delay);
     return (): void => {
-      clearTimeout(id);
+      clearPendingTimeout();
     };
-  }, [value, delay]);
+  }, [clearPendingTimeout]);
 
-  return debouncedValue;
+  return [debouncedValue, setValue];
 }
--- a/src/hooks/useDebounce.ts
+++ b/src/hooks/useDebounce.ts
@@ -1,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+/**
+ * Hook that debounces a value.
+ * @param value - The value to debounce
+ * @param delay - The debounce delay in milliseconds
+ * @returns The debounced value
+ */
+export default function useDebounce<T>(value: T, delay: number): T {
+  const [debouncedValue, setDebouncedValue] = useState(value);
+
+  useEffect(() => {
+    const id = setTimeout(() => {
+      setDebouncedValue(value);
+    }, delay);
+    return (): void => {
+      clearTimeout(id);
+    };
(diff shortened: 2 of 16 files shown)


src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                />
+              </div>
+            </div>
+            <ChangedOnlySwitchWrapper>
+              <Switch
+                label={Label.CHANGED_CONFIGS_ONLY}
+                checked={changedOnly}
+                onChange={(event: ChangeEvent<HTMLInputElement>): void => {
+                  if (!hasChangedConfigs && event.target.checked) {
+                    return;
```
ale8k: I think this might have a ux inconsistency, if the switch is still interactive and we silently ignore no config changed, wouldn't this just allow them to click the switch and nothing will happen?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -79,3 +84,18 @@ export const getConfigInitialValues = (
...
+export const filterConfigsBySearch = (query: string): CategoryDefinition[] => {
+  if (!query.trim()) {
+    return CONFIG_CATEGORIES;
+  }
+  const lowerQuery = query.toLowerCase();
+  return CONFIG_CATEGORIES.map((category) => ({
+    ...category,
+    fields: category.fields.filter(
+      (field) =>
+        field.label.toLowerCase().includes(lowerQuery) ||
```
ale8k, line 97: Should we trim too, so that "my-config-option-with-a-space " works?

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
-            />
+            <div className="row u-no-padding">
+              <div className="col-4">
+                <SearchBox
+                  externallyControlled
+                  name="configSearch"
+                  id="configSearch"
+                  placeholder="Search configurations"
+                  value={searchQuery}
+                  onChange={setSearchQuery}
```
ale8k, line 93: Maybe a debounce?

ale8k (commented):
Minor comments just some fiddly ux bits, won't approve because I'm not confident in this codebase but lgtm otherwise!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                  onClear={() => {
+                    setSearchQuery("");
+                  }}
+                  className="configs__search-input"
+                  aria-label="Search configurations"
+                />
+              </div>
+            </div>
+            <ChangedOnlySwitchWrapper>
```
ale8k: Oh nvm, we don't have a switch right because it is changed only. Sorry aha!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+  }: {
+    children: JSX.Element;
+  }): JSX.Element =>
+    hasChangedConfigs ? (
+      <>{children}</>
+    ) : (
+      <Tooltip message={Label.NO_CHANGED_CONFIGS} position="btm-left">
+        {children}
+      </Tooltip>
+    );
```
huwshimi: It's a bit unconventional to define a nested component here. Instead, I'd probably reverse the logic and do:

```suggestion
  const changedOnlySwitch = (
    <Switch
      label={Label.CHANGED_CONFIGS_ONLY}
      checked={changedOnly}
      onChange={(event: ChangeEvent<HTMLInputElement>): void => {
        if (!hasChangedConfigs && event.target.checked) {
          return;
        }
        setChangedOnly(event.target.checked);
      }}
    />
  );
```

And below where you currently do `<ChangedOnlySwitchWrapper>...` you could put the ternary:
```jsx
{hasChangedConfigs ? (
    changedOnlySwitch
  ) : (
    <Tooltip message={Label.NO_CHANGED_CONFIGS} position="btm-left">
      {changedOnlySwitch}
    </Tooltip>
  )}
```

src/pages/AddModel/_add-model.scss
```diff
@@ -23,6 +23,7 @@
     padding: $spv--large $sph--small 0;
     position: sticky;
     text-align: right;
+    z-index: 2;
```
huwshimi: It would probably be a good idea to use the z index util and leave a comment for why this z-index is required in case it breaks in the future. Like this (minus the !important): https://github.com/canonical/juju-dashboard/blob/92ce07fbd78614668fb447b73a37b100d61bc45d/src/components/ToastCard/_toast-card.scss#L3

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                  aria-label="Search configurations"
+                />
+              </div>
+            </div>
+            <ChangedOnlySwitchWrapper>
+              <Switch
+                label={Label.CHANGED_CONFIGS_ONLY}
+                checked={changedOnly}
+                onChange={(event: ChangeEvent<HTMLInputElement>): void => {
+                  if (!hasChangedConfigs && event.target.checked) {
```
huwshimi: Rather than the input not responding it might be better to disable the switch if there are no unchanged values (though it would need to be enabled if it's toggled on so it can be toggled off).

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                  externallyControlled
+                  name="configSearch"
+                  id="configSearch"
+                  placeholder="Search configurations"
+                  value={searchQuery}
+                  onChange={setSearchQuery}
+                  onClear={() => {
+                    setSearchQuery("");
+                  }}
+                  className="configs__search-input"
```
huwshimi: To achieve the tighter spacing in the design you can just do:

```suggestion
                  className="u-no-margin--bottom"
```

src/pages/AddModel/ConfigsConstraints/_configs-constraints.scss
```diff
@@ -1,22 +1,33 @@
...
+  }
+
+  .configs__search-input {
+    margin-bottom: $sph--small;
+  }
+
+  .p-switch {
+    margin-bottom: 0;
+    padding-bottom: $sph--small;
+  }
```
huwshimi: I've noted on the component how to do this the Vanilla way. The design also has tighter spacing:

Image

```suggestion
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                onChange={(event: ChangeEvent<HTMLInputElement>): void => {
+                  if (!hasChangedConfigs && event.target.checked) {
+                    return;
+                  }
+
+                  setChangedOnly(event.target.checked);
+                }}
+              />
+            </ChangedOnlySwitchWrapper>
             <div className="configs__form-scroll">
```
huwshimi: As the chip doesn't support applying classes to the wrapping element we can apply a negative margin here:

```suggestion
            <div className="configs__form-scroll u-sv-1--top">
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -41,41 +49,93 @@ const ConfigsConstraints = (): JSX.Element => {
...
+                  className="configs__search-input"
+                  aria-label="Search configurations"
+                />
+              </div>
+            </div>
+            <ChangedOnlySwitchWrapper>
+              <Switch
+                label={Label.CHANGED_CONFIGS_ONLY}
+                checked={changedOnly}
+                onChange={(event: ChangeEvent<HTMLInputElement>): void => {
```
huwshimi: For some reason I have to click this switch twice to make it work after I edit a config:
- Start with the switch off.
- Edit a config field.
- Click the switch and notice that nothing happens.
- Click it again and it will switch on.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -79,3 +84,18 @@ export const getConfigInitialValues = (
...
+        field.description.toLowerCase().includes(lowerQuery),
+    ),
+  })).filter(({ fields }) => fields.length > 0);
```
huwshimi: You can use the same Boolean method you used for the other util:

```suggestion
  })).filter(Boolean);
```

huwshimi (commented):
There are two console errors when going to the config and constraints step:

```
Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both).

Input contains an input of type text with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both).
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: We show them a tooltip in this case :)

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: smart, yes

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: We do have a switch but with proper feedback to the user

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: I SWEAR i reversed this yesterday

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: </3

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Interesting. Will take a look!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: done :)

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: I don't see it after my latest fix. Can you please check again and confirm if it still exists?

src/pages/AddModel/ConfigsConstraints/utils.ts

Ninfa-Jeon:
> There are two console errors when going to the config and constraints step:
> 
> ```
> Select elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both).
> 
> Input contains an input of type text with both value and defaultValue props. Input elements must be either controlled or uncontrolled (specify either the value prop, or the defaultValue prop, but not both).
> ```

~~I also don't see this. Can you confirm if this still exists and if yes, please give me details of your setup~~
My bad, there was a lingering default initialisation in one of the files. I have corrected it in my disable-commands PR. Stepping on my own toes here XD

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: Had to undo this change as it was making the search behave in an undesired manner. The problem is in that `filterConfigsBySearch` needs to filter out categories with empty field arrays. The `.filter(Boolean)` at the end doesn't work because objects are always truthy which is why whenever the search over config name and descriptions returned no results, all config categories were still being rendered with no children.

Ninfa-Jeon: Expected:
Screenshot 2026-04-28 at 12 39 45 PM

With this change:
Screenshot 2026-04-28 at 12 39 29 PM

huwshimi: Sorry, yes, ignore that comment, I clearly wasn't paying attention :).

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
huwshimi: Yep this seems to be fixed now!

src/hooks/useDebounce.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("returns the initial value immediately", () => {
+    const { result } = renderHook(() => useDebounce("hello", 250));
+    expect(result.current).toBe("hello");
+  });
+
+  it("debounces updates", () => {
```
huwshimi, line 20: In this test or a second one it should probably also check that if there are multiple new values during the delay period that only the last value gets set.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -16,14 +20,20 @@ import { InputMode } from "./ContentSwitcher/types";
...
+} from "./utils";
 
 const ConfigsConstraints = (): JSX.Element => {
   const { values, setFieldValue } = useFormikContext<
     FormFields & Record<string, string>
   >();
 
   const [changedOnly, setChangedOnly] = useState(false);
+  const [rawSearchQuery, setRawSearchQuery] = useState("");
+  const searchQuery = useDebounce(rawSearchQuery, 250);
```
huwshimi: This is going to cause the component to update on each value change even though the searchQuery hasn't changed.

Instead, it would be better to have useDebounce return a function that is called to update the state inside the hook, then you could do something like:

```suggestion
  const [searchQuery, setSearchQuery] = useDebounce("", 250);
```

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,22 @@
+import { useEffect, useState } from "react";
+
+/**
+ * Hook that debounces a value.
+ * @param value - The value to debounce
+ * @param delay - The debounce delay in milliseconds
+ * @returns The debounced value
+ */
+export default function useDebounce<T>(value: T, delay: number): T {
```
huwshimi: This is going to delay all updates to the value (except for the inital value when the component is mounted). This also needs a way to immediately update the value e.g. when clearing the input (when it does this it should also clear any pending timers).

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -55,27 +83,51 @@ const ConfigsConstraints = (): JSX.Element => {
...
-              ))}
+            <div className="row u-no-padding">
+              <div className="col-4">
+                <SearchBox
+                  name="configSearch"
+                  id="configSearch"
+                  placeholder="Search configurations"
+                  onChange={setRawSearchQuery}
+                  onClear={() => {
+                    setRawSearchQuery("");
```
huwshimi: With the change above to return an update function from the hook, this will need to call that function but also pass a param to skip the debounce delay as we would want the UI to immediately update, something like this (in pseudo code):


```suggestion
                    setSearchQuery("", immediate=true);
```

src/pages/AddModel/_add-model.scss
```diff
@@ -23,6 +23,7 @@
     padding: $spv--large $sph--small 0;
     position: sticky;
     text-align: right;
+    z-index: z("gamma");
```
huwshimi, line 27: It would be good to add a comment about why this is needed so we know what to fix this if it breaks.

src/pages/AddModel/ConfigsConstraints/utils.ts
huwshimi: Looks like this still needs to be done.

huwshimi (commented):
A few small things left for this, mostly around the debounce.

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: oh, I have added a trim to `lowerQuery` a few lines above this @huwshimi

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Thanks for confirming!

src/pages/AddModel/ConfigsConstraints/utils.ts
huwshimi: Sorry, I see now, thanks :)

src/pages/AddModel/_add-model.scss
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: done

src/hooks/useDebounce.ts
Ninfa-Jeon: You're right. Fixed.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx

src/hooks/useDebounce.test.ts
Ninfa-Jeon: Added

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (timeoutRef.current) {
+      clearTimeout(timeoutRef.current);
+      timeoutRef.current = null;
+    }
+  }, []);
+
+  const resolveNextValue = useCallback(
+    (value: SetStateAction<T>): T =>
+      typeof value === "function"
+        ? (value as (value: T) => T)(pendingValueRef.current)
```
huwshimi: The problem is that type of `T` can also be a function so it can't be assumed that it should resolve the value.

I can't think of a simple solution to fix that, so it might be best to just support `T` without the set state function (it's not going to behave like the actual set state function anyway)

src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,65 @@
...
+export default function useDebounce<T>(
+  initialValue: T,
+  delay: number,
+): [T, (value: SetStateAction<T>, options?: { immediate?: boolean }) => void] {
+  const [debouncedValue, setDebouncedValue] = useState(initialValue);
+  const pendingValueRef = useRef(initialValue);
+  const timeoutRef = useRef<null | ReturnType<typeof setTimeout>>(null);
+
+  const clearPendingTimeout = useCallback((): void => {
+    if (timeoutRef.current) {
```
huwshimi: `timeoutRef` is a number so this probably also needs a check that it's not undefined/null.

huwshimi (approved):
Two tiny fixes then this can land!

src/hooks/useDebounce.ts
Ninfa-Jeon: done

src/hooks/useDebounce.ts
Ninfa-Jeon: done

PR #2312, opened 2026-04-29
feat(process): create `upgrade-to` process

andogq:
## Done

- Create `upgrade-to` process for use with JIMM
- Hook up process to upgrade form

## QA

- Open model upgrade panel
- Select an upgrade which requires a migration
- Click upgrade
- Look for:
  - Pending/initiated notifications
  - Periodic loading notifications
  - Notification that the model is reconnecting
  - Success notification on completion
- Model list should immediately update to reflect new versions when complete

Note: If an error occurs during the upgrade, it will not be displayed in any way, and the upgrade will continue to seem like it's loading. Additionally, if the dashboard is reloaded during an upgrade, the upgrade state will be lost.

Closes JUJU-9503

### DIFF
--- a/src/components/ToastCard/_toast-card.scss
+++ b/src/components/ToastCard/_toast-card.scss
@@ -45,7 +45,7 @@
     border-left: 3px solid $color-negative;
   }
 
-  &[data-type="info"] {
+  &[data-type="information"] {
     border-left: 3px solid $color-information;
   }
 
--- a/src/juju/jimm/JIMMV4.ts
+++ b/src/juju/jimm/JIMMV4.ts
@@ -193,6 +193,32 @@ class JIMMV4 extends JIMMV3 {
       this._transport.write(req, resolve, reject);
     });
   }
+
+  /**
+   * Initiate a model upgrade.
+   *
+   * @param modelTag Tag of the model which will be upgraded.
+   * @param targetController Name of controller to migrate the model to.
+   *
+   * @see {@link https://github.com/canonical/jimm/blob/53c606aa41ae6339c7b6cf430c7051bf489b9238/internal/jujuapi/jimm.go#L738}
+   */
+  async upgradeTo(
+    modelTag: string,
+    targetController: string,
+  ): Promise<boolean | undefined> {
+    return new Promise((resolve, reject) => {
+      const req = {
+        type: "JIMM",
+        request: "UpgradeTo",
+        version: 4,
+        params: {
+          "model-tag": modelTag,
+          "target-controller-name": targetController,
+        },
+      };
+      this._transport.write(req, resolve, reject);
+    });
+  }
 }
 
 JIMMV4.NAME = "JIMM";
--- a/src/juju/jimm/api.ts
+++ b/src/juju/jimm/api.ts
@@ -125,3 +125,17 @@ export const supportedJujuVersions = async (
     jimm.supportedJujuVersions(minVersion),
   );
 };
+
+/*
+ * Initiate a model upgrade.
+ */
+export const upgradeTo = async (
+  conn: ConnectionWithFacades,
+  modelTag: string,
+  targetController: string,
+): Promise<boolean> => {
+  return withJimmCheck(
+    conn,
+    async (jimm) => (await jimm.upgradeTo(modelTag, targetController)) ?? false,
+  );
+};
--- a/src/panels/UpgradeModelPanel/UpgradeModelController/Fields/Fields.tsx
+++ b/src/panels/UpgradeModelPanel/UpgradeModelController/Fields/Fields.tsx
@@ -127,31 +127,40 @@ const Fields: FC<Props> = ({
                         void setFieldValue(FieldName.TARGET_CONTROLLER, value);
                       }}
                       options={[
-                        ...(targetControllers?.map((controller) => {
-                          const optionVersion =
-                            getControllerVersion(controller);
-                          return {
-                            label: (
-                              <span className="u-flex">
-                                <span className="u-flex-grow">
-                                  {controller.name}
+                        ...(targetControllers
+                          ?.filter(
+                            (
+                              controller,
+                            ): controller is {
+                              name: string;
+                            } & typeof controller =>
+                              controller.name !== undefined,
+                          )
+                          .map((controller) => {
+                            const optionVersion =
+                              getControllerVersion(controller);
+                            return {
+                              label: (
+                                <span className="u-flex">
+                                  <span className="u-flex-grow">
+                                    {controller.name}
+                                  </span>
+                                  <span>
+                                    {optionVersion ? (
+                                      <Chip
+                                        isReadOnly
+                                        isDense
+                                        isInline
+                                        value={optionVersion}
+                                      />
+                                    ) : null}
+                                  </span>
                                 </span>
-                                <span>
-                                  {optionVersion ? (
-                                    <Chip
-                                      isReadOnly
-                                      isDense
-                                      isInline
-                                      value={optionVersion}
-                                    />
-                                  ) : null}
-                                </span>
-                              </span>
-                            ),
-                            text: controller.name,
-                            value: controller.uuid,
-                          };
-                        }) ?? []),
+                              ),
+                              text: controller.name,
+                              value: controller.name,
+                            };
+                          }) ?? []),
                       ]}
                     />
                   </>
--- a/src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
+++ b/src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
@@ -73,8 +73,6 @@ const UpgradeModelController: FC<Props> = ({
   });
   function triggerMigration(values: FormFields): void {
     const currentVersion = model?.info?.["agent-version"];
-
-    // TODO: Properly handle
     if (!wsControllerURL) {
       throw new Error("wsControllerURL is required");
     }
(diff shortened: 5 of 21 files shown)


src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
```diff
@@ -67,6 +71,31 @@ const UpgradeModelController: FC<Props> = ({
...
+    // TODO: Properly handle
+    if (!wsControllerURL) {
+      throw new Error("wsControllerURL is required");
+    }
+    if (!modelUUID) {
+      throw new Error("modelUUID is missing");
+    }
+    if (!currentVersion) {
+      throw new Error("currentVersion is missing");
+    }
```
andogq: I'm assuming it's safe to rely on yup for handling these?

src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
```diff
@@ -67,6 +71,31 @@ const UpgradeModelController: FC<Props> = ({
...
+        targetVersion: version.version,
+        targetController: values[FieldName.TARGET_CONTROLLER],
+        currentVersion,
+        wsControllerURL,
+        modelUUID,
+        modelURL: wsControllerURL.replace("/api", `/model/${modelUUID}/api`),
+        modelName: model.info?.name ?? "Unknown model"
+      });
+      dispatch(action);
+    }
```
andogq, line 99: @huwshimi This is where the other migration would be triggered instead.

src/juju/jimm/api.ts
```diff
@@ -125,3 +125,17 @@ export const supportedJujuVersions = async (
     jimm.supportedJujuVersions(minVersion),
   );
 };
+
+/*
+ * Initiate a model upgrade.
+ */
+export const upgradeTo = async (
```
huwshimi, line 132: It looks like this needs a test.

src/juju/jimm/JIMMV4.ts
```diff
@@ -193,6 +193,32 @@ class JIMMV4 extends JIMMV3 {
...
+
+  /**
+   * Initiate a model upgrade.
+   *
+   * @param modelTag Tag of the model which will be upgraded.
+   * @param targetController Name of controller to migrate the model to.
+   *
+   * @see {@link https://github.com/canonical/jimm/blob/53c606aa41ae6339c7b6cf430c7051bf489b9238/internal/jujuapi/jimm.go#L738}
+   */
+  async upgradeTo(
```
huwshimi, line 205: This also needs a test.

src/panels/UpgradeModelPanel/UpgradeModelController/Fields/Fields.tsx
```diff
@@ -127,31 +127,40 @@ const Fields: FC<Props> = ({
...
-                                  {controller.name}
+                        ...(targetControllers
+                          ?.filter(
+                            (
+                              controller,
+                            ): controller is {
+                              name: string;
+                            } & typeof controller =>
+                              controller.name !== undefined,
+                          )
```
huwshimi: This could use `utils/filterBoolean` which also handles the type narrowing.

huwshimi, line 138: We can improve this in the future, as a controller always has a name but we need to update some types (not for this PR - it would have wider impact).

src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
huwshimi: Yep I think it would be safe to say that we can't get to this point of these values aren't defined.

src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
```diff
@@ -67,6 +71,33 @@ const UpgradeModelController: FC<Props> = ({
...
+
+    if (needsMigration) {
+      const action = upgradeTo.run({
+        targetVersion: version.version,
+        targetController: values[FieldName.TARGET_CONTROLLER],
+        currentVersion,
+        wsControllerURL,
+        modelUUID,
+        modelURL: wsControllerURL.replace("/api", `/model/${modelUUID}/api`),
```
huwshimi, line 95: Urvashi has implemented a util for this, so if that lands first it would be good to use it here: https://github.com/canonical/juju-dashboard/pull/2285/changes#diff-b8618585b6c69654e35b6b6a0f188d72b59777badd2798cfd1d027a396ed0566 otherwise may be worth leaving a todo.

src/panels/UpgradeModelPanel/UpgradeModelController/UpgradeModelController.tsx
```diff
@@ -67,6 +71,33 @@ const UpgradeModelController: FC<Props> = ({
...
+        modelName: model?.info?.name ?? "Unknown model",
```
huwshimi: I think it's fairly safe to assume you can't get to this point if the model info isn't defined so it'd probably be better to put this above with the throws rather than setting "Unknown model".

src/store/middleware/process/model-upgrade/model-connection-retry-source.ts
```diff
@@ -0,0 +1,74 @@
...
+ * reconnect on the next call.
+ */
+export function createModelConnectionRetry(modelConnection: ManagedConnection) {
+  // A promise which will resolve to the current connection, or `null` if a connection needs to be
+  // initiated.
+  let connection: null | Promise<ManagedConnection> =
+    Promise.resolve(modelConnection);
+
+  return async (): Promise<ConnectionRetryResult> => {
+    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
```
huwshimi: I think this is a legit case for using `??=`.

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+
+import { createProcess } from "../createProcess";
+
+import createModelConnectionRetrySource from "./model-connection-retry-source";
+
+export type UpgradeStatus =
+  | { status: "initiated" }
+  | { status: "loading" }
+  | { status: "pending" }
+  | { status: "reconnecting" };
```
huwshimi: If the status param uses an enum it will save the union. It means we can use the enum elsewhere e.g. when yielding.

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+    targetController,
+  );
+  yield { status: "pending" };
+  const result = await request;
+
+  if (!result) {
+    throw new Error("migration request rejected");
+  }
+
+  yield { status: "initiated" };
```
huwshimi: These yields could use the enum noted above.

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  // Poll for model version.
+  const source = createModelConnectionRetrySource(modelConnection);
+  let statusCount = 0;
+  for await (const retry of source) {
+    if ("version" in retry) {
+      if (retry.version === targetVersion) {
+        break;
```
huwshimi, line 70: It might be good to leave a comment here, something like `// The model has reached the target version so the source can now finish.`

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+          logger.info(`Attempting to reconnect to ${modelName} (${modelUUID})`);
+          break;
+        default:
+          logger.warn("An unknown status was emitted", modelUUID, status);
+          break;
+      }
+
+      if (!message) {
+        // Don't toast anything if there's no message.
+        return {} as PayloadAction;
```
huwshimi: I think this indicates we should be able to return null/undefined from `setStatus`

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+        });
+      } else {
+        return createToast({
+          message: `Model "${modelName}" upgraded to ${targetVersion}`,
+          severity: "positive",
+        });
+      }
+    },
+    setStatus: ({ modelUUID, modelName }, status) => {
+      let message: string | undefined = undefined;
```
huwshimi: This is an assignment and I can't see any reason that this needs to be `undefined` so this should default to `null`.

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+      switch (status.status) {
+        case "pending":
```
huwshimi: These should use the enum values noted above.

src/store/middleware/process/model-upgrade/upgrade-to.ts
```diff
@@ -0,0 +1,172 @@
...
+          logger.info(`${modelName} (${modelUUID}) upgrade pending`);
```
huwshimi: These are going to be quite noisy. I wonder if we should use logger.debug instead?

huwshimi (commented):
Some preliminary comments. I'll do another pass once I've investigate the error I'm seeing.

src/components/ToastCard/_toast-card.scss
```diff
@@ -45,7 +45,7 @@
     border-left: 3px solid $color-negative;
   }
 
-  &[data-type="info"] {
+  &[data-type="information"] {
```
huwshimi, line 48: @andogq I snuck in this fix so that the notifications appear with the correct styles.

andogq: I was wondering why they didn't show.... thank you!

huwshimi (approved):
I've address the comments I've left. This looks good now.

PR #2313, opened 2026-04-29
feat: implement keyboard navigation to the autocomplete input

huwshimi:
## Done

- Add keyboard navigation to the autocomplete input.

## QA

- You can either test this feature in the model upgrades form or via Logs -> Filter.
- Check that you can navigate up and down the options with arrow keys and select an option with enter.
- With the menu closed enter should submit the form.
- With the menu open escape should close the menu, pressing it again should close the panel.

## Details

https://warthogs.atlassian.net/browse/JUJU-9498

### DIFF
--- a/src/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/src/components/AutocompleteInput/AutocompleteInput.tsx
@@ -64,7 +64,6 @@ const handleKeyDown = (
   onSelectItem: (item: string) => void,
   event: React.KeyboardEvent<HTMLInputElement>,
 ): void => {
-  const isTab = event.key === "Tab";
   const isUp = event.key === "ArrowUp";
   const isDown = event.key === "ArrowDown";
   const isUpDown = isUp || isDown;
@@ -87,15 +86,6 @@ const handleKeyDown = (
     // When the menu is open then pressing enter should not do the default action (e.g. submitting a form).
     event.preventDefault();
   }
-  if (isTab && isDropdownOpen) {
-    // Moving focus away from the field should close the dropdown.
-    updateDropdown(
-      setIsDropdownOpen,
-      filteredOptions,
-      setHighlightedOptionIndex,
-      false,
-    );
-  }
   if (
     isEnter &&
     highlightedOptionIndex !== null &&
@@ -104,7 +94,7 @@ const handleKeyDown = (
     // Select the highlighted item in the menu.
     onSelectItem(filteredOptions[highlightedOptionIndex].value);
   }
-  // Pressing up/down or tab/shift+tab should move the selected item up/down.
+  // Pressing up/down should move the selected item up/down.
   if (isUpDown) {
     setHighlightedOptionIndex((prevIndex) => {
       return getNextOptionIndex(filteredOptions, isUp, prevIndex);
@@ -224,6 +214,16 @@ const AutocompleteInput: FC<Props> = ({
             );
             // Call any parent onFocus handlers.
             props.onFocus?.(event);
+          }}
+          onBlur={(event) => {
+            updateDropdown(
+              setIsDropdownOpen,
+              filteredOptions,
+              setHighlightedOptionIndex,
+              false,
+            );
+            // Call any parent onBlur handlers.
+            props.onBlur?.(event);
           }}
           onKeyDown={(event) => {
             handleKeyDown(
--- a/src/components/AutocompleteInput/AutocompleteInput.tsx
+++ b/src/components/AutocompleteInput/AutocompleteInput.tsx
@@ -1,13 +1,13 @@
 import type { InputProps, PropsWithSpread } from "@canonical/react-components";
 import { ContextualMenu, Input } from "@canonical/react-components";
 import type { FC, HTMLProps } from "react";
-import { useId, useMemo, useState } from "react";
+import { useCallback, useId, useMemo, useState } from "react";
 
 import AutocompleteInputDropdown from "./AutocompleteInputDropdown";
 
 type AutocompleteInputItem = {
   label: string;
-  value: number | string;
+  value: string;
 };
 
 type Props = PropsWithSpread<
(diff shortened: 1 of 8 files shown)


src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
```diff
@@ -1,49 +1,101 @@
...
-    event.stopPropagation();
-    const value = (event.target as HTMLLIElement)?.dataset.value;
-    if (value) {
-      onSelectItem(value);
+  // Wait for any rerenders and then scroll to items that are outside the visible scroll area.
+  useLayoutEffect(() => {
+    if (highlightedOptionIndex !== null) {
+      // The event needs to happen after repaint in case the DOM changed.
+      window.requestAnimationFrame(
+        scrollToOption.bind(this, id, highlightedOptionIndex),
```
andogq: It's not a problem, but any particular reason for using bind instead of a closure here?

If not, `useId` would be better for generating these IDs.

src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
```diff
@@ -1,49 +1,101 @@
...
-            >
-              {item.label}
-            </li>
-          ))}
+          {options.map((item, index) => {
+            const selected = index === highlightedOptionIndex;
+            return (
+              <li
+                key={item.value}
+                id={getOptionId(id, index)}
```
andogq, line 67: I feel like id prefer to see maintaining refs to these options directly, rather than computing IDs and querying the DOM.

src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
```diff
@@ -1,49 +1,101 @@
...
+                aria-selected={selected}
+                className={classNames(
+                  "autocomplete-input__dropdown-item p-contextual-menu__link",
+                  {
+                    highlight: selected,
+                  },
+                )}
+                onMouseMove={() => {
```
andogq: Would mouseenter be better here, to prevent spamming events?

andogq (commented):
Looking good overall. All the work going to/from IDs feels a bit spooky, it'd be good to use element Refs instead if possible, just to save the browser some work.

src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
huwshimi: I don't think so, might have been part of an issue I was trying to solve around rerenders but I can switch this to a closure.

src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
huwshimi: Yes! Thanks.

src/components/AutocompleteInput/AutocompleteInputDropdown/AutocompleteInputDropdown.tsx
huwshimi: Thanks, that's actually how it was before I discovered a rerender issue, so I changed it to use ids. However now that I'm using `requestAnimationFrame` this can be reverted to use the refs.

huwshimi (commented):
Thanks, I've addressed those issues.

andogq (approved):

andogq:
My JIMM died halfway through QA-ing this, but I think there just needs to be a few accessibility tweaks to bring it inline with [the guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/combobox).

The biggest one is that Tab/Shift-Tab shouldn't be used for navigating the list items. Additionally, it seems that in [the example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/) they use wrapping with up/down arrow keys in place of start/end keys, so up to you if you want to do that too.

huwshimi:
@andogq thanks, I've addressed both navigation issues.

src/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -19,12 +19,118 @@ type Props = PropsWithSpread<
...
+  }
+  if (
+    isEnter &&
+    highlightedOptionIndex !== null &&
+    filteredOptions[highlightedOptionIndex]
+  ) {
+    // Select the highlighted item in the menu.
+    onSelectItem(filteredOptions[highlightedOptionIndex].value);
+  }
+  // Pressing up/down or tab/shift+tab should move the selected item up/down.
```
andogq:
```suggestion
  // Pressing up/down should move the selected item up/down.
```

src/components/AutocompleteInput/AutocompleteInput.tsx
```diff
@@ -19,12 +19,118 @@ type Props = PropsWithSpread<
...
+        setHighlightedOptionIndex,
+        true,
+      );
+    }
+  }
+  if (isEnter && isDropdownOpen) {
+    // When the menu is open then pressing enter should not do the default action (e.g. submitting a form).
+    event.preventDefault();
+  }
+  if (isTab && isDropdownOpen) {
```
andogq: I wonder if this'd be better triggered when focus is lost on the combobox. If it's working as is though, no need to change it.

andogq (approved):
Tiny comment to update, but looks good overall.

src/components/AutocompleteInput/AutocompleteInput.tsx
huwshimi: Thanks, done.

src/components/AutocompleteInput/AutocompleteInput.tsx
huwshimi: Done.

andogq:
Thanks for these changes Huw, works nicely now.

PR #2314, opened 2026-04-30
fix(models): don't clear model-index screen for error

andogq:
## Done

- Allow rendering content in model-index even if there's an error.

## QA

- Apply diff:

```diff
diff --git a/src/store/middleware/source/model-list.ts b/src/store/middleware/source/model-list.ts
index 758def65..be6e9830 100644
--- a/src/store/middleware/source/model-list.ts
+++ b/src/store/middleware/source/model-list.ts
@@ -12,9 +12,13 @@ export const NOT_AUTHENTICATED_ERROR = "not authenticated with controller";
 export const NO_MODEL_MANAGER_FACADE =
   "ModelManager facade is not available on the connection";
 
+let count = 0;
 export async function getModelList(
   connection: ConnectionWithFacades,
 ): Promise<UserModelList> {
+  if (count++ === 1) {
+    throw new Error(ModelsError.LOAD_LATEST_MODELS);
+  }
   if (!connection.info.user?.identity) {
     throw new Error(NOT_AUTHENTICATED_ERROR);
   }
@@ -47,7 +51,7 @@ export default createSourceMiddleware<
     const connection = meta.connections.wsControllerURL;
 
     return createPollingSource(async () => getModelList(connection), {
-      interval: { seconds: 30 },
+      interval: { seconds: 5 },
     });
   },
   {
```
- Go to model index page
- Loading spinner should show, then models should populate
- After 5 seconds, an error should be shown at the top of the table
- After another 5 seconds, the error will clear

https://github.com/user-attachments/assets/8a7ae8db-ca30-4af0-8afb-6cc87625602f

### DIFF
--- a/src/pages/ModelsIndex/ModelsIndex.tsx
+++ b/src/pages/ModelsIndex/ModelsIndex.tsx
@@ -124,99 +124,94 @@ export default function Models(): JSX.Element {
   const modelCount = blocked + alert + running;
 
   let content: ReactNode = null;
-  if (!modelsError) {
-    if (!modelsLoaded) {
-      return <LoadingSpinner />;
-    } else if (!hasSomeModels) {
-      content = (
-        <div className="models">
-          <h3>{Label.NOT_FOUND}</h3>
-          <p>
-            Learn about <a href={externalURLs.addModel}>adding models</a> or{" "}
-            <a href={externalURLs.modelAccess}>granting access</a> to existing
-            models.
-          </p>
-        </div>
-      );
-    } else {
-      content = (
-        <>
-          <span className="models__controls" data-disabled={modelCount === 0}>
-            <SearchAndFilter
-              filterPanelData={[
-                {
-                  id: 0,
-                  heading: "Cloud",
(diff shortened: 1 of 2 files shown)


src/pages/ModelsIndex/ModelsIndex.tsx
```diff
@@ -124,99 +124,94 @@ export default function Models(): JSX.Element {
...
-                // Reset active filters
-                activeFilters = {
-                  cloud: [],
-                  owner: [],
-                  region: [],
-                  credential: [],
-                  custom: [],
-                };
+  if (!modelsLoaded && !modelsError) {
+    return <LoadingSpinner />;
```
huwshimi, line 128: I think we probably need a test for this case and also that it displays both the error and content so we don't accidentally break it again :).

andogq: Good point, I've updated a test to check for the rows.

huwshimi (approved):

PR #2315, opened 2026-05-01
fix: only fetch model migration data for jimm

huwshimi:
## Done

- Only fetch the Juju versions and model migration targets when using JIMM.

## QA

- Connect to Juju.
- Go to the model list.
- Click the model actions menu.
- You console shouldn't have any errors about fetching jimm data.

### DIFF
--- a/src/hooks/useModelMigrationData.test.ts
+++ b/src/hooks/useModelMigrationData.test.ts
@@ -1,3 +1,5 @@
+import jimmSupportedVersions from "store/middleware/source/jimm-supported-versions";
+import migrationTargets from "store/middleware/source/migration-targets";
 import type { RootState } from "store/store";
 import { generalStateFactory, configFactory } from "testing/factories/general";
 import {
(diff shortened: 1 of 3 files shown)


src/hooks/useModelMigrationData.test.ts
```diff
@@ -30,6 +30,35 @@ describe("useModelMigrationData", () => {
...
+      () => {
+        useModelMigrationData("test1", "eggman@external");
+      },
+      {
+        store,
+      },
+    );
+    expect(
+      actions.find(
+        (dispatch) => dispatch.type === "source/jimm-supported-versions/start",
```
andogq: Maybe over the top, but would it be better to use `jimmSupportedVersions.actions.start.match` instead here, so that the test is still checking the correct thing if the action name ever changes?


```suggestion
        (dispatch) => jimmSuportedVersions.actions.start.match(dispatch),
```

andogq (approved):
One tiny thing, take it or leave it. Thanks for the doc updates!

src/hooks/useModelMigrationData.test.ts
huwshimi: Thanks, done. When I previously tried this it was giving some kind of import loop, but it looks like that has been resolved now.

andogq (approved):

PR #2316, opened 2026-05-04
feat: sort model list by recently updated

huwshimi:
## Done

- Sort the model list by most recently updated to prevent the list being reordered by polling updates.

## QA

- Go to the model list.
- Check that the most recently updated is at the top.

### DIFF
--- a/src/components/ModelTableList/shared.tsx
+++ b/src/components/ModelTableList/shared.tsx
@@ -16,10 +16,13 @@ export const getRegion = (model: ModelData): string | undefined =>
 export const getCredential = (model: ModelData): string =>
   extractCredentialName(model.info?.["cloud-credential-tag"]);
 
-export const getLastUpdated = (model: ModelData): string | undefined => {
+export const getLastUpdated = (
+  model: ModelData,
+): [string | undefined, string | undefined] => {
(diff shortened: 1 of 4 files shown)


src/components/ModelTableList/shared.tsx
```diff
@@ -56,24 +71,23 @@ export const generateTableHeaders = (
...
+    { content: "Controller", sortKey: SortKey.CONTROLLER },
     {
       content: "Last Updated",
-      sortKey: "lastUpdated",
+      sortKey: SortKey.LAST_UPDATED,
       className: "u-align--right",
     },
     {
       content: "Actions",
-      sortKey: "actions",
```
andogq, line 76: No sortkey?

andogq (approved):

src/components/ModelTableList/shared.tsx
huwshimi: The sort key meant you could sort by the actions column, but as the content is only a button it doesn't make sense to have it sortable.

andogq: Ah of course, I was thinking Juju actions for some reason

PR #2317, opened 2026-05-05
chore(dev): disable ipv6 for jimm/lxd cloud init script

huwshimi:
## Done

- Disable ipv6 when launching a jimm + LXD container
- Fix installing playwright browsers for Ubuntu 24.04


andogq (approved):
:shipit:

PR #2319, opened 2026-05-07
feat: display a notification if the upgrade fails

huwshimi:
## Done

- Display a notification if the upgrade job fails.

## QA

- Set up controllers for a migration (see the HACKING doc).
- Add a model to migrate and then in your terminal get ready to delete the model immediately after you start the migration: `juju destroy-model --no-prompt --force --no-wait test-model-name` (don't press enter yet).
- In the dashboard start the migration and then immediately press enter in your terminal to delete the model (it might give an error so you might need to try and delete it a few times).
- Wait for JIMM to attempt the migration (it might take a few minutes while JIMM tries to retry the migration) and then a notification should appear with a message that the migration failed.

## Details

https://warthogs.atlassian.net/browse/JUJU-9507

### DIFF
--- a/src/juju/jimm/api.test.ts
+++ b/src/juju/jimm/api.test.ts
@@ -452,7 +452,7 @@ describe("JIMM API", () => {
   });
 
   describe("upgradeTo", () => {
-    it("makes and upgrade call", async () => {
+    it("makes an upgrade call", async () => {
       const conn = {
         facades: {
           jimM: {
@@ -492,7 +492,7 @@ describe("JIMM API", () => {
   });
 
   describe("jobInfo", () => {
-    it("makes and upgrade call", async () => {
+    it("makes a job info call", async () => {
       const conn = {
         facades: {
           jimM: {
--- a/src/juju/jimm/api.test.ts
+++ b/src/juju/jimm/api.test.ts
@@ -19,6 +19,7 @@ import {
   listMigrationTargets,
   supportedJujuVersions,
   upgradeTo,
+  jobInfo,
 } from "./api";
 
 describe("JIMM API", () => {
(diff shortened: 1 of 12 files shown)


src/juju/jimm/api.test.ts
```diff
@@ -489,4 +490,39 @@ describe("JIMM API", () => {
       ).rejects.toThrow(new Error("Uh oh!"));
     });
   });
+
+  describe("jobInfo", () => {
+    it("makes and upgrade call", async () => {
```
Ninfa-Jeon: I suppose you missed changing the test description to something like:
```suggestion
    it("fetches job status", async () => {
```
Also, I believe the original test should say something like:
```
it("makes an upgrade call", async () => {
```

src/store/middleware/process/model-upgrade/job-info-source.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+const JOB_INFO_POLL_S = 5;
+
+/**
+ * Internal logic for `createJobInfoSource`.
+ *
+ * Saves a connection, and whenever the returned function is called, it will try to call the
+ * `fullStatus` facade method. If the call fails, it will clear the connection, and attempt to
+ * reconnect on the next call.
+ */
```
Ninfa-Jeon: This comment needs to be updated to match the correct description of the function below. Something like:
```suggestion
 * Receives a connection, and whenever the returned function is called, it will call the
 * `jobInfo` facade method.
 */
```
Feel free to change it as you see fit though :)

src/store/middleware/process/model-upgrade/job-info-source.ts
```diff
@@ -0,0 +1,38 @@
...
+  jobId: string,
+) {
+  return async (): Promise<JobInfoResponse> =>
+    jobInfo(controllerConnection, jobId);
+}
+
+/**
+ * A source which will produce the version of a model, or an indication that it's being reconnected
+ * to.
+ */
```
Ninfa-Jeon, line 28: Same as my other comment, this description needs changing.
```suggestion
/**
 * A source which will produce the information for the provided job-id.
 */
```

Ninfa-Jeon (commented):
Just a few minor comment and description changes required.

Ninfa-Jeon:
For some reason, the model I added simply gets destroyed and wiped out of the dashboard after a while with no notifications either (PFA). I did see the notification once but I can't figure out the exact steps I took for it.

https://github.com/user-attachments/assets/26cd2b45-1694-4be6-a14e-fe7400c344de

Ninfa-Jeon:
Okay, I think the way to see this notification is to wait for a bit longer without refreshing the dashboard. My first instinct on seeing the model disappear as a user was to refresh the dashboard which must have wiped the state checks for notification too. Do you think it would be better to keep showing a migrating model in the list even though it has moved from first controller and is yet to appear on the other one?...that could prevent a user from refreshing the dashboard.

Ninfa-Jeon:
The QA is working as expected for this, however, I think there needs to be some visual feedback for when the migration is being retried. It is quite a wait time between the model disappearing and the error notification appearing.
Screenshot 2026-05-08 at 4 07 29 PM

Ninfa-Jeon (approved):

src/juju/jimm/api.test.ts
huwshimi: Done!

src/store/middleware/process/model-upgrade/job-info-source.ts
huwshimi: Oops, thanks, done.

src/store/middleware/process/model-upgrade/job-info-source.ts
huwshimi: Fixed this too.

huwshimi:
> The QA is working as expected for this, however, I think there needs to be some visual feedback for when the migration is being retried. It is quite a wait time between the model disappearing and the error notification appearing. 

Agreed, it would be nice to show some more feedback if it's taking a long time. I think that needs to include a way to know that an upgrade is in progress if you load the dashboard part way through (or refresh, like you did).

PR #2327, opened 2026-05-26
chore: enable owner access reduction for jimm

Ninfa-Jeon:
## Done

- Enabled model owner access reduction flow for JIMM.
This was previously disabled due to a mismatch between how JIMM did access management vs how juju did.
- Updated tests.
- Removed unused label.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Point the dashboard to JIMM and login as a non-JIMM admin, non-superuser.
- Open the "AddModel" page and fill in the mandatory details.
- Go to the "Access Management" tab and add another user to share the model with.
- Check that the model owner's access dropdown remains disabled so far.
- Now, bump the newly added user's access to admin.
- Model owner's dropdown should now be enabled (this was not the case previously).
- Lower the model owner's permission to "read" and add the model.
- The model should be added successfully and the model should appear in the model's list with appropriate user access for both the model owner and the other user.

## Details

https://warthogs.atlassian.net/browse/JUJU-9822

## Recordings

https://github.com/user-attachments/assets/a656dd67-8328-4a37-b7bc-1e2106967349

### DIFF
--- a/src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
+++ b/src/pages/AddModel/AccessManagement/AccessManagement.test.tsx
@@ -259,7 +259,7 @@ describe("AccessManagement", () => {
     expect(adminButtons[0]).not.toHaveAttribute("aria-disabled", "true");
   });
 
-  it("does not enable the active user dropdown when another user has admin access for JIMM", async () => {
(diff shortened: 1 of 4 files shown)


huwshimi (approved):
Thanks, this works now!

PR #2328, opened 2026-05-27
feat: add error state to stepper, validation to access tab

Ninfa-Jeon:
## Done

- Moved the stepper component within AddModel form and into its own component to allow it access to error states in form tabs
- Added error state to steps that appears when the user navigates away from current tab
- Added error label to step that appears whenever there is an error state on a stepper tab
- Added `isNumeric` key to the configCatalog which is then used to render the input as `type=number`
- Introduced email validation logic to JIMM add-model flow
- The validation is run when a user tries to add an email of invalid format to share their model with
- Added truncation and overflow to username/emails

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Point the dashboard to JIMM to test all validation in one setup
- Go to the AddModel page
- Trying navigating away from the first tab with no model name set and with an invalid name set (e.g., `-model`)
- The stepper icon should change to error and an error label must be visible under tab title
- On the configs+constraints tab, look for `net-bond-reconfigure-delay` and try entering a non-numerical value, it should only allow number input
- On the access management tab, try adding an invalid email to the list
- An error label, `Incorrect email format` should appear on clicking the add button
- This should go away when continuing to type further and reappear on add click if the input is still invalid

## Details

https://warthogs.atlassian.net/browse/JUJU-9830

## Screenshots

https://github.com/user-attachments/assets/19dfecda-90a7-4f1e-a78c-dbfcc3d84568

### DIFF
--- a/src/pages/AddModel/AccessManagement/AccessManagement.tsx
+++ b/src/pages/AddModel/AccessManagement/AccessManagement.tsx
@@ -110,6 +110,7 @@ const AccessManagement = (): JSX.Element => {
             variant="search"
             searchButtonType="button"
             inputClassName="u-no-margin--bottom u-sv1--top"
+            dropdownClassName="u-overflow--scroll"
             footerClassName="access-management__dropdown-footer u-no-padding--left u-no-padding--right"
             placeholder={Label.MULTI_SELECT_PLACEHOLDER}
             label={Label.MULTI_SELECT_LABEL}
@@ -123,11 +124,13 @@ const AccessManagement = (): JSX.Element => {
                   <Button
                     appearance="base"
                     type="button"
-                    className="u-align--left u-full-width u-no-margin--bottom u-sv-1--top"
+                    className="u-align--left u-full-width u-no-margin--bottom u-sv-1--top add-user-button"
                     onClick={handleAddUser}
                   >
-                    <Icon name="plus" className="u-sh1--right" />
-                    {trimmedSearchInput}
+                    <span className="u-truncate">
+                      <Icon name="plus" className="u-sh1--right" />
+                      {trimmedSearchInput}
+                    </span>
                     <p
                       className={classNames(
                         "u-text--muted",
--- a/src/pages/AddModel/AccessManagement/AccessManagement.tsx
+++ b/src/pages/AddModel/AccessManagement/AccessManagement.tsx
@@ -5,6 +5,7 @@ import {
   Tooltip,
   type MultiSelectItem,
 } from "@canonical/react-components";
+import classNames from "classnames";
 import { useFormikContext } from "formik";
 import type { JSX } from "react";
 import { useState } from "react";
@@ -33,6 +34,8 @@ import {
   getUserAccess,
 } from "./utils";
 
+const EMAIL_PATTERN = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+
 const AccessManagement = (): JSX.Element => {
   const isJuju = useAppSelector(getIsJuju);
   const wsControllerURL = useAppSelector(getWSControllerURL);
@@ -44,6 +47,7 @@ const AccessManagement = (): JSX.Element => {
   );
   const { values, setFieldValue } = useFormikContext<AddModelFormState>();
   const [searchInput, setSearchInput] = useState("");
+  const [showInvalidEmailHint, setShowInvalidEmailHint] = useState(false);
 
   const activeUserName = userName ? getUserName(userName) : undefined;
   const shareModelWith = values.shareModelWith ?? {};
@@ -81,6 +85,13 @@ const AccessManagement = (): JSX.Element => {
       return;
     }
 
+    if (!isJuju && !EMAIL_PATTERN.test(trimmedSearchInput)) {
+      setShowInvalidEmailHint(true);
+      return;
+    }
+
+    setShowInvalidEmailHint(false);
+
     void setFieldValue(
       `shareModelWith["${trimmedSearchInput}"]`,
       AccessLevel.READ,
@@ -117,8 +128,20 @@ const AccessManagement = (): JSX.Element => {
                   >
                     <Icon name="plus" className="u-sh1--right" />
                     {trimmedSearchInput}
-                    <p className="u-text--muted p-text--small u-sh3 u-sv-1">
-                      {formatHint}
+                    <p
+                      className={classNames(
+                        "u-text--muted",
+                        "p-text--small",
+                        "u-sh3",
+                        "u-sv-1",
+                        {
+                          "is-negative": showInvalidEmailHint,
+                        },
+                      )}
+                    >
+                      {showInvalidEmailHint
+                        ? Label.INCORRECT_EMAIL_FORMAT
+                        : formatHint}
                     </p>
                   </Button>
                 ) : (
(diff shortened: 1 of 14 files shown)


src/pages/AddModel/ConfigsConstraints/configCatalog.ts
```diff
@@ -18,6 +18,40 @@ const BOOLEAN_OPTIONS = [
 ];
 
 export const CONFIG_CATEGORIES: CategoryDefinition[] = [
+  {
+    category: "Cloud-specific configurations",
```
Ninfa-Jeon, line 23: Moved this category of configs up as per a previous discussion with Lukas

src/pages/AddModel/AddModel.tsx
```diff
@@ -184,15 +169,54 @@ const AddModel: FC = () => {
...
-              id={currentStep.key}
-              {...testId(TestId.ADD_MODEL_FORM)}
-              className={currentStep.key}
-              stacked={currentStep.key === StepType.CONFIGURATION_CONSTRAINTS}
-            >
-              {currentStep.content}
-            </FormikFormData>
+            {({ errors }) => (
+              <>
+                <Stepper
```
Ninfa-Jeon: Had to move this within the form as we only need to show error state when the user navigates away from the tab. This seemed the most efficient.

src/pages/AddModel/AccessManagement/AccessManagement.tsx
```diff
@@ -131,9 +154,13 @@ const AccessManagement = (): JSX.Element => {
             disabledItems={ACTIVE_USER ? [ACTIVE_USER] : []}
             items={selectedItems}
             selectedItems={selectedItems}
-            onSearchChange={setSearchInput}
+            onSearchChange={(value) => {
+              setSearchInput(value);
+              setShowInvalidEmailHint(false);
```
Ninfa-Jeon, line 162: The error label is cleared if the user types further into the same query after the error is shown.

src/pages/AddModel/_add-model.scss
```diff
@@ -14,6 +14,16 @@
...
+
+    .stepper-horizontal {
+      margin-bottom: $spv--large;
+
+      .step-optional-content {
+        color: $color-negative;
+        font-size: #{map-get($font-sizes, small)}rem;
+        max-width: fit-content;
+      }
+    }
```
huwshimi, line 26: It looks like you can use `linkProps` on `<Step />` to pass classes to the step so you can remove some of this custom css.

src/pages/AddModel/AddModel.tsx
```diff
@@ -184,15 +164,51 @@ const AddModel: FC = () => {
...
-            <FormikFormData
-              onValidate={setIsValid}
-              id={currentStep.key}
-              {...testId(TestId.ADD_MODEL_FORM)}
-              className={currentStep.key}
-              stacked={currentStep.key === StepType.CONFIGURATION_CONSTRAINTS}
-            >
-              {currentStep.content}
-            </FormikFormData>
+            {({ errors }) => (
```
huwshimi: We don't use this pattern of because it's essentially generating a new component on each render. Instead,  you can use `useFormikContext` to get the errors. However, it will require moving the `<Stepper ...` part into its own component.

src/pages/AddModel/ConfigsConstraints/StackList/StackList.tsx
```diff
@@ -26,7 +26,10 @@ const StackList = ({ visibleConfigs }: Props): JSX.Element => {
             <FormikField
               {...(config.input?.type === "select"
                 ? { component: Select }
-                : { type: "text", placeholder: config.defaultValue })}
+                : {
+                    type: config.isNumeric ? "number" : "text",
```
huwshimi, line 30: It's great that this worked.

huwshimi (commented):
When the model name is empty the tab validation shows "Incorrect model name format." but it should say something about it being required or not provided.

On the first step it's also possible that you don't have access to a cloud or haven't added credentials so we should also validate those as well.

It wasn't introduced in this PR, but I noticed that long usernames break out of the multiselect (could be done in a followup):

Image

src/pages/AddModel/ConfigsConstraints/StackList/StackList.tsx
Ninfa-Jeon: My thoughts, exactly haha

src/pages/AddModel/_add-model.scss
Ninfa-Jeon: the linkProps will forward the props to a link element instead of the span that is rendered when we use the `label`

huwshimi: Ah OK, I see there are two elements with `step-optional-content`. Ignore this suggestion then :).

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: Done!

Ninfa-Jeon:
> On the first step it's also possible that you don't have access to a cloud or haven't added credentials so we should also validate those as well.

As discussed, we don't allow the user into the experience if they don't have access to cloud. As for credentials, I shall have a discussion with Lukas for the most appropriate approach. Thanks!

Ninfa-Jeon:
> When the model name is empty the tab validation shows "Incorrect model name format." but it should say something about it being required or not provided.

I have added separate labels for both cases, thanks for reporting

Ninfa-Jeon:
> It wasn't introduced in this PR, but I noticed that long usernames break out of the multiselect (could be done in a followup):

Fixed in two spots :) 

https://github.com/user-attachments/assets/4713b696-cca3-46f4-842a-f4a4000e7fa3

src/pages/AddModel/AddModelStepper/AddModelStepper.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+const initialValues: AddModelFormState = {
+  modelName: "",
+  cloud: "",
+  region: "",
+  credential: "",
+  [FieldName.CONFIG_INPUT_MODE]: InputMode.LIST,
+  [FieldName.CONFIG_YAML]: "",
+  [FieldName.DISABLED_COMMANDS]: DisableType.NONE,
+};
```
huwshimi: These will need to be created in a beforeEach as there's nothing to prevent them getting mutated in tests.

src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
```diff
@@ -0,0 +1,58 @@
...
+import type { JSX } from "react";
+
+import { type AddModelFormState, Label, type StepType } from "../types";
+
+type Props = {
+  stepDefinitions: Array<{
+    key: StepType;
+    title: string;
+    content: JSX.Element;
+  }>;
```
huwshimi: This should probably get defined in one place rather than here and in AddModel.tsx.

huwshimi (approved):
A couple of small things. This doesn't need another review once they're changed.

src/pages/AddModel/AddModelStepper/AddModelStepper.test.tsx
Ninfa-Jeon: right, done

src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
Ninfa-Jeon: moved to types.

PR #2329, opened 2026-05-28
feat: upgrade model patch versions in place

huwshimi:
## Done

- Add the ability to upgrade a patch version without a controller migration.

## QA

- Set up JIMM with a model that has a [lower version than the controller ](https://github.com/canonical/juju-dashboard/blob/main/HACKING.md#add-a-model-with-a-version-older-than-the-controller).
- Open the upgrade model panel and choose the manual option.
- Click inside the field and select the version that matches the version of the controller it is on.
- Go to the confirmation screen and you shouldn't see an option for selecting a controller.
- Confirm and submit. The upgrade should start and then finish very quickly and the version in the model list should update to the chosen version.

## Details

https://warthogs.atlassian.net/browse/JUJU-9735

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -42,6 +42,7 @@
         "jujucharms",
         "jujulib",
         "microk8s",
+        "upgrader",
         "rebac",
         "snapstore",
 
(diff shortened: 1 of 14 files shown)


Ninfa-Jeon (approved):
All good, thanks!

PR #2331, opened 2026-05-29
fix: only fetch migration targets as admin

huwshimi:
## Done

- Fixed model upgrades so that migration targets are only fetched if the user is a JIMM admin.

## QA

- Add a model that can be upgraded.
- Log in to the dashboard as a JIMM admin.
- Open the model actions menu and the upgrade item should be enabled.
- Add a non JIMM admin user (see the new instructions in the hacking doc if needed).
- Log in to the dashboard as the non-admin user.
- Open the model actions menu and the upgrade item should not be visible and there shouldn't be any errors in the console.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -381,7 +381,7 @@ To be able to add models, first as the admin user give them permission to add mo
 
 Next you will need to log out of the admin account and log in as the user you just created.
 
-NOTE: when logging in as the new user you may need to use another browser or private window if you've previously logged in as jimm-test otherwise you'll get logged in as the admin.
+NOTE: when logging in as the new user you may need to clear your cookies or use another browser or private window if you've previously logged in as jimm-test otherwise you'll get logged in as the admin.
 
 ```shell
 juju unregister jimm-dev
--- a/HACKING.md
+++ b/HACKING.md
@@ -39,6 +39,7 @@ contribute and what kinds of contributions are welcome.
       - [Vanilla React Components](#vanilla-react-components)
   - [Deployed JIMM controller](#deployed-jimm-controller)
     - [Adding users](#adding-users)
+      - [Add model permissions](#add-model-permissions)
     - [Self signed certificates](#self-signed-certificates)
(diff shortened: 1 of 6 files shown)


HACKING.md
```diff
@@ -369,9 +370,31 @@ form.
...
+To be able to add models, first as the admin user give them permission to add models on the cloud and controller, substituting the user's email address and controller name (for the microk8s setup it would be controller-qa-microk8s).
+
+```shell
+~/jimm/jaas add-permission user-test@example.com can_addmodel cloud-localhost
+~/jimm/jaas add-permission user-test@example.com can_addmodel controller-qa-lxd
+```
+
+Next you will need to log out of the admin account and log in as the user you just created.
+
+NOTE: when logging in as the new user you may need to use another browser or private window if you've previously logged in as jimm-test otherwise you'll get logged in as the admin.
```
Ninfa-Jeon: Clearing the cookies also works btw

src/components/ModelActions/ModelActions.tsx
```diff
@@ -56,28 +56,29 @@ const generateLinks = (
...
-    >
-      {Label.UPGRADE}
-      {upgradeDataLoaded ? null : <Spinner className="u-sh1" />}
-    </Button>
-  );
+  const upgradeButton: ReactNode =
+    isJuju || !isJIMMControllerAdmin ? null : (
+      <Button
+        className="p-contextual-menu__link"
+        disabled={!isJIMMControllerAdmin || !hasUpgrades}
```
Ninfa-Jeon: I believe this is not needed anymore

```suggestion
        disabled={!hasUpgrades}
```

Ninfa-Jeon (approved):
One minor correction. Everything else looks good, thanks.

HACKING.md
huwshimi: Thanks, I've added that.

src/components/ModelActions/ModelActions.tsx
huwshimi: Thanks! Fixed.

PR #2332, opened 2026-05-29
feat: add validation for missing credentials

Ninfa-Jeon:
## Done

- Added validation to AddModel form for when the user is missing cloud credentials
- Added appropriate error labels to the stepper for various combinations of errors

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Log into the dashboard as a new user with cloud+controller access but no credentials added
- Try navigating through the add-model flow with multiple combinations of invalid inputs
  - No model name, no credential
  - Invalid model name, no credential
  - Valid model name, no credential
  - Invalid model name, valid credential
  - All valid values to add model.
- All these scenarios should work well for both Juju and JIMM.
- The user should be able to add a model with all valid inputs

## Details

https://warthogs.atlassian.net/browse/JUJU-9892

## Screenshots

https://github.com/user-attachments/assets/a17693fc-a92f-4a6e-b749-29e0b8b3cc06

### DIFF
--- a/src/pages/AddModel/AddModel.tsx
+++ b/src/pages/AddModel/AddModel.tsx
@@ -171,6 +171,8 @@ const AddModel: FC = () => {
               ...getConfigInitialValues(CONFIG_CATEGORIES),
             }}
             validationSchema={validationSchema}
+            // Mark credential as touched on mount as Vanilla doesn't display validation until the field loses focus.
+            initialTouched={{ credential: true }}
             onSubmit={handleCreateClick}
           >
             <>
--- a/src/pages/AddModel/AddModel.tsx
+++ b/src/pages/AddModel/AddModel.tsx
@@ -38,6 +38,7 @@ import {
   type AddModelFormState,
   type StepDefinition,
 } from "./types";
+import { getCredentialError } from "./utils";
 
 const MODEL_NAME_PATTERN = /^[a-z0-9][a-z0-9-]*$/;
 
@@ -45,6 +46,16 @@ const validationSchema = Yup.object().shape({
   modelName: Yup.string()
     .matches(MODEL_NAME_PATTERN, Label.INCORRECT_MODEL_NAME_ERROR)
     .required("Required"),
+  credential: Yup.string().test({
+    test: function (value) {
+      if (value !== undefined && value !== "") {
+        return true;
+      }
+      return this.createError({
+        message: getCredentialError(this.parent.cloud),
+      });
+    },
+  }),
 });
 
 const stepDefinitions: StepDefinition[] = [
--- a/src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
+++ b/src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
@@ -22,26 +22,27 @@ const AddModelStepper = ({
       steps={stepDefinitions.map(({ key, title }, index) => {
         const isPrevious = index < currentStepIndex;
         const isCurrent = index === currentStepIndex;
-        let hasModelNameError = false;
-        let hasCredentialError = false;
         let label = undefined;
+        const labels = [];
         if (index === 0) {
           if (errors.modelName) {
-            hasModelNameError = true;
-            label =
+            labels.push(
               values.modelName !== ""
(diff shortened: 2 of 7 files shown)


src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
```diff
@@ -22,21 +22,31 @@ const AddModelStepper = ({
...
+          if (errors.credential) {
+            hasCredentialError = true;
+            label = hasModelNameError
+              ? `${label} + 1 issue`
+              : Label.REQUIRED_CREDENTIAL_ERROR;
           }
         }
-        const previousIcon = hasModelNameError ? "error" : "success";
+        const previousIcon =
+          hasModelNameError || hasCredentialError ? "error" : "success";
```
huwshimi: I'm not sure this is simpler, but this could incrementally build a list of errors, which would make it easier if more errors are added.

```suggestion
        let label = undefined;
        const labels = [];
        if (index === 0) {
          if (errors.modelName) {
            labels.push(
              values.modelName !== ""
                ? Label.INCORRECT_MODEL_NAME_ERROR
                : Label.REQUIRED_MODEL_NAME_ERROR,
            );
          }
          if (errors.credential) {
            labels.push(Label.REQUIRED_CREDENTIAL_ERROR);
          }
          if (labels.length > 0) {
            [label] = labels;
            if (labels.length > 1) {
              label = `${label} + ${labels.length - 1} issue`;
            }
          }
        }
        const previousIcon = label ? "error" : "success";
```

src/pages/AddModel/AddModel.tsx
```diff
@@ -38,13 +38,24 @@ import {
...
   modelName: Yup.string()
     .matches(MODEL_NAME_PATTERN, Label.INCORRECT_MODEL_NAME_ERROR)
     .required("Required"),
+  credential: Yup.string().test({
+    test: function (value) {
+      if (value !== undefined && value !== "") {
+        return true;
+      }
+      return this.createError({
+        message: getCredentialError(this.parent.cloud),
```
huwshimi, line 55: Is there any way it can know the parent type so we don't have to check the cloud type in the util?

huwshimi (commented):
One suggestion and one question. Also I've replied on mattermost, but the credential empty select can be validated on mount using the initalTouched object.

src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
Ninfa-Jeon: Oh, yes! this is perfect. thanks!

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: ~~I didn't know about the typing when I implemented this but looks like there is something that we can use. One interesting thing to note, although the naming says `this.parent` it is actually only possible to access sibling level fields.~~
Correction: none of the hacks work so I'm type casting the `this.parent` in my next push.

src/pages/AddModel/AddModelStepper/AddModelStepper.tsx
Ninfa-Jeon: done

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: as discussed, I have made no changes for this.

src/pages/AddModel/AddModel.tsx
```diff
@@ -171,6 +171,7 @@ const AddModel: FC = () => {
               ...getConfigInitialValues(CONFIG_CATEGORIES),
             }}
             validationSchema={validationSchema}
+            initialTouched={{ credential: true }}
```
huwshimi, line 175: I think it would be helpful to leave a comment about why this is needed.

huwshimi (approved):
Thanks, one tiny comment about a comment.

src/pages/AddModel/AddModel.tsx
Ninfa-Jeon: done, thanks!

PR #2340, opened 2026-06-09
fix: broken mandatory details screen

Ninfa-Jeon:
## Done

- Fixed broken mandatory details page on safari (and probably on other browsers too)
- Improved css rules for mandatory details

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Open the AddModel page in safari browser
- The mandatory details tab should display correctly

## Details
The `position: fixed;` in `_add-model.scss` that we are using to correct dynamic heights moved the entire page out of the hierarchy forcing the page to take as much width as its children.
I originally used `width: stretch;` to make this work, however, this property doesn't work on all browsers.

## Screenshots

BEFORE
Screenshot 2026-06-09 at 7 32 15 PM

AFTER
Screenshot 2026-06-09 at 7 32 33 PM

### DIFF
--- a/src/layout/MainContent/MainContent.tsx
+++ b/src/layout/MainContent/MainContent.tsx
@@ -28,28 +28,26 @@ const MainContent: FC<Props> = ({
         "l-main__content--has-secondary-nav": hasSecondaryNav,
       })}
     >
-      <>
-        {hasSecondaryNav && !loading ? (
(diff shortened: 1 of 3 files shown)


huwshimi (approved):
I wasn't able to reproduce the original issue, but this version also works, so if it fixes it for  you then that's great.

PR #2342, opened 2026-06-10
fix: add missing window titles

huwshimi:
## Done

- Added window titles to pages where they were missing (e.g. the add model form).

## QA

- Go to the add model form and the window title should be "Add model"
- Refresh the page and the window title should show "Add model" again (previously refreshing it would show "404").


Ninfa-Jeon (approved):
All good, Thanks!

PR #2344, opened 2026-06-11
feat: implement `show more` for disable command

Ninfa-Jeon:
## Done

- Implemented `TruncatedCommands.tsx` to either show the entire comma-separated list of disabled commands or truncate them with a "show more" button
- This is implemented in a way that the text is truncated after it overflows the device width making it dynamic
- It follows similar pattern to `TruncatedTooltip.tsx` which is why I extracted common logic into a custom hook `useTruncated`
- Added tests

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View the AddModel -> ConfigsConstraints page on different screen resolutions
- See that the "show more" button only appears when the list overflows screen and is truncated

## Details

https://warthogs.atlassian.net/browse/JUJU-9834

## Screenshots

https://github.com/user-attachments/assets/8afcab2b-fa26-4582-85df-fa626a4b0440

### DIFF
--- a/src/hooks/useTruncated.test.tsx
+++ b/src/hooks/useTruncated.test.tsx
@@ -1,82 +1,72 @@
-import { render, screen } from "@testing-library/react";
-import type { JSX } from "react";
+import { act, renderHook } from "@testing-library/react";
+import type { MockInstance } from "vitest";
 
-import useTruncated from "./useTruncated";
-
-const Harness = ({ enabled = true }): JSX.Element => {
-  const { ref, truncated } = useTruncated(enabled);
-
-  return (
-    <>
-      <span ref={ref}>Some content</span>
-      <span data-testid="truncated-state">{truncated.toString()}</span>
-    </>
-  );
-};
+import useTruncated, { checkTruncated } from "./useTruncated";
 
 describe("useTruncated", () => {
-  const offsetWidth = Object.getOwnPropertyDescriptor(
-    HTMLElement.prototype,
-    "offsetWidth",
-  );
-  const scrollWidth = Object.getOwnPropertyDescriptor(
-    HTMLElement.prototype,
-    "scrollWidth",
-  );
   const resizeObserver = globalThis.ResizeObserver;
+  let observe: MockInstance;
 
   beforeEach(() => {
+    observe = vi.fn();
+
     globalThis.ResizeObserver = vi.fn(() => {
       return {
-        observe: vi.fn(),
+        observe,
         unobserve: vi.fn(),
         disconnect: vi.fn(),
       } as unknown as ResizeObserver;
     }) as unknown as typeof ResizeObserver;
-
-    Object.defineProperty(HTMLElement.prototype, "offsetWidth", {
-      configurable: true,
-      value: 1000,
-    });
-    Object.defineProperty(HTMLElement.prototype, "scrollWidth", {
-      configurable: true,
-      value: 1000,
-    });
   });
 
   afterEach(() => {
     globalThis.ResizeObserver = resizeObserver;
-
-    if (offsetWidth) {
-      Object.defineProperty(HTMLElement.prototype, "offsetWidth", offsetWidth);
-    }
-    if (scrollWidth) {
-      Object.defineProperty(HTMLElement.prototype, "scrollWidth", scrollWidth);
-    }
   });
 
-  it("returns false when the content fits", () => {
-    render(<Harness />);
-    expect(screen.getByTestId("truncated-state")).toHaveTextContent("false");
-  });
+  describe("checkTruncated", () => {
+    it("returns false when the content fits", () => {
+      const element = {
+        offsetWidth: 1000,
+        scrollWidth: 1000,
+      } as HTMLElement;
 
-  it("returns true when the content is truncated", () => {
-    Object.defineProperty(HTMLElement.prototype, "offsetWidth", {
-      configurable: true,
-      value: 200,
+      expect(checkTruncated(element)).toBe(false);
     });
 
-    render(<Harness />);
-    expect(screen.getByTestId("truncated-state")).toHaveTextContent("true");
+    it("returns true when the content is truncated", () => {
+      const element = {
+        offsetWidth: 200,
+        scrollWidth: 1000,
+      } as HTMLElement;
+
+      expect(checkTruncated(element)).toBe(true);
+    });
+  });
+
+  it("sets truncated state from the observed element when enabled", () => {
+    const element = {
+      offsetWidth: 200,
+      scrollWidth: 1000,
+    } as HTMLElement;
+
+    const { result, rerender } = renderHook(
+      ({ enabled }) => useTruncated(enabled),
+      { initialProps: { enabled: false } },
+    );
+
+    act(() => {
+      result.current.ref.current = element;
+    });
+    rerender({ enabled: true });
+
+    expect(result.current.truncated).toBe(true);
+    expect(observe).toHaveBeenCalledTimes(1);
+    expect(observe).toHaveBeenCalledWith(element);
   });
 
   it("does not observe when disabled", () => {
-    Object.defineProperty(HTMLElement.prototype, "offsetWidth", {
-      configurable: true,
-      value: 200,
-    });
-
-    render(<Harness enabled={false} />);
-    expect(screen.getByTestId("truncated-state")).toHaveTextContent("false");
+    const { result } = renderHook(() => useTruncated(false));
+    expect(result.current.truncated).toBe(false);
+    expect(observe).not.toHaveBeenCalled();
   });
 });
--- a/src/hooks/useTruncated.test.tsx
+++ b/src/hooks/useTruncated.test.tsx
@@ -1,0 +1,82 @@
+import { render, screen } from "@testing-library/react";
+import type { JSX } from "react";
+
+import useTruncated from "./useTruncated";
+
+const Harness = ({ enabled = true }): JSX.Element => {
+  const { ref, truncated } = useTruncated(enabled);
+
+  return (
+    <>
+      <span ref={ref}>Some content</span>
+      <span data-testid="truncated-state">{truncated.toString()}</span>
+    </>
+  );
+};
+
+describe("useTruncated", () => {
+  const offsetWidth = Object.getOwnPropertyDescriptor(
+    HTMLElement.prototype,
+    "offsetWidth",
+  );
+  const scrollWidth = Object.getOwnPropertyDescriptor(
+    HTMLElement.prototype,
+    "scrollWidth",
+  );
+  const resizeObserver = globalThis.ResizeObserver;
+
+  beforeEach(() => {
+    globalThis.ResizeObserver = vi.fn(() => {
+      return {
+        observe: vi.fn(),
+        unobserve: vi.fn(),
+        disconnect: vi.fn(),
+      } as unknown as ResizeObserver;
+    }) as unknown as typeof ResizeObserver;
+
+    Object.defineProperty(HTMLElement.prototype, "offsetWidth", {
+      configurable: true,
+      value: 1000,
+    });
+    Object.defineProperty(HTMLElement.prototype, "scrollWidth", {
+      configurable: true,
+      value: 1000,
+    });
+  });
+
+  afterEach(() => {
+    globalThis.ResizeObserver = resizeObserver;
+
+    if (offsetWidth) {
+      Object.defineProperty(HTMLElement.prototype, "offsetWidth", offsetWidth);
+    }
+    if (scrollWidth) {
+      Object.defineProperty(HTMLElement.prototype, "scrollWidth", scrollWidth);
+    }
+  });
+
+  it("returns false when the content fits", () => {
+    render(<Harness />);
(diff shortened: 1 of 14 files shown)


Ninfa-Jeon:
I have not implemented the `show x more` functionality as calculating the number of items would be a challenge in a dynamically truncating piece of text.

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.tsx
huwshimi, line 1: This has ended up similar enough to TruncatedTooltip that I think I could be worth sharing the logic. It would probably be easiest to do that via a custom hook, but could also be done with a shared component.

huwshimi (commented):
A couple of things in the gif below:
- For some reason when you click on the text when it is truncated then the ellipsis disappears and the text just gets cut off.
- When the text is expanded then the "Show less" button has a background.

Image

Also on small screens the show more link is indented:

Image

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/_truncated-details.scss
```diff
@@ -0,0 +1,15 @@
+.truncated-details {
+  &__content {
+    vertical-align: bottom;
+  }
+
+  &__is-collapsed {
+    display: inline-block;
+    max-width: 90%;
```
huwshimi: Using a width is a bit fragile as it relies on a fixed text size, so if the text changes, the font isn't available or the user has changed their browser font size this % would also need to change.

Instead, you could remove these styles and instead, when the content is collapsed set `truncated-details` to display as `flex` and on the show more button set `text-wrap: nowrap`.

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/TruncatedDetails.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }, [checkTruncated, expanded, resizeObserver]);
+
+  return (
+    <div className="truncated-details">
+      <span
+        id={contentId}
+        ref={truncatedNode}
+        className={classNames("truncated-details__content", {
+          "truncated-details__is-collapsed": !expanded,
+          "truncated-details__is-expanded": expanded,
```
huwshimi: For BEM these states should use [modifiers](https://getbem.com/naming/).

```suggestion
          "truncated-details--is-collapsed": !expanded,
          "truncated-details--is-expanded": expanded,
```

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.tsx
Ninfa-Jeon: Done. I have extracted the common logic to a custom hook

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/_truncated-details.scss
Ninfa-Jeon: Done.

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/TruncatedDetails.tsx
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/_truncated-details.scss
```diff
@@ -0,0 +1,18 @@
+.truncated-details {
+  &__content {
+    vertical-align: bottom;
+  }
+
+  &__toggle {
+    background-color: $color-x-light !important;
```
huwshimi: I think if you increase the class specificity you won't need `!important`. You might just be able to add `[aria-expanded=true]` to the class (or it might need `button[aria-expanded=true]`)

huwshimi (commented):
I'm still seeing that strange issue where it stops showing the ellipsis when you click on the text (I'm using Chrome):

Image

src/pages/AddModel/ConfigsConstraints/TruncatedDetails/_truncated-details.scss
Ninfa-Jeon: I have fixed this but I am simply unable to see the issue with the ellipsis vanishing

andogq:
When on very small screens (<620px wide), the header nav on the top pushes the cancel/back/next/create buttons at the bottom beyond the end of the screen
Screen Shot 2026-06-16 at 13 48 44

andogq:
> I'm still seeing that strange issue where it stops showing the ellipsis when you click on the text (I'm using Chrome):
> Image

After a bit of digging, this is actually a (severely under-documented) Chrome accessibility feature added earlier this year: https://chromestatus.com/feature/5146265241387008

Essentially, the engine will change `text-overflow` from `ellipsis` to `clip` temporarily when the user interacts with the overflowed element. Why this isn't reflected in the inspector is beyond me, but that's the reason for it at least.

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.tsx
```diff
@@ -0,0 +1,54 @@
...
+    >
+      <span
+        id={contentId}
+        ref={truncatedNode}
+        className={classNames("truncated-details__content", {
+          "u-truncate": !expanded,
+          "u-sh1--right": expanded,
+        })}
+      >
+        <b>Disables commands:</b> {content}
```
andogq, line 35: Doesn't quite feel right for `Disabled commands` to be hardcoded in a component called `TruncatedDetails`, it isn't obvious that the component is only for lists of commands.

I think either move that text into another prompt, or rename the component.

src/hooks/useTruncated.ts
```diff
@@ -0,0 +1,44 @@
...
+    checkTruncated();
+    // Watch the content for resizes to check if the truncation changes.
+    resizeObserver.observe(element);
+    return (): void => {
+      resizeObserver.unobserve(element);
+    };
+  }, [checkTruncated, enabled, resizeObserver]);
+
+  return { ref, truncated };
+}
```
andogq, line 44: It'll need some tweaking (just did this in the GitHub comment, haven't tested it), but it'd be good to see the `checkTruncated` logic hoisted/decoupled from the React-specific stuff. Then we could test the logic directly, and leave the state-wrangling to the hook.

```suggestion
function checkTruncated<T extends HTMLElement>(element: T): boolean {
  // Check to see if the content is larger than the frame, in which case the
  // CSS will be truncating the element.
  return element.offsetWidth < element.scrollWidth;
}

/**
 * Hook that tracks whether a DOM element is visually truncated.
 * @param enabled - Whether truncation checks should actively run.
 * @returns The element ref and whether its content is truncated.
 */
export default function useTruncated<T extends HTMLElement>(
  enabled = true,
): { ref: RefObject<null | T>; truncated: boolean } {
  const ref = useRef<T>(null);
  const [truncated, setTruncated] = useState(false);

  const resizeObserver = useMemo(
    () => new ResizeObserver(() => {
      setTruncated(
        ref.current
          ? checkTruncated(ref.current)
          : false
      );
    }),
    [],
  );

  useEffect(() => {
    const element = ref.current;
    if (!element || !enabled) {
      return;
    }
    // Do an initial check for whether the content is truncated.
    setTruncated(checkTruncated(element));
    // Watch the content for resizes to check if the truncation changes.
    resizeObserver.observe(element);
    return (): void => {
      resizeObserver.unobserve(element);
    };
  }, [enabled, resizeObserver]);

  return { ref, truncated };
}
```

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.test.tsx
andogq, line 1: I don't super love the prototype hacks in these tests, it feels a bit fragile and inconsistent. If possible, I think there should be a spy on the `useTruncated` hook in order to get access to the `ref` it returns, so it can be manipulated directly.

src/hooks/useTruncated.test.tsx
andogq, line 1: Same deal here, I'd prefer to see usage of `ref` instead of overwriting the prototype for the tests.

andogq (commented):
Seems good overall, just some changes mainly in the tests.

I haven't left it as a comment anywhere, but I'm a little unsure of the aria attributes for this kind of UI, as the overflowed items are still within the DOM, and are fully accessible to a screen reader. The screen reader only finds the show more/less button after traversing the overflowed content, so potentially a focus trap could be used to improve this? Not a massive deal though.

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.test.tsx
Ninfa-Jeon: I think an even better test would just mock the return value of `useTruncated` since the hook itself is tested elsewhere and the component is just dependent on the output of the hook, it makes sense to mock the output altogether, simplifying this by a lot.

src/hooks/useTruncated.test.tsx
Ninfa-Jeon: done

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.test.tsx
Ninfa-Jeon: Done as stated above.

src/hooks/useTruncated.ts
Ninfa-Jeon: Done

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.tsx
Ninfa-Jeon: renamed to `TruncatedCommands`

Ninfa-Jeon:
> I'm a little unsure of the aria attributes for this kind of UI, as the overflowed items are still within the DOM, and are fully accessible to a screen reader. The screen reader only finds the show more/less button after traversing the overflowed content, so potentially a focus trap could be used to improve this? Not a massive deal though.

A focus trap would be the wrong tool here. Focus traps are for modal or composite interactions where you must contain keyboard focus; for an inline “show more/less” control they usually make accessibility worse, not better. - is what copilot suggests.
My two cents on this would be, if the screen reader is already going through the entire list then the button is a visual-only affordance, which isn't that big of an issue.

Ninfa-Jeon:
> When on very small screens (<620px wide), the header nav on the top pushes the cancel/back/next/create buttons at the bottom beyond the end of the screen

Bug ticket here: https://github.com/canonical/juju-dashboard/issues/2348
And PR with the fix here: https://github.com/canonical/juju-dashboard/pull/2349

andogq:
> A focus trap would be the wrong tool here. Focus traps are for modal or composite interactions where you must contain keyboard focus; for an inline “show more/less” control they usually make accessibility worse, not better. - is what copilot suggests. My two cents on this would be, if the screen reader is already going through the entire list then the button is a visual-only affordance, which isn't that big of an issue.

Ah, focus trap may've been the wrong term. It may be advantageous for a screen-reader to be presented with an option to descend into the list of disabled commands, or be shown the more/less button, rather than it needing to go through the entire list before reaching any point of interactivity. If you check it out with an accessibility inspector, you'll see what I mean.

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    vi.restoreAllMocks();
+  });
+
+  it("does not render a toggle when the text fits", () => {
+    render(
+      <TruncatedCommands items={["destroy-controller", "destroy-model"]} />,
+    );
+
+    expect(
+      screen.queryByRole("button", { name: "Show more" }),
```
andogq: Might be good to test that `Show less` is also not present, or just a button with class `truncated-commands__toggle`.

src/hooks/useTruncated.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }) as unknown as typeof ResizeObserver;
+  });
+
+  afterEach(() => {
+    globalThis.ResizeObserver = resizeObserver;
+  });
+
+  describe("checkTruncated", () => {
+    it("returns false when the content fits", () => {
+      const element = document.createElement("span");
```
andogq: I don't think it's necessary to create a full element here. Since you're mocking the resize observer, it should be fine to just use a standard object here, with the properties that you need.

andogq (approved):
Just two tweaks to the tests, then it's good to merge! I'll approve it, but if you could address those things then feel free to merge

src/pages/AddModel/ConfigsConstraints/TruncatedCommands/TruncatedCommands.test.tsx
Ninfa-Jeon: using a testid instead

PR #2346, opened 2026-06-15
test: add e2e tests for add-model flow

Ninfa-Jeon:
## Done

- Added end-to-end tests for the following four cases:
  - a simple test to see that a model can be added (no interactions besides the mandatory details tab)
  - a test to confirm that any configs+constraints set during add-model are reflected
  - a test to confirm that interacting with disable commands works
  - a test to check that sharing the model via access management tab and lowering the model owner's access works 
- Updated the playwright config to only store video artifacts and screenshots for failing tests.
- Bumped `jimm-version` in workflows to use the changes from [this](https://github.com/canonical/jimm/pull/1979) PR
- Updated the value being sent for `qualifier` for 4.0 workflows -> they expect the user ID without the `user-` prefix

[Successful workflows](https://github.com/canonical/juju-dashboard/actions/runs/27614235845/job/81646103136?pr=2346) (without noise from other tests)

## QA

- Pull code
- Open http://0.0.0.0:8036/
- All e2e tests pass

## Details

https://warthogs.atlassian.net/browse/JUJU-9835


e2e/base/add-model.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+        .locator("tr", { hasText: currentModel.name })
+        .and(page.locator("tr", { hasText: currentModel.owner.displayName })),
+    ).toBeInViewport();
+
+    // Verify that the model has the correct configs and constraints applied
+    await owner.cliLogin(browser);
+    await exec(`juju switch '${currentModel.qualifiedName}'`);
+    await expect
+      .poll(async () => {
+        const { stdout } = await exec("juju model-config");
```
andogq: If you pass the `--format=json` flag, you can just parse the JSON output and directly access the `arch` and `defaultSpace` key. Might be a bit more robust than using regex.

e2e/base/add-model.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  test("shares the model during add-model and downgrades the owner's access", async ({
+    browser,
+    page,
+  }) => {
+    await owner.dashboardLogin(page, urls.models.addModel);
+
+    // Fill in the mandatory details and go to Access Management step
+    await page.locator('input[name="modelName"]').fill(currentModel.name);
+    await page.getByText("Access Management (optional)").click();
```
andogq: We should be using labels for this kind of text

andogq (approved):
Nice work, just two small changes

e2e/base/add-model.spec.ts
Ninfa-Jeon: done

e2e/base/add-model.spec.ts
Ninfa-Jeon: done

PR #2356, opened 2026-06-17
chore: upgrade `vitest` and `@vitest/coverage-v8` to 4.1.9

andogq:


src/store/middleware/process/model-upgrade/upgrade-to.test.ts
```diff
@@ -79,8 +79,9 @@ describe("upgradeTo", () => {
     });
 
     it("throws an error if request isn't successful", async ({ expect }) => {
-      expect.assertions(3);
```
Ninfa-Jeon, line 82: i suppose this line isn't needed anymore?

Ninfa-Jeon (approved):
Nice work, just one small question about a comment.

src/store/middleware/process/model-upgrade/upgrade-to.test.ts
andogq: Just a leftover from my debugging, whoops!

PR #2375, opened 2026-06-25
feat: source middleware

andogq:
## Done

Merge source implementations to use a single middleware instance. This reduces the size of the callback chain for every action dispatched via redux.

## QA

- Ensure dashboard works as expected
  - Model list
  - Model details
  - Model migration

## Details

JUJU-9491

### DIFF
--- a/src/store/middleware/source-middleware.test.ts
+++ b/src/store/middleware/source-middleware.test.ts
@@ -2,7 +2,11 @@ import type { PayloadAction } from "@reduxjs/toolkit";
 import { createAction } from "@reduxjs/toolkit";
 import type { Mock } from "vitest";
 
-import { createSourceMiddleware, SourceManager } from "./source-middleware";
+import type { SourceInstance } from "./source-middleware";
+import createMiddleware, {
+  createSourceInstance,
+  SourceManager,
+} from "./source-middleware";
 
 describe("SourceManager", () => {
   describe("start", () => {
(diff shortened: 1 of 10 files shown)


src/store/middleware/source-middleware.test.ts
```diff
@@ -240,150 +217,247 @@ describe("source middleware", () => {
...
+      const { createSource } = instance;
+      instance.createSource = (): ReturnType<typeof createSource> => {
+        const source = createSource();
+
+        for (const action of ["start", "stop", "invalidate"] as const) {
+          spies[action] = vi.spyOn(source, action);
+        }
+
+        return source;
+      };
```
Ninfa-Jeon, line 398: too much whitespace :)


```suggestion
      instance.createSource = (): ReturnType<typeof createSource> => {
        const source = createSource();
        for (const action of ["start", "stop", "invalidate"] as const) {
          spies[action] = vi.spyOn(source, action);
        }
        return source;
      };
```

src/store/middleware/source/index.ts
```diff
@@ -1,17 +1,26 @@
...
+  void,
+  RootState,
+  Store["dispatch"]
+> {
+  return createMiddleware([
+    modelList as SourceInstance<unknown, unknown>,
+    jimmSupportedVersions as SourceInstance<unknown, unknown>,
+    migrationTargets as SourceInstance<unknown, unknown>,
+    userCredentials as SourceInstance<unknown, unknown>,
+    cloudInfo as SourceInstance<unknown, unknown>,
```
Ninfa-Jeon, line 24: I'm curious if you have any ideas as to how this can have improved typing in the future

Ninfa-Jeon (approved):
All good. One open-ended question and minor nitpick.

src/store/middleware/source/index.ts
andogq: I haven't had a chance to have a proper play around with it, but at this point it seems to be a TS limitation. A good way around it would be to use a builder pattern, where each callback accepts a generic parameter and performs the type erasure before adding it to the collection.

PR #2383, opened 2026-06-26
fix(e2e): pack single charm for e2e tests

Ninfa-Jeon:
## Done

- Updated `build.sh` to accept an optional third argument that is passed directly to `charmcraft pack`
- Updated the charm setup actions to pass `--platform ubuntu@24.04:amd64` to the build script, so e2e workflows only pack the single charm needed for the CI runner.

## Details
- The `charmcraft pack` produces four `.charm` files after [this](https://github.com/canonical/juju-dashboard/pull/2370) change so when the deploy step ran:

    ```
    juju deploy /path/to/juju-dashboard*.charm dashboard
    ```
  the shell expanded the glob to multiple filenames, making juju see ["dashboard"] as an unrecognised extra argument making the e2e tests fail (see [this](https://github.com/canonical/juju-dashboard/actions/runs/28207192235)).

## QA

- Pull code
- Open http://0.0.0.0:8036/
- All charm setup jobs in e2e workflows should pass (except the ones skipped due to failing `Start JIMM` job)

### DIFF
--- a/charms/k8s-charm/build.sh
+++ b/charms/k8s-charm/build.sh
@@ -7,6 +7,7 @@ ROOT_DIR="$SCRIPT_PATH/../../"
 
 BUILD_TYPE="${1:-source}"
 DASHBOARD_RESOURCE="${2:-}"
+PACK_ARGS="${3:-}"
 
 DASHBOARD_IMAGE_ID=""
 
@@ -55,7 +56,7 @@ ATTEMPT=1
 SUCCESS=false
 
 while [ $ATTEMPT -le $MAX_ATTEMPTS ]; do
-    if (cd "$SCRIPT_PATH" && charmcraft pack); then
+    if (cd "$SCRIPT_PATH" && charmcraft pack $PACK_ARGS); then
         SUCCESS=true
         break # Success, exit retry loop
     else
--- a/charms/k8s-charm/build.sh
+++ b/charms/k8s-charm/build.sh
@@ -7,7 +7,7 @@ ROOT_DIR="$SCRIPT_PATH/../../"
 
 BUILD_TYPE="${1:-source}"
 DASHBOARD_RESOURCE="${2:-}"
-PACK_ARGS="${3:-}"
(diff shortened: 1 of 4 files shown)


charms/k8s-charm/build.sh
```diff
@@ -7,6 +7,7 @@ ROOT_DIR="$SCRIPT_PATH/../../"
 
 BUILD_TYPE="${1:-source}"
 DASHBOARD_RESOURCE="${2:-}"
+ALL_PLATFORMS="${3:-}"
```
huwshimi: I wonder if it might be better to reverse this logic. I think I'd expect that if I ran `./build.sh` it would build the platforms defined in charmcraft.yaml.

The third param could take a string that gets passed as PACK_ARGS below, so then in the two charm actions you could do something like 

```
      run: |
        "${{ github.workspace }}/charms/k8s-charm/build.sh" \
          "${{ env.build_type }}" \
          "${{ inputs.dashboard-resource }}" \
          "--platform ubuntu@24.04:amd64"
          >> "$GITHUB_OUTPUT"
```

Ninfa-Jeon: Good idea...done!

charms/k8s-charm/build.sh
```diff
@@ -55,7 +56,7 @@ ATTEMPT=1
 SUCCESS=false
 
 while [ $ATTEMPT -le $MAX_ATTEMPTS ]; do
-    if (cd "$SCRIPT_PATH" && charmcraft pack); then
+    if (cd "$SCRIPT_PATH" && charmcraft pack $PACK_ARGS); then
```
huwshimi: Should these be quoted too prevent execution? That might require a condition so that an empty string isn't supplied as a param.

Ninfa-Jeon: ive pushed with something that should do both

huwshimi (approved):
Thanks for the fix. It worked for me locally as well.

PR #2384, opened 2026-06-30
chore: remove number to string cast

Ninfa-Jeon:
## Done

- Removed `number` to `string` casting and coercion in the add model flow
- Handle `undefined` defaults gracefully without `""` fallbacks in data model

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Try fiddling with the numeric fields on the ConfigsConstraints tab on AddModel page
- All should work fine

### DIFF
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
@@ -7,6 +7,7 @@ import { externalURLs } from "urls";
 
 import { InputMode } from "../../types";
 import { CONFIG_CATEGORIES } from "../configCatalog";
+import { CONSTRAINT_CATEGORIES } from "../constraintsCatalog";
 import { FieldName, Label } from "../types";
 
 import CategoriesListing from "./CategoriesListing";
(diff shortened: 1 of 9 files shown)


src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -81,10 +80,10 @@ export const buildConfigsConstraintsPayload = (
 
 export const getConfigInitialValues = (
   categories: CategoryDefinition[],
-): Record<string, string> =>
-  categories.reduce<Record<string, string>>((values, category) => {
+): Record<string, number | string> =>
+  categories.reduce<Record<string, number | string>>((values, category) => {
     category.fields.forEach((field) => {
-      values[field.label] = field.defaultValue?.toString() ?? "";
+      values[field.label] = field.defaultValue ?? "";
```
andogq: Is an empty string the right thing to use as a default value for non-string values?

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -18,24 +17,24 @@ export const isConfigChanged = (
   // If there's a default value, check if current differs from it
   // This includes catching empty string ("") vs default
   if (defaultValue !== undefined) {
-    return currentValue !== defaultValue.toString();
+    return currentValue !== defaultValue;
   }
 
   // No default value: only changed if non-empty
-  return currentValue.length > 0;
+  return currentValue !== "";
```
andogq, line 24: Wouldn't comparing to an empty string only work if the current value is a string?

andogq (commented):
Nice and simple change!

One small thing, it'd be good to see some tests added to this, to make sure that numeric fields are handled in the form: https://github.com/canonical/juju-dashboard/blob/main/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: it makes comparison easier for when we are checking for changes. there isn't really anything we can use for numeric and boolean values

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: we would get to this point only when there is no default value for this field, in which case, we are setting the initial value to `""` and comparing this way then makes sense.

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: I'd be concerned about a number or boolean being set to an empty string however. Would it be better to leave it as undefined in those cases, or maybe 0 or false?

Ninfa-Jeon: i think i was using 0 for numeric fields which in practice might be wrong because a user might set a numeric value to 0 explicitly which would be a change...i can try keeping the numeric fields undefined and i am already setting the numeric constraints to an empty third option because i dont have defaults for constraints

Ninfa-Jeon: Ahh, i remember why I didn't go with `undefined` defaults in the first place. Setting formik field initial values to `undefined` does not originally track the field and then the user actually setting this value throws an error in the console. Changing to `undefined` would cause React's uncontrolled→controlled warning for any text/number input the user touches.

andogq: That's fine, we could just leave the field as uncontrolled if that's a problem. 
I do agree with your point that we can't just use 0 as a default for numeric fields, so `undefined` is reasonable given that it's the truth

src/pages/AddModel/ConfigsConstraints/utils.ts
andogq: I get that it'd technically work in that case, but just feels a bit weird that it'd sometimes be comparing a string with a number. If you don't find another way to handle it tho, then that's ok

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
```diff
@@ -318,4 +319,88 @@ describe("CategoriesListing", () => {
...
+      const cpuPowerInput = screen.getByLabelText("cpu-power");
+      expect(cpuPowerInput).toHaveAttribute("type", "number");
+    });
+
+    it("treats a filled numeric field as changed and shows it in changed-only view", async () => {
+      render(
+        <Formik
+          initialValues={{
+            [FieldName.CONSTRAINT_INPUT_MODE]: InputMode.LIST,
+            cores: "4",
```
andogq: This should be a number in initial values

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
```diff
@@ -318,4 +319,88 @@ describe("CategoriesListing", () => {
...
+        yamlPlaceholder: Label.MODEL_CONSTRAINT_PLACEHOLDER,
+        searchName: "constraintSearch",
+      };
+    });
+
+    it("renders numeric fields correctly", () => {
+      render(
+        <Formik
+          initialValues={{ [FieldName.CONSTRAINT_INPUT_MODE]: InputMode.LIST }}
+          onSubmit={vi.fn()}
```
andogq: Will be good to test that submitted values come back as a number too

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
```diff
@@ -318,4 +319,88 @@ describe("CategoriesListing", () => {
       screen.getByLabelText("container-networking-method"),
     ).toBeInTheDocument();
   });
+
+  describe("numeric fields", () => {
+    let constraintProps: CategoriesListingProps;
```
andogq: Why is this being redefined, can we just use defaultProps that's used in other test cases?

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
```diff
@@ -318,4 +319,88 @@ describe("CategoriesListing", () => {
...
+      // Non-numeric, unchanged field should be hidden
+      expect(screen.queryByLabelText("spaces")).not.toBeInTheDocument();
+    });
+
+    it("includes numeric field values in YAML output", async () => {
+      render(
+        <Formik
+          initialValues={{
+            [FieldName.CONSTRAINT_INPUT_MODE]: InputMode.LIST,
+            cores: "8",
```
andogq: Should be a number too

andogq (commented):
Nearly there, nice work! Just a tweak on the test and dealing with that controlled field we discussed.

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
Ninfa-Jeon: fixed

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
Ninfa-Jeon: ive added one such test under AddModel.test.tsx

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: i have been going in circles for this and the undefined value...i cant think of anything for this case unfortunately...maybe im having a block in my logic...let me know if you think of something better!

src/pages/AddModel/ConfigsConstraints/utils.ts
Ninfa-Jeon: as discussed, i have moved the empty string fallback to the initialValues setting and the logic everywhere else handles both numeric and undefined values

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
andogq, line 1: Much more thorough tests, nice work!

andogq (approved):
These changes are looking good, nice work!

PR #2387, opened 2026-07-03
feat: add semver and controller version gate utilities

Ninfa-Jeon:
## Done

- Added selector `getControllerSupportsModelConfigSchema` that evaluates the controller for its version and if it is above the minimum version to support `ModelConfigSchema` facade method
- Extended `isHigherSemver` to support checking for "higher than or equal to: cases
- Added tests

## QA

- Pull code
- Open http://0.0.0.0:8036/
- All tests should pass

## Details

https://warthogs.atlassian.net/browse/JUJU-9831

### DIFF
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -11,6 +11,9 @@ export const OIDC_POLL_INTERVAL = 5 * 60 * 1000;
 // The local storage key for enabled feature flags list
 export const ENABLED_FLAGS = "flags";
 
+// The minimum Juju version that supports the modelConfigSchema facade method.
+export const MODEL_CONFIG_SCHEMA_MIN_VERSION = "4.0.12";
+
 // The options to build a boolean Select
 export const BOOLEAN_OPTIONS = [
   { label: "True", value: "true" },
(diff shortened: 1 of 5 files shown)


huwshimi (commented):
Rather than checking the controller version, it might be better to check the facade version (`>= 8`), or that the method exists on the facade. This would make it more reliable if the feature doesn't end up being released in 4.0.12 (it's currently only a draft release) and would also allow it to work if the feature was backported to older Juju releases.

Ninfa-Jeon:
> Rather than checking the controller version, it might be better to check the facade version (>= 8), or that the method exists on the facade. This would make it more reliable if the feature doesn't end up being released in 4.0.12 (it's currently only a draft release) and would also allow it to work if the feature was backported to older Juju releases.

Oh, with this idea, I don't think a PR is needed at all since we can directly check within the middleware instance (which I will write in a later PR). Something like:
```
typeof connection.facades.cloud?.modelConfigSchema === "function"
```
Closing this PR in favor of that. Thanks!

PR #2389, opened 2026-07-07
feat(utils): add normalisation utils

Ninfa-Jeon:
## Done
- Bumped `@canonical/jujulib` to latest version
- Added two utilities:
  - `resolveFieldDefault` - which resolves a field's default value based on the hierarchy `controller > region (if known) > juju default`
  - `generateCategoryDefinitions` - which takes both the schema and the defaults returned by the facade methods (yet to introduce) and normalises their responses into a single `CategoryDefinition[]` which can be used by the UI allowing for easier switching between static and dynamic states.

## Details
https://warthogs.atlassian.net/browse/JUJU-10097

### DIFF
--- a/src/store/middleware/source/util.test.ts
+++ b/src/store/middleware/source/util.test.ts
@@ -1,0 +1,144 @@
+import type { ModelDefaultsResult } from "@canonical/jujulib/dist/api/facades/model-manager/ModelManagerV11";
+
+import { BOOLEAN_OPTIONS, FALLBACK_CATEGORY } from "consts";
+import { modelConfigSchemaFieldFactory } from "testing/factories/juju/CloudV8";
+
+import { generateCategoryDefinitions } from "./util";
+
+describe("generateCategoryDefinitions", () => {
+  it("returns a single category with the FALLBACK_CATEGORY sentinel", () => {
+    const result = generateCategoryDefinitions({}, {});
+    expect(result).toHaveLength(1);
+    expect(result[0].category).toBe(FALLBACK_CATEGORY);
+  });
+
+  it("returns an empty fields array for an empty schema", () => {
+    const result = generateCategoryDefinitions({}, {});
+    expect(result[0].fields).toHaveLength(0);
+  });
+
+  it("handles multiple fields in the schema", () => {
+    const config: ModelDefaultsResult["config"] = {
+      "default-space": { default: { value: "my-space" } },
+    };
+    const result = generateCategoryDefinitions(
+      {
+        "default-space": modelConfigSchemaFieldFactory.build({
+          type: "string",
+          description: "The default network space",
+        }),
+        "disable-network-management": modelConfigSchemaFieldFactory.build({
+          type: "bool",
+        }),
+        "net-bond-reconfigure-delay": modelConfigSchemaFieldFactory.build({
+          type: "int",
+        }),
+      },
+      config,
+    );
+    expect(result[0].fields).toHaveLength(3);
+    expect(result[0].fields[0].label).toBe("default-space");
+    expect(result[0].fields[0].description).toBe("The default network space");
+    expect(result[0].fields[0].defaultValue).toBe("my-space");
+    expect(result[0].fields[1].valueType).toBe("boolean");
+    expect(result[0].fields[2].valueType).toBe("number");
+  });
+
+  it("leaves defaultValue undefined when field has no default", () => {
+    const result = generateCategoryDefinitions(
+      { "default-space": modelConfigSchemaFieldFactory.build() },
+      {},
+    );
+    expect(result[0].fields[0].defaultValue).toBeUndefined();
+  });
+
+  it("prefers controller override over other defaults", () => {
+    const config: ModelDefaultsResult["config"] = {
+      "default-space": {
+        default: { value: "default-space" },
+        controller: { value: "controller-space" },
+        regions: [
+          { "region-name": "us-east-1", value: { value: "us-east-space" } },
+        ],
+      },
+    };
+    const result = generateCategoryDefinitions(
+      { "default-space": modelConfigSchemaFieldFactory.build() },
+      config,
+    );
+    expect(result[0].fields[0].defaultValue).toBe("controller-space");
+  });
+
+  it("applies region default when region is provided and controller defaults are absent", () => {
+    const config: ModelDefaultsResult["config"] = {
+      "default-space": {
+        default: { value: "global-default" },
+        regions: [
+          { "region-name": "us-east-1", value: { value: "us-east-space" } },
+        ],
+      },
+    };
+    const result = generateCategoryDefinitions(
+      { "default-space": modelConfigSchemaFieldFactory.build() },
+      config,
+      "us-east-1",
+    );
+    expect(result[0].fields[0].defaultValue).toBe("us-east-space");
+  });
+
+  it("falls back to juju-wide default when other defaults are absent", () => {
+    const result = generateCategoryDefinitions(
+      { "default-space": modelConfigSchemaFieldFactory.build() },
+      {
+        "default-space": {
+          default: { value: "global-default" },
+          regions: [
+            { "region-name": "us-east-1", value: { value: "us-east-space" } },
+          ],
+        },
+      },
+    );
+    expect(result[0].fields[0].defaultValue).toBe("global-default");
+  });
+
+  it("handles bool type properly", () => {
+    const result = generateCategoryDefinitions(
+      {
+        "disable-network-management": modelConfigSchemaFieldFactory.build({
+          type: "bool",
+        }),
+      },
+      {},
+    );
+    expect(result[0].fields[0].valueType).toBe("boolean");
+    expect(result[0].fields[0].input?.type).toBe("select");
+    expect(result[0].fields[0].input?.options).toEqual(BOOLEAN_OPTIONS);
+  });
+
+  it.each(["int", "float"])("handles %s type properly", (type) => {
+    const result = generateCategoryDefinitions(
+      { "some-field": modelConfigSchemaFieldFactory.build({ type }) },
+      {},
+    );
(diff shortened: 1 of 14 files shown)


src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+  ConfigFieldValue,
+} from "pages/AddModel/ConfigsConstraints/types";
+
+/**
+  AdditionalProperties is a single-key object where the key name is
+  arbitrary — the actual value is always the first (and only) entry.
+  Applies the priority hierarchy: controller > region (if known) > juju default.
+*/
+const resolveFieldDefault = (
+  defaults: ModelDefaultsResult["config"][string] = {},
```
huwshimi: It might be a bit more reliable to use the `ValueOf` util type from react-components here, or just import `ModelDefaults` and use that instead.

Also, the type suggests that the object will always be defined, so if that's not the case it would be good to leave a comment here about why the default empty object is required.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+  selectedRegion?: string,
+): ConfigFieldValue => {
+  const { controller, regions, default: jujuDefaults } = defaults;
+
+  if (controller) {
+    return Object.values(controller)[0] as ConfigFieldValue;
```
huwshimi: Because AdditionalProperties values are `any` I don't think these type annotations are needed are they?

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+  schema: ModelConfigSchemaResult["schema"],
+  defaultsConfig: ModelDefaultsResult["config"],
+  selectedRegion?: string,
+): CategoryDefinition[] => {
+  const fields: CategoryDefinition["fields"] = Object.entries(schema ?? {}).map(
+    ([label, field]) => {
+      const isNumeric = field.type === "int" || field.type === "float";
+      const fieldValues = field.values ?? [];
+
+      const entry: CategoryDefinition["fields"][number] = {
```
huwshimi: Because you have to access the array item to get its type is the reason I usually advocate for not having nested object types. Instead you could move the `field` value to a type (e.g. CategoryDefinitionField) and import that here.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+        label,
+        description: field.description ?? "",
```
huwshimi: Is there any reason that `description` can't accept `undefined`? Generally assigning an empty string to get around a type issue is a sign that the type isn't correct.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+        defaultValue: resolveFieldDefault(
+          defaultsConfig[label],
+          selectedRegion,
+        ),
+      };
+
+      if (field.type === "bool") {
+        entry.valueType = "boolean";
```
huwshimi: This and the other `valueType` entries below should probably accept an enum rather than a magic string.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+        entry.input = { type: "select", options: BOOLEAN_OPTIONS };
```
huwshimi: Same here and below where `type` should take an enum.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+      } else if (isNumeric) {
+        entry.valueType = "number";
+      } else if (fieldValues.length > 0) {
+        entry.input = {
+          type: "select",
+          options: fieldValues.map((fieldValue) => ({
+            label: `${fieldValue}`,
```
huwshimi, line 87: `fieldValue` is typed as an object so something doesn't seem right here. If we're expecting the real type to be different we should have some handling to make sure it's the type we expect.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,80 @@
...
+            value: `${fieldValue}`,
+          })),
+        };
+      }
+
+      return entry;
+    },
+  );
+
+  return [{ category: FALLBACK_CATEGORY, fields }];
```
huwshimi: Hard coding the category to an empty string doesn't seem useful. Maybe the type should be updated to accept `null` or `undefined` as the category.

src/store/middleware/source/util.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    expect(result[0].fields[0].valueType).toBe("number");
+    expect(result[0].fields[0].input).toBeUndefined();
+  });
+
+  it("sets a select input with options when values are present", () => {
+    const result = generateCategoryDefinitions(
+      {
+        "firewall-mode": modelConfigSchemaFieldFactory.build({
+          type: "string",
+          values: ["instance", "global", "none"] as unknown as object[],
```
huwshimi, line 135: This could have a comment to explain that the real value types are different from the type provided by Juju.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,87 @@
...
+} from "pages/AddModel/ConfigsConstraints/types";
+import { InputType, ValueType } from "pages/AddModel/ConfigsConstraints/types";
+
+/**
+  AdditionalProperties is a single-key object where the key name is
+  arbitrary — the actual value is always the first (and only) entry.
+  Applies the priority hierarchy: controller > region (if known) > juju default.
+*/
+const resolveFieldDefault = (
+  defaults: ModelDefaults = {}, // `ModelDefaults` has all optional fields, which means the value from the Record lookup could be undefined at runtime.
```
huwshimi: I think this comment might not be correct. An object that doesn't have any optional values defined will still be an object, not `undefined`. Or are you saying that when none of them of the values are defined the API returns `undefined` instead of an empty object (in which case this comment could be tweaked to make that clear)?

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,87 @@
...
+        entry.valueType = ValueType.BOOLEAN;
+        entry.input = { type: InputType.SELECT, options: BOOLEAN_OPTIONS };
+      } else if (isNumeric) {
+        entry.valueType = ValueType.NUMBER;
+      } else if (fieldValues.length > 0) {
+        entry.input = {
+          type: InputType.SELECT,
+          options: fieldValues.map((fieldValue) => ({
+            // Values are plain primitives and not wrapped objects which is why
+            //  we don't need extra logic to extract the value from a single-key object.
```
huwshimi: Hmm.. maybe I'm just not understanding this comment, but `fieldValue` is of type `object` and converting that to a string like you do below would result in `"[object Object]"`.

src/store/middleware/source/util.ts
Ninfa-Jeon: oh yes, I got confused with what you were asking. the reason this might be undefined is because we are doing a Record lookup for this object which means it may or may not be present in the facade response. I will remove the fallback and introduce a branch for handling undefined gracefully

src/store/middleware/source/util.ts
Ninfa-Jeon: I have added a more verbose comment for this. Basically, there is a discrepancy in the juju codebase between what the schema says vs what the underlying Go code does. No object wrapper is converting the values to object, which is why the values are coming back as primitives which don't need object to string conversion.

huwshimi: OK, do you know what the primitive types can be? We also don't want strings like `"undefined"` or `"null"`.

Ninfa-Jeon: I have already brought this up with Simone and he thinks there could be an improvement there so that we don't have to assume anything but my understanding is that there can be `string`, `int`, `float` and `bool` as the only accepted config values, i.e., the `values` is dependent on `type`

huwshimi: I think this is probably a case where we want to validate the value is a type that we expect, as APIs always do something unexpected, especially in a case like this that has user provided data.

Ninfa-Jeon: okay, i have pushed some logic which validates the values against the field type that we are expecting...then we only populate options when the types match

SimoneDutto: yes, the `value` is dependent on `type` so to do the conversion i would guess you only need to leave `value` to `[]any` and later on convert it to `[]type`. The reason is that way is that i don't think you can express a json schema with generic types.

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+  AdditionalProperties is a single-key object where the key name is
+  arbitrary — the actual value is always the first (and only) entry.
+  Applies the priority hierarchy: controller > region (if known) > juju default.
+*/
+const resolveFieldDefault = (
+  // `ModelDefaultsResult["config"]` is a `Record<string, ModelDefaults>`, so
+  // a lookup by key is typed as `ModelDefaults` but may be `undefined` at runtime.
+  defaults: ModelDefaults | undefined,
```
huwshimi:
```suggestion
  defaults?: ModelDefaults,
```

src/store/middleware/source/util.ts
```diff
@@ -0,0 +1,98 @@
...
+  const fields: CategoryDefinitionField[] = Object.entries(schema ?? {}).map(
+    ([label, field]) => {
+      const isNumeric = field.type === "int" || field.type === "float";
+
+      // The response schema types `values` as `object[]`, but the underlying
+      //  Go type is `[]any` which serializes enum values as primitives on
+      //  the wire. We validate each entry matches the declared field type
+      //  to guard against unexpected API responses.
+      const fieldValues = (field.values ?? []).filter(
+        (fieldValue) => typeof fieldValue === field.type,
```
huwshimi, line 66: This could do with a test.

huwshimi (approved):
Thanks for the improvements. Two small comments.

PR #2390, opened 2026-07-09
feat: add preset-defaults middleware

Ninfa-Jeon:
## Done

- Added `model-config-defaults` source which calls both `modelDefaultsForClouds` and `modelConfigSchema` methods
- It leverages the normalisation util `generateCategoryDefinitions` to consolidate the responses from both into `CategoryDefinition[]` for the UI
- The UI is currently NOT wired to this middleware
- Moved types around so that the middleware is not importing from the UI 

## Details
https://warthogs.atlassian.net/browse/JUJU-10099

### DIFF
--- a/src/juju/types.ts
+++ b/src/juju/types.ts
@@ -8,6 +8,7 @@ import type CharmsV6 from "@canonical/jujulib/dist/api/facades/charms/CharmsV6";
 import type ClientV8 from "@canonical/jujulib/dist/api/facades/client/ClientV8";
 import type { FullStatus } from "@canonical/jujulib/dist/api/facades/client/ClientV8";
 import type CloudV7 from "@canonical/jujulib/dist/api/facades/cloud/CloudV7";
+import type CloudV8 from "@canonical/jujulib/dist/api/facades/cloud/CloudV8";
 import type ControllerV12 from "@canonical/jujulib/dist/api/facades/controller/ControllerV12";
 import type ModelManagerV10 from "@canonical/jujulib/dist/api/facades/model-manager/ModelManagerV10";
 import type {
@@ -48,7 +49,7 @@ export type Facades = {
   block?: BlockV2;
   charms?: CharmsV6;
   client?: ClientV8;
-  cloud?: CloudV7;
+  cloud?: CloudV7 | CloudV8;
   controller?: ControllerV12;
   modelManager?: ModelManagerV10 | ModelManagerV11;
   modelUpgrader?: ModelUpgraderV1;
--- a/src/pages/AddModel/ConfigsConstraints/utils.ts
+++ b/src/pages/AddModel/ConfigsConstraints/utils.ts
@@ -290,7 +290,10 @@ export function validateAndParseYAML(
           const allowedValues = (field.input.options ?? []).map(
             ({ value: optionValue }) => optionValue,
           );
-          if (!allowedValues.includes(stringValue)) {
+          if (
+            stringValue === undefined ||
+            !allowedValues.includes(stringValue)
+          ) {
             invalidValues.push({
               line,
               message: getYAMLErrorMessage(YAMLErrorType.INVALID_VALUES, {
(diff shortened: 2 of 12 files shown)


src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -290,7 +290,10 @@ export function validateAndParseYAML(
           const allowedValues = (field.input.options ?? []).map(
             ({ value: optionValue }) => optionValue,
           );
-          if (!allowedValues.includes(stringValue)) {
+          if (
+            stringValue === undefined ||
```
Ninfa-Jeon, line 294: Needed this guard as I moved types around and replaced the `SelectProps` from react-components with  locally defined `SelectOption` to avoid importing from the library in middleware

src/juju/types.ts
```diff
@@ -48,7 +48,7 @@ export type Facades = {
   block?: BlockV2;
   charms?: CharmsV6;
   client?: ClientV8;
-  cloud?: CloudV7;
+  cloud?: CloudV8;
```
huwshimi: Is CloudV8 only available on Juju 4? If so this will need to be `CloudV7 | CloudV8` so it has the types for Juju 3.6.

src/store/middleware/source/model-config-defaults.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  // modelConfigSchema is only available on CloudV8 (Juju >= 4.0.12).
+  // Skip both calls if the schema method is unavailable — the UI will fall
+  // back to the static catalog. A follow-up PR will handle the defaults-only
+  // path for older Juju versions.
+  if (typeof connection.facades.cloud?.modelConfigSchema !== "function") {
+    return [];
+  }
+
+  const [defaultsResponse, schemaResponse] = await Promise.all([
```
huwshimi: This looks like it needs some error handling if the calls fail.

src/store/middleware/source/model-config-defaults.ts
```diff
@@ -0,0 +1,102 @@
...
+      throw new Error("connection not provided");
+    }
+
+    const connection = meta.connections.wsControllerURL;
+
+    if (!connection?.info.user?.identity) {
+      throw new Error(NOT_AUTHENTICATED_ERROR);
+    }
+
+    return createPollingSource(
```
huwshimi, line 95: Are these values likely to change? Aren't they fixed for each release of Juju so we'd only get new defaults if the Controller/Model version updated?

If not, this could be converted to a process instead (or a very long timeout).

Ninfa-Jeon: While it is unlikely, the values can change if the user explicitly sets controller config defaults but I do think it could use a long timeout. As far as I remember correctly, this cannot be a process because it is a fetch-like call. Processes can be used when the operation is similar to a put or post.

src/store/middleware/source/model-config-defaults.ts
Ninfa-Jeon: Isn't the error handling already in place for source middleware? we don't have to explicitly do it, @andogq can you confirm?

andogq: The middleware will catch thrown errors, yes. However you should still do your best to find error cases, and make sure they can be handled in predictable/sensible ways (eg. an error value that explains what actually went wrong, and could be presented to the user).

More specifically, with `Promise.all`, only one rejection will be returned, so if both fail then information will be lost. It's probably better to use `Promise.allSettled`, see what failed, and construct a sensible error for each possible outcome.

Ninfa-Jeon: Noted. One thing to note for the future PRs is that we probably will not be surfacing these errors to the UI...not currently anyway. I could make it so that we fallback to static catalog if that happens though

src/juju/types.ts
Ninfa-Jeon: Done

src/store/middleware/source/model-config-defaults.ts
Ninfa-Jeon: Done

src/store/middleware/source/model-config-defaults.ts
Ninfa-Jeon: As discussed, I have increased the timeout to 5 minutes

huwshimi (approved):
Looks good, thanks.

PR #2391, opened 2026-07-13
feat: updated model list design (JUJU-9353)

andogq:
- #2392
- #2393

## Done

- Create `DataTable` component for complex table layouts
  - Row selection
  - Sorting by column
  - Grouping by column with spanned cell
  - Validated props via TS types
- Re-implement model list table with new `DataTable`

Note: This is only the table layout, additional functionality has not been included in this PR:
- Bulk model actions
- Table pagination

## QA

- View models index
- Ensure navigation to models via links works
- Ensure actions menu works
- Ensure table updates when state changes (destroying model, adding new model, etc)
- Ensure changing grouping changes layout
- Ensure table sorting functions


src/components/DataTable/types.ts
```diff
@@ -0,0 +1,95 @@
...
+  /**
+   * Alignment of the cells in this column.
+   */
+  align?: "center" | "left" | "right";
+  /**
+   * Class name to apply to all cells within this column.
+   */
+  className?: string;
+  /**
+   * map operation to extract data for this column from the row.
```
Ninfa-Jeon: Tbh a bit conflicted about this one because I know the operation is small case but putting this here just in case you'd wanna make this consistent with the other comments
```suggestion
   * Map operation to extract data for this column from the row.
```

src/components/DataTable/useToggleSelect.ts
```diff
@@ -0,0 +1,92 @@
...
+  }, [selected, keys]);
+
+  const toggle = useCallback(
+    (key: T) => {
+      if (!keys.includes(key)) {
+        return;
+      }
+
+      setSelected((previous) => {
+        // Try remove the value.
```
Ninfa-Jeon:
```suggestion
        // Try to remove the value.
```

src/components/DataTable/useToggleSelect.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(result.current.state).toEqual("none");
+    });
+
+    it("correctly detects all", ({ expect }) => {
+      const { result } = renderHook(() => useToggleSelect(KEYS));
+      act(() => {
+        for (const key of KEYS) {
+          result.current.toggle(key);
+        }
+      });
```
Ninfa-Jeon, line 104: any reason why you did not use `toggleAll` here?

src/components/DataTable/useSortState.ts
```diff
@@ -0,0 +1,41 @@
+import { useCallback, useState } from "react";
+
+/**
+ * Managed state for sorting.
+ */
+type SortState<K> = {
+  /**
+   * Currently selected sort column and direction, or `null` if no sort is active.
+   */
+  sort: { key: K; direction: "ascending" | "descending" } | null;
```
Ninfa-Jeon: I think this could be extracted to its own type as I see its usage below and in other files too. wdyt?

src/components/DataTable/useSortState.ts
```diff
@@ -0,0 +1,41 @@
...
+    direction: "ascending" | "descending";
+  } | null>(null);
+
+  const toggleSort = useCallback((key: K) => {
+    setSort((existing) => {
+      if (existing?.key === key) {
+        if (existing.direction === "ascending") {
+          return { key: existing.key, direction: "descending" };
+        } else {
+          return null;
```
Ninfa-Jeon, line 31: Maybe I'm not understanding the requirement here but if I'm reading this right, then if the sort is already applied and is in ascending order, then on toggle we switch it to descending. But the reverse doesn't seem to be true. If the sort direction is currently descending, we don't reverse it to ascending and instead return `null`.  Why is that? Is the name `toggleSort` misleading?

src/components/DataTable/useSortState.test.ts
```diff
@@ -0,0 +1,98 @@
...
+      });
+      expect(result.current.sort).toEqual({
+        key: "other",
+        direction: "ascending",
+      });
+    });
+  });
+
+  describe("from descending", () => {
+    it("clears sort on same key toggle", ({ expect }) => {
```
Ninfa-Jeon, line 51: ah, my bad. this explains it. ignore the previous comment

src/components/DataTable/useRows.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+    expect(result.current).toEqual([
+      expect.objectContaining({ key: "abc123" }),
+      expect.objectContaining({ key: "def456" }),
+    ]);
+  });
+
+  it.for([
+    ["adds", true],
+    ["doesn't add", false],
+  ] as const)("%s checkbox if selectable", ([_, selectable], { expect }) => {
```
Ninfa-Jeon: the test names would be:
- `adds checkbox if selectable`
- `doesn't add checkbox if selectable`
maybe this would read better?:

```suggestion
  it.for([
    ["adds", true],
    ["doesn't add", false],
  ] as const)("%s checkbox if %s", ([_, selectable], { expect }) => {
```

src/components/DataTable/useHeaders.tsx
```diff
@@ -0,0 +1,86 @@
...
+  columns,
+  groupBy,
+  sort,
+  toggleSort,
+}: {
+  columns: FlatTableColumn<keyof TValues, TRow, TValues[keyof TValues]>[];
+  groupBy?: keyof TValues;
+  sort: { key: keyof TValues; direction: "ascending" | "descending" } | null;
+  toggleSort: (column: keyof TValues) => void;
+}): {
```
Ninfa-Jeon: Can you maybe extract this to a prop type above the hook definition? This is hard to read currently

src/components/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,137 @@
...
+    groupBy,
+    sort,
+    select,
+    selectable,
+  });
+
+  const columnCount = useMemo(
+    () =>
+      orderedHeaders.length +
+      (prefixHeader === null ? 1 : 0) +
```
Ninfa-Jeon: I think this check is inverted

```suggestion
      (prefixHeader !== null ? 1 : 0) +
```

src/urls.ts
```diff
@@ -10,9 +10,10 @@ export enum ModelTab {
   LOGS = "logs",
   SECRETS = "secrets",
 }
+export const VALID_GROUPINGS = ["status", "cloud", "owner"] as const;
```
Ninfa-Jeon: maybe this should live in `consts.ts`?

src/components/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,233 @@
...
+    [models, controllers],
+  );
+
+  const column = useMemo(
+    () => DataTable.columnBuilder<(typeof tableData)[number]>(),
+    [],
+  );
+
+  return (
+    <>
```
Ninfa-Jeon: is this fragment required?

Ninfa-Jeon (approved):
Great job on this one, Tom! No big issues at all. Some minor comments/test changes, a few questions and one possible bug (or gap in my understanding ;)).

Ninfa-Jeon:
Found two points of improvement during QA but overall it looks fine to me.

1. Maybe the orderBy prefix could be wrapped in a `TruncatedTooltip` when the text in long
Screenshot 2026-07-16 at 4 24 54 PM

2. For header cells with long texts, it is hard to tell if a sorting is being applied because the chevron icon is getting hidden behind the ellipsis along with overflowing text

https://github.com/user-attachments/assets/2265b335-cbf3-4fd9-a317-a4ecbbcd06c0

src/components/DataTable/useToggleSelect.test.ts
andogq: `toggleAll` has additional logic beyond selecting every cell. This test is meant to detect if the `state` is correctly derived, so doing it in the most direct way (selecting each key individually) helps isolate what we're actually trying to test.

src/components/DataTable/useSortState.ts
andogq: It's a very standard usage pattern of table sorting. A given column is in one of there sort states: ascending, descending, or none. If our column click just toggled between ascending/descending, then there would be no way for the user to remove the sort. So instead, our sorting logic is just a little state machine to transition through the three states.

src/components/DataTable/DataTable.tsx
andogq: Very good catch, thank you

andogq:
Unfortunately, the sorting chevrons are completely handled with Vanilla's default styles, which I've purposefully tried to use as much as possible without alterations (for example, the entire header column will display as a link, but only the text can actually be clicked). These are upstream issues that should be fixed in Vanilla.

src/components/DataTable/useToggleSelect.test.ts
Ninfa-Jeon: Makes sense.

src/pages/ModelsIndex/types.ts
```diff
@@ -5,3 +5,6 @@ export enum Label {
 export enum TestId {
   COMPONENT = "ModelsIndex",
 }
+
+export const VALID_GROUPINGS = ["status", "cloud", "owner"] as const;
```
Ninfa-Jeon, line 9: I believe this can now be removed?

Ninfa-Jeon:
I don't see the sort function particularly working for the `configuration` column. Do we know on what basis the sort should work on it? If not, we should disable sorting on that column.

src/components/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,263 @@
...
+                    className="models__version"
+                  />
+                ) : null}
+              </TruncatedTooltip>
+            );
+          },
+        }),
+        actions: column({
+          header: "Actions",
+          collapse: true,
```
Ninfa-Jeon: This prop is messing with the alignment of this column
```suggestion
```
PFA before and after removing it:

Image

Image

Ninfa-Jeon:
Okay there are a few things that I've found during QA that seem odd. For some reason, I can't find relevant figma designs for it but from `main` I understand that:
1. When a user orders the list by cloud, a new column for region is created and the ordering is done by plain `cloud` not `cloud/region` combination. This also means that in `cloud` ordering, an extra column is rendered, specifically for `region`
Screenshot 2026-07-18 at 9 04 00 PM

2. In all orderings, the model name column is the very first column

3. Between different orderings, the `owner` and `status` columns switch places unless the selected ordering is `cloud` in which case they appear in the order `owner` -> `status` -> `region`.
Screenshot 2026-07-18 at 9 05 18 PM
Screenshot 2026-07-18 at 9 05 26 PM

src/components/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,263 @@
...
+          className: "status",
+          map: ({ model }) => getModelStatusGroupData(model).highestStatus,
+          renderCell: (status) => (
+            <Status status={status}>
+              <TruncatedTooltip message={status} className="u-capitalise">
+                {status}
+              </TruncatedTooltip>
+            </Status>
+          ),
+        }),
```
Ninfa-Jeon: The ordering can be fixed if we move this block to after the `Owner` column below.

src/components/ModelTableList/ModelTable/ModelTable.tsx
```diff
@@ -1,317 +0,0 @@
...
-          className: classNames({ "dying-model": isDying }),
-        },
-        {
-          ...testId(TestId.COLUMN_REGION),
-          content: (
-            <TruncatedTooltip message={region}>{region}</TruncatedTooltip>
-          ),
-          className: classNames({ "dying-model": isDying }),
-        },
-      ),
```
Ninfa-Jeon, line 192: this was how we conditionally rendered the `region` column when we ordered the table by clouds

src/components/ModelTable/ModelTable.tsx
```diff
@@ -0,0 +1,263 @@
...
+            return <TruncatedTooltip message={owner}>{owner}</TruncatedTooltip>;
+          },
+        }),
+        cloud: column({
+          header: "Cloud/Region",
+          sortable: true,
+          className: "region",
+          map: ({ model }) => model.info?.["cloud-tag"],
+          renderCell: (_cloudTag, { model }) => <CloudCell model={model} />,
+        }),
```
Ninfa-Jeon, line 214: Something like the following should help you render the `region` column conditionally. I have not found a way to update the `cloud/region` column that appears first in the table during `cloud` ordering though.
```suggestion
        }),
        ...(groupBy === "cloud"
          ? {
              region: column({
                header: "Region",
                sortable: true,
                className: "region",
                map: ({ model }) => model.info?.["cloud-tag"],
                renderCell: (_cloudTag, { model }) => {
                  const region = getRegion(model);
                  if (region === null) {
                    return null;
                  }

                  return (
                    <TruncatedTooltip message={region}>
                      {region}
                    </TruncatedTooltip>
                  );
                },
              }),
            }
          : {}),
```

src/components/ModelTable/_model-table.scss
```diff
@@ -1,4 +1,18 @@
-.models .p-main-table {
```
Ninfa-Jeon, line 1: I think a lot of these rules for columns can be defined just once. Looking at the designs, I think the `model` and `configuration` columns are the widest, the `actions` column is as narrow as possible and the rest share the same width. By only adjusting widths for these three columns and allowing the rest to span equally would help avoid setting widths explicitly for each view.

Ninfa-Jeon (commented):
I tried to fix a few things myself so that you don't have to do all of it, however I couldn't get to the cloud ordering without having the `/region` in the first column and actually consolidating the css rules to make things look a bit more uniform between views.

Ninfa-Jeon (changes requested):

src/components/ModelTable/ModelTable.tsx
Ninfa-Jeon: This still seems to be an issue. I see it on Safari as well as on Chrome. But not on Firefox.

Ninfa-Jeon (approved):
Thanks, Tom. All looks good now.

PR #2394, opened 2026-07-15
refactor: make configs-constraints array fields in Formik

Ninfa-Jeon:
## Done

- Restructured `configCatalog` and `constraintsCatalog` to support usage of an array type field
- Updated all utils to work on this restructured data
- This is the primary set of changes required for the eventual wiring of configs and their defaults with the UI. If left unchanged, Formik throws an error in console stating an uncontrolled value was changed to controlled due to the dynamic nature of incoming fields.

Note: I haven't used `FieldArray` here as it is only needed when you're dynamically adding/removing items from a Formik array. This might be a future requirement but it would be unused presently.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Make sure the entire add-model flow works correctly with no errors or bugs

### DIFF
--- a/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
@@ -3,22 +3,61 @@ import userEvent from "@testing-library/user-event";
 import { Formik } from "formik";
 import { vi } from "vitest";
 
+import { BOOLEAN_OPTIONS } from "consts";
 import { externalURLs } from "urls";
 
 import { InputMode } from "../../types";
-import { CONFIG_CATEGORIES } from "../configCatalog";
-import { FieldName, Label } from "../types";
+import { ConfigCategory } from "../configCatalog";
+import {
+  type ConfigFieldEntry,
+  FieldName,
+  InputType,
+  Label,
+  ValueType,
+} from "../types";
 
 import CategoriesListing from "./CategoriesListing";
 import type { Props as CategoriesListingProps } from "./types";
 
 describe("CategoriesListing", () => {
   let defaultProps: CategoriesListingProps;
+  let initialValues: {
+    [FieldName.CONFIG_FIELDS]: ConfigFieldEntry[];
+    [FieldName.CONFIG_INPUT_MODE]: InputMode;
+  };
 
   beforeEach(() => {
+    initialValues = {
+      [FieldName.CONFIG_FIELDS]: [
+        {
+          label: "default-space",
+          value: "my-space",
+          defaultValue: "",
+          category: ConfigCategory.NETWORKING,
+          arrayIndex: 0,
+        },
+        {
+          label: "net-bond-reconfigure-delay",
+          value: 15,
+          defaultValue: 17,
+          category: ConfigCategory.NETWORKING,
+          arrayIndex: 1,
+          valueType: ValueType.NUMBER,
+        },
+        {
+          label: "disable-network-management",
+          value: true,
+          defaultValue: false,
+          category: ConfigCategory.NETWORKING,
+          arrayIndex: 2,
+          valueType: ValueType.BOOLEAN,
+        },
+      ],
+      [FieldName.CONFIG_INPUT_MODE]: InputMode.LIST,
+    };
     defaultProps = {
       title: Label.CONFIGS_TITLE,
-      categoriesList: CONFIG_CATEGORIES,
+      arrayName: FieldName.CONFIG_FIELDS,
       inputMode: FieldName.CONFIG_INPUT_MODE,
       yamlKey: FieldName.CONFIG_YAML,
       changedOnlyLabel: Label.CHANGED_CONFIGS_ONLY,
@@ -39,10 +78,7 @@ describe("CategoriesListing", () => {
 
   it("renders properly", () => {
     render(
-      <Formik
-        initialValues={{ [FieldName.CONFIG_INPUT_MODE]: InputMode.LIST }}
-        onSubmit={vi.fn()}
-      >
+      <Formik initialValues={initialValues} onSubmit={vi.fn()}>
         <CategoriesListing {...defaultProps} />
       </Formik>,
     );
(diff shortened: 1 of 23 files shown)


src/pages/AddModel/ConfigsConstraints/configCatalog.ts
```diff
@@ -1,527 +1,560 @@
...
-          type: InputType.SELECT,
-          options: BOOLEAN_OPTIONS,
-        },
-      },
-    ],
   },
   {
+    label: "vpc-id",
+    description: "Use a specific VPC network",
+    category: "Cloud-specific configurations",
```
huwshimi: These should definitely be enums so you have a central place to maintain the categories.

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
huwshimi, line 1: If you think it would be easier to maintain I think you could keep the fields as an array per category and then map them to a flat field array (you could do that inside this file at run time or possibly inside the AddModel component). That way you wouldn't also need the util to construct the categories.

Up to you, if you think this is easy enough to maintain then that's fine.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -172,23 +160,13 @@ export const getYAMLErrorMessage = (
...
-      for (const field of fields) {
-        acc[field.label] = field;
-      }
-      return acc;
-    },
-    {} as Record<string, CategoryDefinitionField>,
-  );
-
-  const validValues: Record<string, ConfigFieldValue> = {};
+  const validValues = [...entries];
```
huwshimi: We have the `clone-deep` library in the dashboard that you could use here, which would also save you from spreading each of the objects below.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -236,28 +214,25 @@ export function validateAndParseYAML(
         ? yamlString.slice(0, pair.key.range[0]).split("\n").length
         : 0;
 
-      const field = fieldsByLabel[key];
-      if (!field) {
+      const index = entries.findIndex((entry) => entry.label === key);
+      if (index === -1) {
```
huwshimi: If you cloned the entries (see comment above) then you could just do const field = validValues.find(...

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -236,28 +214,25 @@ export function validateAndParseYAML(
...
-          message: getYAMLErrorMessage(YAMLErrorType.INVALID_VALUES, {
-            key,
-          }),
+          message: getYAMLErrorMessage(YAMLErrorType.INVALID_VALUES, { key }),
         });
         continue;
       }
 
       const { value } = pair.value;
+      const field = entries[index];
```
huwshimi: With a cloned list the field would have been assigned above.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -304,20 +277,22 @@ export function validateAndParseYAML(
...
           }
         }
-        validValues[key] = stringValue;
       }
+
+      // Treat null as an empty string so inputs stay controlled.
+      validValues[index] = {
+        ...field,
+        value: (value ?? "") as ConfigFieldValue,
+      };
```
huwshimi: If the field had been found above then you could just do `field.value = ...`

src/store/middleware/source/model-config-defaults.test.ts
```diff
@@ -58,7 +58,9 @@ describe("getModelConfigDefaults", () => {
       results: [
         {
           config: {
-            "default-space": { default: { value: "my-space" } },
+            "default-space": {
+              default: "my-space" as unknown as Record<string, unknown>,
```
huwshimi: Is the API returning something different to what we expect? If so these `as unknown as` annotations could do with a comment.

huwshimi (commented):
This is looking good. A few comments and questions.

src/store/middleware/source/model-config-defaults.test.ts
Ninfa-Jeon: oh wait, this can be removed...i think it was needed at some point of the refactor but is redundant now...this annotation is only needed at util level test because juju is returning properties of type `AdditionalProperties`

src/pages/AddModel/ConfigsConstraints/configCatalog.ts
Ninfa-Jeon: That was my first plan of action actually. But I figured that whenever we start receiving categories from Juju, they will probably come as a property per field, just the was we have here. But, if you reckon the other way round would be easier for this static structure, i can do that as well...it would save me a lot of manual changes here too

huwshimi: I don't have strong feelings either way!

Ninfa-Jeon: I think I'd like to keep the catalogs this way for now. Because either way, if we were to receive similar format from the facade, we would have to construct the categories there anyway (unless the facade decides to give us an easier to work with structure).

src/pages/AddModel/ConfigsConstraints/CategoriesListing/CategoriesListing.test.tsx
```diff
@@ -3,22 +3,60 @@ import userEvent from "@testing-library/user-event";
...
+  };
 
   beforeEach(() => {
+    initialValues = {
+      [FieldName.CONFIG_FIELDS]: [
+        {
+          label: "default-space",
+          value: "my-space",
+          defaultValue: "",
+          category: "Networking & Firewall",
```
huwshimi: I'd probably update these tests to use the enum as well.

src/pages/AddModel/ConfigsConstraints/utils.ts
```diff
@@ -172,23 +161,13 @@ export const getYAMLErrorMessage = (
...
-      for (const field of fields) {
-        acc[field.label] = field;
-      }
-      return acc;
-    },
-    {} as Record<string, CategoryDefinitionField>,
-  );
-
-  const validValues: Record<string, ConfigFieldValue> = {};
+  const validValues = cloneDeep(entries);
```
huwshimi, line 170: I misunderstood the loop on line 290 (I thought it was comparing the value in entries to the updated value in validValues), so why does this need to be cloned? It might also be handy to leave a comment for the future.

huwshimi (approved):

PR #2395, opened 2026-07-15
feat: use configs and defaults in UI from the facades

Ninfa-Jeon:
Depends on https://github.com/canonical/juju-dashboard/pull/2394

## Done

- Wired the middleware to fetch and store model config schema and defaults to the UI.

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Try viewing the add model flow at Juju 4.0.12 (live) and any other version before that (static)
- The configs and constraints tab should work fine either way and the model should be added succesfully
- A simple way to distinguish between values from the facade and the static ones is the configuration `fan-config`. This was deprecated in Juju v4.0 so the facade will not return it.

## Details

https://warthogs.atlassian.net/browse/JUJU-10098

### DIFF
--- a/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
@@ -1,8 +1,18 @@
-import { act, fireEvent, screen, within } from "@testing-library/react";
+import {
+  act,
+  fireEvent,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { Formik } from "formik";
-import { vi } from "vitest";
 
+import {
+  cloudInfoStateFactory,
+  jujuStateFactory,
+} from "testing/factories/juju/juju";
+import { rootStateFactory } from "testing/factories/root";
 import { renderComponent } from "testing/utils";
 import { externalURLs } from "urls";
 
@@ -35,6 +45,41 @@ describe("ConfigsConstraints", () => {
       [FieldName.CONSTRAINT_YAML]: "",
       [FieldName.DISABLED_COMMANDS]: DisableType.NONE,
     };
+  });
+
+  it("replaces configFields with live entries when store data arrives", async () => {
+    const liveEntry = {
+      label: "live-field",
+      value: "live-value",
+      defaultValue: "live-value",
+      category: null,
+      arrayIndex: 0,
+    };
+    const state = rootStateFactory.withGeneralConfig().build({
+      juju: jujuStateFactory.build({
+        cloudInfo: cloudInfoStateFactory.build({
+          clouds: { "cloud-aws": { type: "ec2" } },
+        }),
+        modelConfigDefaults: {
+          errors: null,
+          loading: false,
+          defaults: { ec2: [liveEntry] },
+        },
+      }),
+    });
+    renderComponent(
+      <Formik
+        initialValues={{ ...initialValues, cloud: "cloud-aws" }}
+        onSubmit={vi.fn()}
+      >
+        <ConfigsConstraints />
+      </Formik>,
+      { state },
+    );
+
+    await waitFor(() => {
+      expect(screen.getByLabelText("live-field")).toBeInTheDocument();
+    });
   });
 
   it("renders properly", () => {
--- a/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
@@ -8,8 +8,10 @@ import {
 import userEvent from "@testing-library/user-event";
 import { Formik } from "formik";
 
+import { actions as jujuActions } from "store/juju/slice";
 import {
   cloudInfoStateFactory,
+  configFieldEntryFactory,
   jujuStateFactory,
 } from "testing/factories/juju/juju";
 import { rootStateFactory } from "testing/factories/root";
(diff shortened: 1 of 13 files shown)


src/store/juju/selectors.test.ts
```diff
@@ -3315,6 +3316,45 @@ describe("getCloudInfoState", () => {
...
+            ec2: [
+              {
+                label: "firewall-mode",
+                category: null,
+                value: "instance",
+                defaultValue: "instance",
+                arrayIndex: 0,
+              },
+            ],
+          },
```
andogq: Is it worth adding this to the factory?

src/pages/AddModel/MandatoryDetails/MandatoryDetails.tsx
```diff
@@ -121,6 +122,35 @@ const MandatoryDetails = (): JSX.Element => {
...
+      dispatch(
+        modelConfigDefaultsSource.actions.stop({
+          wsControllerURL,
+          cloudTag: selectedCloud,
+          providerType,
+          selectedRegion: values.region || undefined,
+        }),
+      );
+    };
+  }, [dispatch, wsControllerURL, selectedCloud, providerType, values.region]);
```
andogq, line 144: It might be good to save this param as a variable, so it can be re-used in the `stop`.

```suggestion
  useEffect(() => {
    if (!wsControllerURL || !selectedCloud || !providerType) {
      return;
    }
    const param = {
      wsControllerURL,
      cloudTag: selectedCloud,
      providerType,
      selectedRegion: values.region || undefined,
    };
    dispatch(
      modelConfigDefaultsSource.actions.start(param),
    );
    return (): void => {
      dispatch(
        modelConfigDefaultsSource.actions.stop(param),
      );
    };
  }, [dispatch, wsControllerURL, selectedCloud, providerType, values.region]);
```

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
```diff
@@ -15,10 +21,30 @@ import { DISABLED_COMMAND_OPTIONS } from "./disabledCommandOptions";
...
+  // When live config fields arrive, atomically replace the configFields array.
+  const liveEntries = providerType
+    ? (modelConfigDefaults.defaults[providerType] ?? null)
+    : null;
+
+  useEffect(() => {
+    if (!liveEntries || liveEntries.length === 0) {
+      return;
+    }
+    void setFieldValue(FieldName.CONFIG_FIELDS, liveEntries, false);
```
andogq, line 54: This will clear the value a user already had if they start typing before it loads, it should probably be merged with existing values so they don't loose their work.

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
```diff
@@ -37,6 +47,41 @@ describe("ConfigsConstraints", () => {
     };
   });
 
+  it("replaces configFields with live entries when store data arrives", async () => {
```
andogq, line 53: This test should render the component normally first, assert `live-field` isn't present, update the config defaults, and then re-render, to better simulate the async loading.

It (or another test) should also test the behaviour if the user has already filled in a form before the data loads.

src/pages/AddModel/MandatoryDetails/MandatoryDetails.test.tsx
```diff
@@ -211,7 +222,40 @@ describe("MandatoryDetails", () => {
...
+    await userEvent.selectOptions(
+      screen.getByLabelText(Label.REGION),
+      "us-east-1",
+    );
+
+    await waitFor(() => {
+      const starts = actions.filter(
+        (dispatch) => dispatch.type === "source/model-config-defaults/start",
+      );
+      expect(starts.length).toBeGreaterThanOrEqual(2);
```
andogq: This should be an exact check `toEqual(2)`. If it's making extra fetches due to re-renders, it's likely a bug.

Also it'd be good to check the payload of both the `start` calls (to ensure it's actually changing when the dropdown is altered).

andogq (commented):
Looks pretty good overall.

My biggest problem is that the incoming async defaults can usurp the user's existing work, meaning that they will loose any work they've already done. I can see two options that'd fix this:
- A little loading indicator on the side of the form (such as next to the radio buttons/search box) showing that the default values are loading.
- A dialog similar to the one switching from yaml->form, which shows if the incoming values would overwrite the user's existing work, allowing them to accept the overwriting, or be switched to the yaml view with their data, so they can edit it.

Either way, we don't want the user being surprised at the form suddenly changing, or them loosing their data (imagine if a field was renamed between the hardcoded values, and the config defaults that arrive).

---

Aside from that, a few tiny tweaks to the tests!

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.tsx
Ninfa-Jeon: Finalising a screen for this with Lukas

Ninfa-Jeon: Added a temporary stub until I hear from him

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
```diff
@@ -37,6 +50,110 @@ describe("ConfigsConstraints", () => {
...
+      { state },
+    );
+
+    const configsSection = screen.getByRole("region", {
+      name: Label.CONFIGS_TITLE,
+    });
+    expect(
+      within(configsSection).getByText(/Fetching configurations./),
+    ).toBeInTheDocument();
+
```
andogq, line 118:
```suggestion
    expect(
      within(configsSection).getByText(/Fetching configurations./),
    ).toBeInTheDocument();
    // All config inputs should be disabled while defaults are loading so the
    // user cannot enter values that will be overwritten on arrival.
    within(configsSection)
      .getAllByRole("textbox")
      .forEach((input) => {
        expect(input).toBeDisabled();
      });
  });

```

I thin it's ok to merge these tests, they're close enough and have a lot of duplicated assertions/setup.

andogq (approved):
One tiny test tweak! :shipit:

andogq:
QA ok, 'Fetching configurations' message is alright until we can get some better designs for it. Nice work!

PR #2398, opened 2026-07-20
feat: introduce SkeletonPlaceholder component

Ninfa-Jeon:
## Done

- Added the `SkeletonPlaceholder` component
- This is rendered when the configuration defaults are being fetched
- Removed the warning text in favor of skeletons (confirmed with UX)

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Try navigating to the ConfigsConstraints tab with slow internet.
- View the loading state

## Screenshots
Screenshot 2026-07-20 at 3 44 37 PM

## Details
Part of https://warthogs.atlassian.net/browse/JUJU-10185


src/components/ModelTable/ModelTable.tsx
```diff
@@ -77,6 +77,7 @@ export default function ModelTable({ models, groupBy }: Props): JSX.Element {
 
   return (
     <DataTable
+      selectable={false}
```
Ninfa-Jeon, line 80: i think this change got reverted at some point

src/scss/custom/_skeleton.scss
```diff
@@ -0,0 +1,27 @@
...
+    var(--vf-color-background-hover) calc(50% + 2rem)
+  );
+  background-size: 300% 100%;
+  border-radius: 4px;
+  display: inline-block;
+  min-height: 1.5rem;
+  overflow: hidden;
+  pointer-events: none;
+  width: 100%;
+}
```
andogq, line 17: Since this skeleton/throber class is useful for more than just this page, I think it'd be good to hoist it or make a mix-in for it.

src/pages/AddModel/ConfigsConstraints/SkeletonPlaceholder/SkeletonPlaceholder.test.tsx
```diff
@@ -0,0 +1,27 @@
+import { render, screen } from "@testing-library/react";
+
+import SkeletonPlaceholder from "./SkeletonPlaceholder";
+
+describe("SkeletonPlaceholder", () => {
+  beforeEach(() => {
+    vi.spyOn(Math, "floor").mockReturnValue(0);
```
andogq:
```suggestion
    vi.spyOn(Math, "random").mockReturnValue(0);
```
Mocking `random` is probably a bit clearer. However, it'd be better to have `delay` as a prop, which defaults to a random value.

src/pages/AddModel/ConfigsConstraints/SkeletonPlaceholder/SkeletonPlaceholder.tsx
```diff
@@ -0,0 +1,30 @@
...
+  className?: string;
+  loading?: boolean;
+};
+
+const SkeletonPlaceholder = ({
+  children,
+  className,
+  loading = true,
+}: Props): JSX.Element => {
+  const delay = Math.floor(Math.random() * 750);
```
andogq: This will be evaluated every render, you probably want it in a `useMemo` which defaults to a `delay` prop, so it can be injected for testing.

andogq (approved):
Main suggestion is to move the skeleton CSS into a standalone class/mixin for reusability (I'm sure we'll need this pattern else where).

I'm unsure of the accessibility implications of the skeleton, I'd imagine that we'd want some kind of label or announcer, but I'm not 100% sure. Potentially this is something we can work on in further improvements to it.

andogq:
Few thoughts from QA:
- YAML textarea input is disabled, should this also be a skeleton?
- The `configCatalog` and `constraintsCatalog` are still hardcoded in the dashboard, these can be removed now, right?

Ninfa-Jeon:
> * YAML textarea input is disabled, should this also be a skeleton?

No, that will be enabled with some logic (will do in another PR)

> * The `configCatalog` and `constraintsCatalog` are still hardcoded in the dashboard, these can be removed now, right?

No, we still need `configCatalog` for Juju < v4.0.12 and fallbacks, and `constraintsCatalog` doesn't have live data fetching

andogq (approved):

PR #2405, opened 2026-07-22
feat: introduce confirmation modal for switch while loading

Ninfa-Jeon:
## Done

- Enabled YAML textarea for inputs when the configs are still loading
- Introduced a new confirmation modal for when the user tries to switch back to list view

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View the add model flow at Juju > v4.0.12
- Switch to a slower network and try viewing the ConfigsConstraints tab
- The loading mode should be visible with skeletons in list view for configurations
- On switching to YAML view, you should see the input field enabled
- Currently, you can freely switch between list and yaml views
- Try entering something in the input field and then attempt to switch to list view
- You should see a confirmation modal with appropriate message

## Details

https://warthogs.atlassian.net/browse/JUJU-10185

### DIFF
--- a/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
@@ -371,13 +371,19 @@ describe("ConfigsConstraints", () => {
     );
 
     const dialog = screen.getByRole("dialog", {
-      name: "YAML inputs will be lost",
+      name: "YAML configuration will be lost",
     });
     expect(dialog).toBeInTheDocument();
     await userEvent.click(
       screen.getByRole("button", { name: "Keep editing in YAML" }),
     );
     expect(dialog).not.toBeInTheDocument();
+    expect(
+      within(configsSection).getByRole("radio", { name: InputMode.YAML }),
+    ).toBeChecked();
+    expect(within(configsSection).getByRole("textbox")).toHaveValue(
+      "default-space: my-space",
+    );
   });
 
   it("does not show the confirmation modal on switching to list mode if no inputs were made", async () => {
@@ -413,7 +419,7 @@ describe("ConfigsConstraints", () => {
 
     expect(
       screen.queryByRole("dialog", {
-        name: "YAML inputs will be lost",
+        name: "YAML configuration will be lost",
       }),
     ).not.toBeInTheDocument();
     expect(
@@ -453,7 +459,7 @@ describe("ConfigsConstraints", () => {
     );
 
     const dialog = screen.getByRole("dialog", {
-      name: "YAML inputs will be lost",
+      name: "YAML configuration will be lost",
     });
     expect(dialog).toBeInTheDocument();
     await userEvent.click(
--- a/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
+++ b/src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
@@ -338,4 +338,135 @@ describe("ConfigsConstraints", () => {
     ) as HTMLInputElement;
     expect(defaultSpaceInput.value).toBe("my-space");
   });
+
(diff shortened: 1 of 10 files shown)


src/pages/AddModel/ConfigsConstraints/ConfirmSwitchModal/ConfirmSwitchModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+          void setFieldValue(FieldName.CONFIG_YAML, "");
+          void setFieldValue(FieldName.CONFIG_INPUT_MODE, InputMode.LIST);
+          onClose();
+        }}
+        close={onClose}
+      >
+        <p>
+          List view fields are not ready yet. If you switch now, you will lose
+          your current YAML inputs. You can still create a model in the YAML
+          view.
```
andogq: Do we have any copy for this? If not, I feel it could be simplified.

> Model configuration data has not been loaded. Returning to list view will discard the current YAML configuration.

src/pages/AddModel/ConfigsConstraints/ConfirmSwitchModal/ConfirmSwitchModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+const ConfirmSwitchModal = ({
+  onClose,
+}: ConfirmSwitchModalProps): JSX.Element => {
+  const { setFieldValue } = useFormikContext<FormFields>();
+  const { Portal } = usePortal();
+
+  return (
+    <Portal>
+      <ConfirmationModal
+        title="YAML inputs will be lost"
```
andogq:
```suggestion
        title="YAML configuration will be lost"
```

src/pages/AddModel/ConfigsConstraints/ConfirmSwitchModal/ConfirmSwitchModal.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const onClose = vi.fn();
+    renderComponent(
+      <Formik initialValues={initialValues} onSubmit={vi.fn()}>
+        <ConfirmSwitchModal onClose={onClose} />
+      </Formik>,
+    );
+
+    await userEvent.click(
+      screen.getByRole("button", { name: "Keep editing in YAML" }),
+    );
```
andogq, line 62: Would be good to ensure existing values weren't cleared

src/pages/AddModel/ConfigsConstraints/ConfigsConstraints.test.tsx
```diff
@@ -338,4 +338,135 @@ describe("ConfigsConstraints", () => {
...
+    );
+
+    const dialog = screen.getByRole("dialog", {
+      name: "YAML inputs will be lost",
+    });
+    expect(dialog).toBeInTheDocument();
+    await userEvent.click(
+      screen.getByRole("button", { name: "Keep editing in YAML" }),
+    );
+    expect(dialog).not.toBeInTheDocument();
```
andogq, line 380: Would also be good to ensure the configs are retained

andogq (approved):
A few tiny tweaks to tests and copy

src/pages/AddModel/ConfigsConstraints/ConfirmSwitchModal/ConfirmSwitchModal.tsx
Ninfa-Jeon: Do you think we can keep the last line `You can still create a model in the YAML view.` to let the user know they are not blocked?

andogq: I don't think we need to keep it, as it should be indicated in other ways (eg. the user can see the 'Add model' button is still active, so there's nothing that would indicate they cannot). If you feel otherwise though, happy for you to leave it in!

PR #2406, opened 2026-07-23
fix: ensure applications are selectable without search

### LINKED ISSUE #2403: Applications only selectable with an active search query
**Describe the bug**

When I am viewing the applications list on the model page, I cannot select the rows in the application table as the checkboxes are not visible. They only show when I have a query in the search box, meaning I need to make a dummy query just to select the applications.

**To Reproduce**
1. Select a model with multiple applications
2. Notice there's no checkbox in the table
3. In the 'Filter applications' box, type ` ` (a single space) and press enter
4. Notice that the checkboxes show up and can be used

**Expected behavior**
- Checkboxes are visible always
- 'Run action' button is always visible, but is disabled if there's nothing selected

**Screenshots**

No checkboxes when no search filter:

Image

Checkboxes with search filter (it should always be like this):

Image

Gargoth:
Closes #2403

## Done

- Removed `queryParams.filterQuery` as a requirement for setting `selectable` to `true` in `LocalAppsTable.tsx`
    - `selectable` is what determines whether the "Run Action" and the checkboxes are visible.
- Modified tests in `LocalAppsTable.test.tsx` to fit new behavior.

## QA

- Navigate to a model with multiple applications and ensure behavior matches the screenshots of the **After** row below.
- Ensure `yarn lint` succeeds
- Ensure `yarn test` succeeds
    - This covers the new behavior introduced in this diff and regression tests with user permissions.

## Screenshots

|        | Without search | With search |
| ------ | :------------: | :---------: |
| Before | image |  image |
| After  | image |   image |

## Details

For more context, the behavior this diff modifies was introduced in #1629 and might have been an explicit decision made a while ago, but this leads to an awkward flow where we need a dummy query select applications and run actions on them.

### DIFF
--- a/src/pages/EntityDetails/Model/ApplicationsTab/LocalAppsTable/LocalAppsTable.test.tsx
+++ b/src/pages/EntityDetails/Model/ApplicationsTab/LocalAppsTable/LocalAppsTable.test.tsx
@@ -101,20 +101,20 @@ describe("LocalAppsTable", () => {
     expect(screen.queryAllByRole("row")).toHaveLength(8);
   });
 
-  it.each([
-    { url: url },
-    { url: searchURL }
-  ])("shows the select column regardless of search", ({ url }) => {
-    renderComponent(
-      <LocalAppsTable
-        applications={state.juju.modelData?.test123.applications}
-      />,
-      { path, url, state },
-    );
-    expect(
-      screen.queryAllByRole("columnheader")[0].querySelector("input"),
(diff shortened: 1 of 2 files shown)


andogq (approved):
Nice work!

andogq:
QA'd nicely, merge when you're ready! :shipit:

Gargoth:
Thanks! :shipit:

PR #2408, opened 2026-07-23
fix(ci): fix docker and machine charm E2E tests

andogq:
- Add `<base href="/" />` to `index.html`: With the inclusion of `base` in `vite.config.ts` (b19a8246), vite now generates relative paths for all assets. This change ensures relative assets are still pulled from the root.
- Default `app_base_url` to `""`: ensures proper NGINX template expansion when running as a docker container (similar to #2411).

### DIFF
--- a/charms/k8s-charm/src/config.py
+++ b/charms/k8s-charm/src/config.py
@@ -23,7 +23,7 @@ class Config:
         dashboard_root: str,
         analytics_enabled: bool,
         port: int,
-        base_app_url: str | None = None,
+        base_app_url: str,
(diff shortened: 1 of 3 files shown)


Ninfa-Jeon:
I looked into the failing JIMM Docker tests and figured that they were also broken because of the same set of changes and needed fixing. I have done that along with a few more corrections to `add-model` suite in my PR here: https://github.com/canonical/juju-dashboard/pull/2411. If you think they seem alright, please close this and review them.

Ninfa-Jeon (approved):
Thanks for the changes!

PR #2411, opened 2026-07-23
fix: flakey add-model e2e tests

Ninfa-Jeon:
## Done

- Fixed flakey Add Model end-to-end tests
- Also corrected the base app url which needed updating after this change https://github.com/canonical/juju-dashboard/commit/b19a8246e847d03e6b9a8e3454ca2867f3dad026
- On some occasions, the list was not updated and a message to reload the dashboard was rendered even when the operation had succeeded (PFA the artefact from [one such flakey run](https://github.com/canonical/juju-dashboard/actions/runs/29991418005/job/89154948129?pr=2408#step:9:1008))
test-failed-1

## QA

- Pull code
- Open http://0.0.0.0:8036/
- All e2e tests should pass with no flakey behaviour
- The JIMM Docker workflow should not fail with a 404 screen (https://github.com/canonical/juju-dashboard/actions/runs/29991418005/job/89228465921?pr=2408#step:14:7098)

### DIFF
--- a/e2e/base/add-model.spec.ts
+++ b/e2e/base/add-model.spec.ts
@@ -4,6 +4,7 @@ import { Label as ModelActionsLabel } from "components/ModelActions/types";
 import { Label as AccessManagementLabel } from "pages/AddModel/AccessManagement/types";
 import { Label as ConfigsConstraintsLabel } from "pages/AddModel/ConfigsConstraints/types";
 import { Label as AddModelLabel } from "pages/AddModel/types";
+import { ModelsError } from "store/middleware/types";
 import urls from "urls";
 
 import { JujuEnv, test } from "../fixtures/setup";
@@ -69,6 +70,10 @@ test.describe("Add model", () => {
     await page
       .getByRole("button", { name: AddModelLabel.CREATE_BUTTON })
       .click();
+    await expect(page).toHaveURL(urls.models.index);
+
+    // Reloading the page before checking fetches the updated list of models.
+    await owner.reloadDashboard(page);
     await expect(
       page
         .locator("tr", { hasText: currentModel.name })
@@ -120,6 +125,10 @@ test.describe("Add model", () => {
     await page
       .getByRole("button", { name: AddModelLabel.CREATE_BUTTON })
       .click();
+    await expect(page).toHaveURL(urls.models.index);
+
+    // Reloading the page before checking fetches the updated list of models.
+    await owner.reloadDashboard(page);
     await expect(
       page
         .locator("tr", { hasText: currentModel.name })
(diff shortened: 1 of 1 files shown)


e2e/base/add-model.spec.ts
```diff
@@ -167,6 +176,19 @@ test.describe("Add model", () => {
...
+      // "Unable to list or update models" error banner. In that case, reload the page to recover.
+      if (
+        await page
+          .locator('[role="alert"]', {
+            hasText: ModelsError.LIST_OR_UPDATE_MODELS,
+          })
+          .isVisible()
+      ) {
+        await owner.reloadDashboard(page);
+      }
```
Ninfa-Jeon, line 191: I have made this conditional for two tests in this suite because they have chained operations soon after model creation succeeds, i.e., setting up a block or sharing the model. Simply reloading the page after navigating to the models list sometimes doesn't give the dashboard enough time to even trigger the chained operation. This is why, we only reload once we see the error banner.

charms/k8s-charm/src/config.py
```diff
@@ -113,6 +113,7 @@ def write(self, write_config_js=True, write_nginx=True, write_index=True):
             ),
             dashboard_root=os.environ.get("DASHBOARD_ROOT", "/srv"),
             port=int(os.environ.get("DASHBOARD_PORT", 8080)),
+            base_app_url=os.environ.get("DASHBOARD_BASE_APP_URL", ""),
```
Ninfa-Jeon, line 117: The fix defaults `base_app_url` to `""` (matching what `charm.py` uses when there's no ingress), which makes nginx render correct location blocks and the `<base href="">` renders as empty, which works fine when the app is served at the root path.

Previously, `index.charm.html` was never emitted to `build`, so it never existed at `/srv/index.charm.html` in the Docker container, and `config.py`'s `env.get_template("index.charm.html")` threw a TemplateNotFound exception.

Looking at `config.py`'s `write()` method, the exception would propagate up through `config.write(...)` in `__main__`, which has no `try/except`, so the whole `else` block would crash. Crucially, `write_index=True` never ran, so the Vite-built `index.html` (which had absolute asset paths before setting base to `/`) was left untouched and served correctly.

After setting base to `/`, `index.charm.html` was added to the Vite build output and so `get_template` succeeds → `config.py` overwrites `/srv/index.html`. This exposed the `base_app_url=None` bug that had always been latent in `config.py`'s `__main__` causing a 404. 

This can be confirmed by looking at the build output in [this](https://github.com/canonical/juju-dashboard/actions/runs/29665070164/job/88134025001) run on main: 

Image

vs. the build output in this PR:

Image

andogq: This seems to be the actual solution, however this comment does not relate to the fix. I've adapted this change back into #2408, with proper docs.

e2e/base/add-model.spec.ts
andogq, line 1: This seems like an OK change, but I'll let you pull it into a standalone PR/rebase without the other changes, so we can look at it in isolation!

Ninfa-Jeon:
Closing this PR in favour of the following:
1. The base app url fix has been adopted in another PR and upstreamed: https://github.com/canonical/juju-dashboard/pull/2408
2. The failing workflow run linked in the description is fixed in this PR instead https://github.com/canonical/juju-dashboard/pull/2413
3. The fix introduced here might not be needed once the other fix gets merged (this PR can be opened again if this specific issue reappears)

PR #2433, opened 2026-08-10
chore(deps): major dep updates

huwshimi:
## Done

Major dep updates from https://github.com/canonical/juju-dashboard/pull/2325 sans eslint + typescript changes.


actions/package.json
```diff
@@ -16,7 +16,7 @@
     "@actions/github": "9.1.1"
   },
   "devDependencies": {
-    "@octokit/types": "16.0.0",
+    "@octokit/types": "17.0.0",
```
andogq, line 19: `16.0.0` will still be a transitive dep until `@actions/github` releases a new version (as it installs `@octokit/types`), but it's not a massive deal given it's a dev dep.

andogq (approved):
One small thought about octokit, but still good to merge.

actions/package.json
huwshimi: Ah OK, thanks.

PR #2436, opened 2026-08-12
fix: model table interactions

### LINKED ISSUE #2404: Destroying model's actions are enabled in the model's list page
I am able to interact with the deleting model on the model's list page.

1. Go to /models
2. Click on try to delete a model
3. The row for deleting model should be enabled when the destruction is in progress 

**Expected behavior**
All actions and the row for such a model should be disabled

**Screenshots**

Image

Ninfa-Jeon:
## Done

- Added disabled and loading states to `DataTable` rows
- Added disabled state to `ModelActions` for when an upgrade is in progress
- Moved the spinner rendered during loading (upgrading) or disabling (destroying) to be shown in the checkbox column (as per the latest [screens](https://www.figma.com/design/Rs1v3IS0oY2RK4CU4yo4Ly/25.10-JAAS-model-management---Juju?node-id=452-44876&p=f&m=dev))
- Updated styles and tests accordingly
- Added tests to improve coverage

## QA

- Pull code
- Open http://0.0.0.0:8036/
- View the model table in selectable and non-selectable states
- Try deleting and upgrading models in each
- A spinner should appear under checkbox column in selectable state
- A spinner should appear in the hidden icon column in non-selectable state
- While destroying, the entire row should stay disabled
- While upgrading, the the model actions should stay disabled while the row is open to other interactions
- Try interacting with the select-all toggle when some rows are in destroying and upgrading states. It should ignore such rows.

## Details

Fixes https://github.com/canonical/juju-dashboard/issues/2404
Fixes https://warthogs.atlassian.net/browse/JUJU-10223

## Screenshots

BEFORE:
Screenshot 2026-08-12 at 5 58 34 PM
Screenshot 2026-08-12 at 5 58 55 PM

AFTER:
Screenshot 2026-08-12 at 5 56 03 PM
Screenshot 2026-08-12 at 5 56 54 PM
Screenshot 2026-08-12 at 5 56 58 PM

### DIFF
--- a/src/components/DataTable/DataTable.test.tsx
+++ b/src/components/DataTable/DataTable.test.tsx
@@ -1,0 +1,140 @@
+import { screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import { renderComponent } from "testing/utils";
+
+import DataTable from "./DataTable";
+import type { DataTableProps } from "./types";
+
+describe("DataTable", () => {
+  let initialProps: DataTableProps<
+    { id: string; name: string; age: number },
+    string,
+    Record<string, number | string>
+  >;
+
+  beforeEach(() => {
+    initialProps = {
+      getKey: ({ id }): string => id,
+      columns: {
+        name: { header: "Name", map: ({ name }): string => name },
+        age: { header: "Age", map: ({ age }): number => age },
+      },
+      data: [
+        { id: "1", name: "Alice", age: 30 },
+        { id: "2", name: "Bob", age: 25 },
+      ],
+    };
+  });
+
+  it("renders properly with column headers and row data", () => {
+    renderComponent(<DataTable {...initialProps} />);
+    expect(
+      screen.getByRole("columnheader", { name: "Name" }),
+    ).toBeInTheDocument();
+    expect(
+      screen.getByRole("columnheader", { name: "Age" }),
+    ).toBeInTheDocument();
+    expect(screen.getByRole("cell", { name: "Alice" })).toBeInTheDocument();
+    expect(screen.getByRole("cell", { name: "Bob" })).toBeInTheDocument();
+  });
+
+  it("renders the no rows message when data is empty", () => {
+    renderComponent(<DataTable {...initialProps} data={[]} />);
+    expect(screen.getByText("No rows")).toBeInTheDocument();
+  });
+
+  it("renders a custom no rows message", () => {
+    renderComponent(
+      <DataTable {...initialProps} data={[]} noRowsMessage="Nothing here" />,
+    );
+    expect(screen.getByText("Nothing here")).toBeInTheDocument();
+  });
+
+  describe("selectable", () => {
+    it("renders the select-all checkbox when the rows are selectable", () => {
+      renderComponent(<DataTable {...initialProps} />);
+      expect(
+        screen.getByRole("checkbox", { name: "Select all" }),
+      ).toBeInTheDocument();
+    });
+
+    it("does not render the select-all checkbox when selectable=false", () => {
+      renderComponent(<DataTable {...initialProps} selectable={false} />);
+      expect(
+        screen.queryByRole("checkbox", { name: "Select all" }),
+      ).not.toBeInTheDocument();
+    });
+
+    it("shows 'Deselect all' label when all rows are selected", async () => {
+      renderComponent(<DataTable {...initialProps} />);
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      await userEvent.click(selectAll);
+      expect(
+        screen.getByRole("checkbox", { name: "Deselect all" }),
+      ).toBeInTheDocument();
+    });
+
+    it("shows indeterminate state when some rows are selected", async () => {
+      renderComponent(<DataTable {...initialProps} />);
+      const rowCheckboxes = screen.getAllByRole("checkbox", {
+        name: /Select|Deselect/,
+      });
+      // Click a single row checkbox (skip the select-all at index 0)
+      await userEvent.click(rowCheckboxes[1]);
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      expect((selectAll as HTMLInputElement).indeterminate).toBe(true);
+    });
+  });
+
+  describe("isRowDisabled", () => {
+    it("marks a row as disabled", () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      const tableRows = screen.getAllByRole("row");
+      const aliceRow = tableRows.find((row) =>
+        row.textContent?.includes("Alice"),
+      );
+      const bobRow = tableRows.find((row) => row.textContent?.includes("Bob"));
+      expect(aliceRow).toHaveAttribute("aria-disabled", "true");
+      expect(bobRow).not.toHaveAttribute("aria-disabled", "true");
+    });
+
+    it("excludes disabled rows from select all", async () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      await userEvent.click(selectAll);
+      expect(screen.getByRole("checkbox", { name: "Select 2" })).toBeChecked();
+      expect(
+        screen.queryByRole("checkbox", { name: "Select 1" }),
+      ).not.toBeInTheDocument();
+    });
+  });
+
+  describe("isRowLoading", () => {
+    it("excludes loading rows from select all", async () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowLoading={(row) => row.id === "1"} />,
+      );
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      await userEvent.click(selectAll);
+      expect(screen.getByRole("checkbox", { name: "Select 2" })).toBeChecked();
+    });
+  });
+});
+
+describe("DataTable.columnBuilder", () => {
+  it("returns a column definition unchanged", () => {
+    const builder = DataTable.columnBuilder<{
+      id: string;
+      name: string;
+      age: number;
+    }>();
+    const col = builder({ header: "Name", map: (row) => row.name });
+    expect(col.header).toBe("Name");
+    expect(col.map({ id: "1", name: "Alice", age: 30 })).toBe("Alice");
+  });
+});
--- a/src/components/DataTable/DataTable.test.tsx
+++ b/src/components/DataTable/DataTable.test.tsx
@@ -1,4 +1,4 @@
-import { screen } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import { renderComponent } from "testing/utils";
@@ -68,8 +68,9 @@ describe("DataTable", () => {
 
     it("shows 'Deselect all' label when all rows are selected", async () => {
       renderComponent(<DataTable {...initialProps} />);
-      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
-      await userEvent.click(selectAll);
+      await userEvent.click(
+        screen.getByRole("checkbox", { name: "Select all" }),
+      );
       expect(
         screen.getByRole("checkbox", { name: "Deselect all" }),
       ).toBeInTheDocument();
@@ -77,13 +78,72 @@ describe("DataTable", () => {
 
     it("shows indeterminate state when some rows are selected", async () => {
       renderComponent(<DataTable {...initialProps} />);
-      const rowCheckboxes = screen.getAllByRole("checkbox", {
-        name: /Select|Deselect/,
-      });
-      // Click a single row checkbox (skip the select-all at index 0)
-      await userEvent.click(rowCheckboxes[1]);
+      // Click a single row checkbox — initially unchecked rows have "Deselect <key>" label.
+      await userEvent.click(
+        screen.getByRole("checkbox", { name: "Deselect 1" }),
+      );
       const selectAll = screen.getByRole("checkbox", { name: "Select all" });
       expect((selectAll as HTMLInputElement).indeterminate).toBe(true);
+    });
+
+    it("clearing select-all does not affect a disabled row's checkbox", async () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      // Select all selectable rows (only row 2).
+      await userEvent.click(
+        screen.getByRole("checkbox", { name: "Select all" }),
+      );
+      expect(screen.getByRole("checkbox", { name: "Select 2" })).toBeChecked();
+
+      // Now deselect all — row 1's checkbox should still be unchecked, unchanged.
+      await userEvent.click(
+        screen.getByRole("checkbox", { name: "Deselect all" }),
+      );
+      expect(
+        screen.getByRole("checkbox", { name: "Deselect 2" }),
+      ).not.toBeChecked();
+      expect(
+        screen.getByRole("checkbox", { name: "Deselect 1" }),
+      ).not.toBeChecked();
+    });
+
+    it("'Select all' does nothing when all rows are disabled", async () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowDisabled={() => true} />,
+      );
+      // Clicking it does nothing — both row checkboxes remain unchecked.
+      await userEvent.click(
+        screen.getByRole("checkbox", {
+          name: "Select all",
+        }),
+      );
+      expect(
+        screen.getByRole("checkbox", { name: "Select all" }),
+      ).toBeInTheDocument();
+      const rowCheckboxes = screen.getAllByRole("checkbox", {
+        name: /Deselect [^a]/,
+      });
+      rowCheckboxes.forEach((checkbox) => expect(checkbox).not.toBeChecked());
+    });
+
+    it("deselects a previously selected row when it becomes disabled", async () => {
+      const { rerender } = render(
+        <DataTable {...initialProps} isRowDisabled={() => false} />,
+      );
+      // Select row 1.
+      await userEvent.click(
+        screen.getByRole("checkbox", { name: "Deselect 1" }),
+      );
+      expect(screen.getByRole("checkbox", { name: "Select 1" })).toBeChecked();
+
+      // Now disable row 1 — it should be removed from the selection.
+      rerender(
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      expect(
+        screen.getByRole("checkbox", { name: "Deselect 1" }),
+      ).not.toBeChecked();
     });
   });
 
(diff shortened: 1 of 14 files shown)


src/components/ModelTable/_model-table.scss
```diff
@@ -13,19 +13,21 @@
...
 
-    &__name {
-      flex-shrink: 1;
-      overflow: hidden;
-    }
+  tr.is-disabled {
+    opacity: 0.5;
+    pointer-events: none;
+    user-select: none;
   }
```
andogq: This class is set in `useRows.tsx` (within the `DataTable` component), but it's only included in this SCSS file. If this is the approach to be used, this class must be defined in a SCSS file specific for `DataTable`.

Additionally, this is not an accessible way to 'disable' a row (pointer-events and user-select). It would be better to directly disable the interactable elements within the row (buttons/inputs/etc).

src/components/ModelTable/_model-table.scss
```diff
@@ -13,19 +13,21 @@
...
+    width: 2rem;
+  }
 
-    &__status {
-      flex-shrink: 0;
-      margin-right: $sph--small;
-    }
+  .icon-column-header {
+    padding-top: 0.9rem;
+  }
```
andogq: These classes are use within `DataTable`, so should be defined there.

andogq (commented):
My main question here is what does it mean for a row to be 'disabled'?

If it's intended to disable certain interactable items within the row (eg. the context menu, or a button), then the disable attribute should be directly added to them within the `renderCell` callback in the column. If it helps, the entire row data is passed as the second parameter to `renderCell`, so it can be directly referenced in every render call.

If it's intended to disable the checkbox to select a row, then that comes with its own set of problems and questions:
- If the row was selected and disabled, and the user toggles "Select all" (to clear the selection), does this checkbox change state?
- If the only row is selected and disabled, now what does the "Select all" button do?
- Is it clear to the user that a disabled row is correlated with the model being deleted (as opposed to a permission problem, a connectivity problem, etc)?

Instead, I propose the following changes:
- If a specific thing in a cell should be disabled, disable it directly in the `renderCell` method. This includes making text grey, disabling links, buttons, etc.
- Don't alter any selection logic or add any row disabling to `DataTable`. Instead, check for disabled items before doing an action with them. Eg. if we had bulk model deletion, and a model in the process of being deleted is selected, disable the bulk-delete button with a message ('cannot delete model that is already being deleted').
- Depending on Lukas' thoughts on spinner placement, it would be good to have it as an icon inline with a cell, such as below:

```jsx
          loading: column({
              header: "",
              className: "p-table__cell--icon-placeholder",
              collapse: true,
              // TODO: Determine loading state here.
              map: () => false,
              renderCell: (loading) => {
                  return loading ? <Spinner /> : null;
              }
          }),
```

image

src/components/ModelTable/_model-table.scss
Ninfa-Jeon: done

src/components/ModelTable/_model-table.scss
Ninfa-Jeon: done

Ninfa-Jeon:
> If it's intended to disable certain interactable items within the row (eg. the context menu, or a button), then the disable attribute should be directly added to them within the `renderCell` callback in the column. If it helps, the entire row data is passed as the second parameter to `renderCell`, so it can be directly referenced in every render call.

The entire row needs to be rendered with 0.5 opacity besides the interactable items disabled.

> 
> If it's intended to disable the checkbox to select a row, then that comes with its own set of problems and questions:
> 
> * If the row was selected and disabled, and the user toggles "Select all" (to clear the selection), does this checkbox change state?

the disabled row is automatically deselected, and "select all" correctly ignores disabled rows

> * If the only row is selected and disabled, now what does the "Select all" button do?

The select-all checkbox becomes a no-op. I could potentially disable the "Select all" button for such a case.

> * Is it clear to the user that a disabled row is correlated with the model being deleted (as opposed to a permission problem, a connectivity problem, etc)?

Yes, because we show the message "Destroying model..." under it

> 
> Instead, I propose the following changes:
> 
> * If a specific thing in a cell should be disabled, disable it directly in the `renderCell` method. This includes making text grey, disabling links, buttons, etc.

Since the entire row needs to change its behaviour, disabling the row element from within the `DataTable` is most suitable. As opposed to per cell disabling where we wouldn't be able to set the `aria-disabled` from `renderCell`.

> * Don't alter any selection logic or add any row disabling to `DataTable`. Instead, check for disabled items before doing an action with them. Eg. if we had bulk model deletion, and a model in the process of being deleted is selected, disable the bulk-delete button with a message ('cannot delete model that is already being deleted').

As confirmed previously, the selection logic needed updates due to the latest designs.

src/components/ModelTable/_model-table.scss
Ninfa-Jeon: splitting the disabling rules between row opacity and pointer-events/user-select has introduced somewhat redundant code. Let me know how you would like to proceed on that.

src/components/DataTable/_data-table.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  .is-row-disabled {
+    opacity: 0.5;
+  }
+
+  // Apply to interactive elements within a disabled row via renderCell.
+  .is-disabled {
+    pointer-events: none;
+    user-select: none;
+  }
```
andogq: We should use the `inert` HTML property instead, as it has significantly better semantics for assistive technologies.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

src/components/DataTable/_data-table.scss
```diff
@@ -0,0 +1,22 @@
+.data-table {
+  .icon-column {
```
andogq, line 2: Should rely on vanilla styles for icon styling in tables: https://vanillaframework.io/docs/base/tables#icons

src/components/DataTable/DataTable.tsx
```diff
@@ -87,19 +100,19 @@ export default function DataTable<
   );
 
   return (
-    <Table>
+    <Table className="data-table">
       <Table.Header>
         {prefixHeader}
-        {selectable ? (
-          <Table.Header.Cell collapse>
+        <Table.Header.Cell className="icon-column icon-column-header">
```
andogq, line 106: Should also use the material icon styles. Also having an empty header cell is really not ideal for accessibility.

src/components/DataTable/useRows.tsx
```diff
@@ -95,7 +105,8 @@ export default function useRows<TRow, TKey extends React.Key, TValues>({
   const fullRows = useMemo(
     () =>
       sortedRows.map((row, i) => {
-        const checked = select.selected.includes(row.key);
+        const { disabled, loading, key } = row;
```
andogq: Calls to `isRowDisabled` and `isRowLoading` should be done here instead, to prevent re-calculating and rendering all of the rows and sorting whenever something is enabled or disabled.

src/components/DataTable/useRows.tsx
```diff
@@ -123,22 +134,27 @@ export default function useRows<TRow, TKey extends React.Key, TValues>({
...
 
-        const label = checked ? `Select ${row.key}` : `Deselect ${row.key}`;
+        const label = checked ? `Select ${key}` : `Deselect ${key}`;
 
         return (
-          <Table.Row key={row.key}>
+          <Table.Row
+            key={key}
+            aria-disabled={disabled || undefined}
+            className={classNames({ "is-row-disabled": disabled })}
```
andogq: This should be the `inert` attribute instead.

src/components/DataTable/useRows.tsx
```diff
@@ -123,22 +134,27 @@ export default function useRows<TRow, TKey extends React.Key, TValues>({
...
+          >
             {prefixColumn}
-            {selectable ? (
-              <Table.Cell>
+            <Table.Cell className="icon-column">
```
andogq, line 146: This should be the vanilla icon class.

src/components/InlineCheckbox/InlineCheckbox.tsx
```diff
@@ -33,7 +33,7 @@ export default function InlineCheckbox<L extends string>({
   }, [indeterminate]);
 
   return (
-    <label className="p-checkbox">
+    <label className="p-checkbox u-no-padding--top">
```
andogq, line 36: What's the need for this to be overridden, only to be re-added manually with the icon column class?

src/components/ModelTable/ModelTable.tsx
```diff
@@ -93,44 +99,32 @@ export default function ModelTable({ models, groupBy }: Props): JSX.Element {
...
-                  </div>
-                </div>
+                <TruncatedTooltip
+                  message={name}
+                  className={classNames({
+                    "is-disabled": isDying(uuid) || !!getUpgrade(uuid),
+                  })}
+                >
+                  <ModelDetailsLink modelName={name} qualifier={qualifier}>
+                    {model.model.name}
```
andogq: Why not use `name` here?

src/components/ModelTable/ModelTable.tsx
```diff
@@ -93,44 +99,32 @@ export default function ModelTable({ models, groupBy }: Props): JSX.Element {
...
-                          className="models__version"
-                          modelName={name}
-                          qualifier={qualifier}
-                        />
-                      )}
-                    </TruncatedTooltip>
-                  </div>
-                </div>
+                <TruncatedTooltip
+                  message={name}
```
andogq: Would be good to include the version in the tooltip. Eg: `model name (3.6)`

src/components/ModelTable/ModelTable.tsx
andogq, line 1: All of the class names setting `is-disabled` in here are not needed. Rely on the `inert` attribute being set by the table row in the data table, and the styling associated with that.

src/components/ModelActions/ModelActions.test.tsx
```diff
@@ -120,6 +120,21 @@ describe("ModelActions", () => {
...
+  it("displays the actions menu in disabled state", () => {
+    renderComponent(
+      <ModelActions
+        qualifier="eggman@external"
+        modelUUID="abc123"
+        modelName="test-model"
+        disabled
+      />,
+    );
+    expect(screen.getByRole("button", { name: Label.TOGGLE })).toHaveAttribute(
```
andogq, line 132: Should be testing if it's disabled: https://vitest.dev/api/browser/assertions.html#tobedisabled

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+    it("does not render the select-all checkbox when selectable=false", () => {
+      renderComponent(<DataTable {...initialProps} selectable={false} />);
+      expect(
+        screen.queryByRole("checkbox", { name: "Select all" }),
+      ).not.toBeInTheDocument();
+    });
+
+    it("shows 'Deselect all' label when all rows are selected", async () => {
+      renderComponent(<DataTable {...initialProps} />);
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
```
andogq: Why is this bound if it's only used once?

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+      await userEvent.click(selectAll);
+      expect(
+        screen.getByRole("checkbox", { name: "Deselect all" }),
+      ).toBeInTheDocument();
+    });
+
+    it("shows indeterminate state when some rows are selected", async () => {
+      renderComponent(<DataTable {...initialProps} />);
+      const rowCheckboxes = screen.getAllByRole("checkbox", {
+        name: /Select|Deselect/,
```
andogq: Check for specific variant, it should be deterministic. In this case, it'll be "Select" initially.

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+      renderComponent(
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      const tableRows = screen.getAllByRole("row");
+      const aliceRow = tableRows.find((row) =>
+        row.textContent?.includes("Alice"),
+      );
+      const bobRow = tableRows.find((row) => row.textContent?.includes("Bob"));
+      expect(aliceRow).toHaveAttribute("aria-disabled", "true");
+      expect(bobRow).not.toHaveAttribute("aria-disabled", "true");
```
andogq, line 161: Use `toBeDisabled` here too

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <DataTable {...initialProps} isRowDisabled={(row) => row.id === "1"} />,
+      );
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      await userEvent.click(selectAll);
+      expect(screen.getByRole("checkbox", { name: "Select 2" })).toBeChecked();
+      expect(
+        screen.queryByRole("checkbox", { name: "Select 1" }),
+      ).not.toBeInTheDocument();
+    });
+  });
```
andogq, line 175: We need more tests for all the edge cases I have previously mentioned relating to the interaction between disabled/loading/selecting rows.

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  describe("isRowLoading", () => {
+    it("excludes loading rows from select all", async () => {
+      renderComponent(
+        <DataTable {...initialProps} isRowLoading={(row) => row.id === "1"} />,
+      );
+      const selectAll = screen.getByRole("checkbox", { name: "Select all" });
+      await userEvent.click(selectAll);
+      expect(screen.getByRole("checkbox", { name: "Select 2" })).toBeChecked();
```
andogq, line 184: Actually need to check that the first row is not selected too.

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+describe("DataTable.columnBuilder", () => {
+  it("returns a column definition unchanged", () => {
+    const builder = DataTable.columnBuilder<{
+      id: string;
+      name: string;
+      age: number;
+    }>();
+    const col = builder({ header: "Name", map: (row) => row.name });
+    expect(col.header).toBe("Name");
```
andogq: we can check for identity here.

src/components/DataTable/useRows.test.tsx
```diff
@@ -103,22 +104,60 @@ describe("useRows", () => {
...
+        createProps([{ key: "abc123", value: 123, otherValue: true }], {
+          selectable,
+          isRowLoading: () => loading,
+        }),
+      );
+
+      const row = result.current[0] as React.ReactElement<{
+        children: React.ReactNode[];
+      }>;
+      const [, childSelectable] = row.props.children;
```
andogq:
```suggestion
      const [_, childSelectable] = row.props.children;
```
use a placeholder here to keep it clear.

andogq (changes requested):
- Looks like there's lint issues that need to be addressed.
- Get rid of `is-dying` class, and use the `inert` attribute set on the actual table row. Then target the elements using that attribute in css.
- Try use the Vanilla table icon styling instead of manually creating types and overriding styles.

src/components/ModelTable/ModelTable.tsx
Ninfa-Jeon: Pretty sure that was already there, good call out

src/components/DataTable/_data-table.scss
Ninfa-Jeon: Isn't that just for icon placeholders though?

src/components/DataTable/DataTable.tsx
Ninfa-Jeon: This is also why I previously mentioned losing the `selectable` prop entirely. I will start work on the batch destruction coming week anyway.

src/components/InlineCheckbox/InlineCheckbox.tsx
Ninfa-Jeon: That is because the `icon-column` class is making the icon header as well as the icon row cells appear well aligned with the other cells. PFA with/without my approach:
Screenshot 2026-08-20 at 8 22 05 PM Screenshot 2026-08-20 at 8 22 41 PM

Minor difference really, but it was pmo XD

src/components/ModelTable/ModelTable.tsx
Ninfa-Jeon: This needed extra unrelated changes (request for raise)

src/components/ModelActions/ModelActions.test.tsx
Ninfa-Jeon: If I remember correctly this assertion doesn't seem to work well for our codebase. Will try anyway

src/components/DataTable/DataTable.test.tsx
Ninfa-Jeon: the tests I wrote are for extra coverage, we can improve it further in another PR :)

src/components/DataTable/_data-table.scss
andogq: From my testing, it seems to work for anything that is icon-like (particularly in sizing), including checkboxes and spinners. Wouldn't hurt to try, as it would be less for us to maintain.

src/components/DataTable/DataTable.test.tsx
andogq: This selector still needs to be changed

src/components/DataTable/DataTable.test.tsx
andogq: That's fine, but it's good practice to have tests covering the features that are added in a PR. There shouldn't be many needed, maybe 4 or 5 to cover the edge cases we've discussed.

src/components/DataTable/DataTable.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+    const builder = DataTable.columnBuilder<{
+      id: string;
+      name: string;
+      age: number;
+    }>();
+    const mapFn = (row: { id: string; name: string; age: number }): string =>
+      row.name;
+    const col = builder({ header: "Name", map: mapFn });
+    expect(col.header).toBe("Name");
+    expect(col.map).toBe(mapFn);
```
andogq:
```suggestion
    const colRaw = { header: "Name", map: mapFn };
    const col = builder(colRaw);
    expect(col).toBe(colRaw);
```

Since the builder function just returns whatever it was passed (it's the identity function), we can perform a direct comparison between the two objects.

andogq (approved):
Nearly there. A few more things that you need to do, so please make sure they're done before merging.
- Test the usage of the vanilla icon class, it'd be very convenient if it works. If it doesn't work, then that's ok.
- Use the `isDisabled` attribute in tests
- Add missing tests for the edge cases we've discussed. Since they're unexpected and tricky to reason about, tests are important to ensure that whatever behaviour we end with is consistent and repeatable.

src/components/ModelActions/ModelActions.test.tsx
Ninfa-Jeon: As stated above, I can confirm that this assertion doesn't work on our codebase as it doesn't take into account the `aria-disabled` attribute when checking for disabled state.

src/components/DataTable/DataTable.test.tsx
Ninfa-Jeon: Doesn't work with `toBeDisabled` selector.

src/components/DataTable/_data-table.scss
Ninfa-Jeon: The column doesn't look right to me when the class is replaced with `.p-table__cell--icon-placeholder`. Maybe I am not understanding what you're asking for. I will follow up with another PR for this once you explain it to me.

PR #2438, opened 2026-08-13
chore(security): Add eslint security plugin

huwshimi:
## Done

- Added eslint security plugin
- Fix issues.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -15,6 +15,8 @@ import tseslint from "typescript-eslint";
 import { defineConfig } from "eslint/config";
 import { includeIgnoreFile } from "@eslint/compat";
 import cspellPlugin from "@cspell/eslint-plugin";
+import sdl from "@microsoft/eslint-plugin-sdl";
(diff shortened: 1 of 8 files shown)


eslint.config.mjs
```diff
@@ -183,6 +199,7 @@ export default defineConfig(
       "@typescript-eslint/require-await": "off",
       "@typescript-eslint/no-unsafe-assignment": "off",
       "@typescript-eslint/init-declarations": "off",
+      "@microsoft/sdl/no-insecure-url": "off",
```
andogq, line 202: If this has been disabled, do we need all the other `eslint-disable-line` entries?

andogq (approved):
One tiny change

eslint.config.mjs
huwshimi: This block only applies to test files, it's still enabled everywhere else.

andogq: whoops, you're right. :shipit:

PR #2467, opened 2026-09-01
feat: add destroy models panel

Ninfa-Jeon:
Depends on https://github.com/canonical/juju-dashboard/pull/2466

## Done

- Introduced a new panel for bulk model destruction
- Introduced a store state for saving models selected for bulk destruction
- Added logic for switching experiences based on the number of models selected
  - If one selected, show single model destruction modal
  - If multiple selected, show the new flow with a side panel
- Added tests 

Note: The side panel is currently stubbed for configurations within the model. They will be added in https://warthogs.atlassian.net/browse/JUJU-10310

## QA

- Pull code
- Open http://0.0.0.0:8036/
- On the models list page, try interacting with the checkboxes in the table and beginning the destroy flow
- Do it 
  - with a single model
  - with multiple models
  - from model actions
  - using the `Review & Destroy` button
- Verify the information on the button, the labels in the side panel, etc.
- Verify that only the single destroy can be triggered

## Details

https://warthogs.atlassian.net/browse/JUJU-10309

## Screenshots

https://github.com/user-attachments/assets/182b7bf1-e0b0-473f-92ad-2935417e850a

### DIFF
--- a/src/pages/ModelsIndex/ModelsIndex.tsx
+++ b/src/pages/ModelsIndex/ModelsIndex.tsx
@@ -3,14 +3,16 @@ import {
   Icon,
   Notification as ReactNotification,
   SearchAndFilter,
+  usePortal,
 } from "@canonical/react-components";
 import type { SearchAndFilterChip } from "@canonical/react-components/dist/components/SearchAndFilter/types";
 import classNames from "classnames";
 import type { JSX, ReactNode } from "react";
-import { useId, useMemo, useState } from "react";
+import { useCallback, useId, useMemo, useState } from "react";
 import { Link, useNavigate } from "react-router";
 
 import ChipGroup from "components/ChipGroup";
+import DestroyModelDialog from "components/DestroyModelDialog";
 import LoadingSpinner from "components/LoadingSpinner";
 import ModelTable from "components/ModelTable";
 import SegmentedControl from "components/SegmentedControl";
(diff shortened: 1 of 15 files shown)


src/pages/ModelsIndex/ModelsIndex.tsx
```diff
@@ -100,6 +117,34 @@ export default function Models(): JSX.Element {
...
+
+  const handleReviewAndDestroy = useCallback(() => {
+    if (selectedModelUUIDs.length === 1) {
+      // Single model: open the existing per-model dialog directly.
+      openPortal();
+    } else {
+      // Multiple models: seed the store and open the destroy models panel.
+      if (wsControllerURL) {
+        dispatch(
+          jujuActions.selectModelsForDestruction({
```
huwshimi, line 129: This state should probably be cleaned up when the panel closes.

huwshimi (approved):
One thing (possibly as a followup) is it that you can select models that you don't have permission to destroy. The destroy model action is disabled in the actions menu, but the checkbox is enabled.

Ninfa-Jeon:
> One thing (possibly as a followup) is it that you can select models that you don't have permission to destroy. The destroy model action is disabled in the actions menu, but the checkbox is enabled

Yes! We have decided to let the user select all models from the table. The ones that cannot be deleted will be handled inside the sidebar. Effectively, the user still won't be able to delete those models. Just select them; to make the experience uniform and without complicating the checkbox logic :)

src/pages/ModelsIndex/ModelsIndex.tsx
Ninfa-Jeon: Done

PR #2470, opened 2026-09-02
feat: populate batch destroy drawer with model info

Ninfa-Jeon:
## Done

- Created `AccordionContent` component with info table and action buttons (the buttons don't do anything currently)
- Created `AccordionTitle` component that can show striked items based on whether they are deletable
- Expanded `useCanConfigureModelWithUUID` to provide `unitCount` and `destroyBlockedReason`
- Added/updated tests accordingly

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Try interacting with the bulk destruction flow with models having various resources (applications, CMRS, storage)
- The model info for each such model inside the drawer's accordion should be correctly rendered
- Empty models shouldn't have a table

Note: Edge cases regarding empty models, button gating, tooltip labels, etc will be done in https://warthogs.atlassian.net/browse/JUJU-10324

## Details
https://warthogs.atlassian.net/browse/JUJU-10310

## Screenshots

https://github.com/user-attachments/assets/b0e84387-5a02-467f-b2ee-6e79bf37efb9

### DIFF
--- a/src/components/Panel/_panel.scss
+++ b/src/components/Panel/_panel.scss
@@ -36,6 +36,7 @@
     padding: $spv--large $sph--small 0;
     position: sticky;
     text-align: right;
+    z-index: z("gamma");
 
     @include mobile {
       padding-left: 0;
--- a/src/hooks/useModelDestructionData.test.tsx
+++ b/src/hooks/useModelDestructionData.test.tsx
@@ -3,6 +3,7 @@ import type { JSX, PropsWithChildren } from "react";
 import { Provider } from "react-redux";
 import { BrowserRouter, Route, Routes } from "react-router";
 
+import * as useCanConfigureModelModule from "hooks/useCanConfigureModel";
 import type { RootState } from "store/store";
 import { rootStateFactory } from "testing/factories";
 import {
@@ -19,7 +20,9 @@ import {
 } from "testing/factories/juju/juju";
 import { createStore } from "testing/utils";
 
-import useModelDestructionData from "./useModelDestructionData";
+import useModelDestructionData, {
+  DestroyBlockedReason,
+} from "./useModelDestructionData";
 
 const generateContainer =
   (state: RootState, path: string, url: string) =>
@@ -38,6 +41,13 @@ const generateContainer =
   };
 
 describe("useModelDestructionData", () => {
+  beforeEach(() => {
+    vi.spyOn(
+      useCanConfigureModelModule,
+      "useCanConfigureModelWithUUID",
+    ).mockReturnValue(true);
+  });
+
   it("should return initial empty state when modelStatusData is null or empty", () => {
     const state = rootStateFactory.build({
       juju: jujuStateFactory.build({
@@ -56,6 +66,8 @@ describe("useModelDestructionData", () => {
     expect(result.current.connectedOffers).toEqual([]);
     expect(result.current.storageIDs).toEqual([]);
     expect(result.current.showInfoTable).toBe(false);
+    expect(result.current.unitCount).toBe(0);
+    expect(result.current.destroyBlockedReason).toBe(DestroyBlockedReason.NONE);
   });
 
   it("should correctly count applications and machines and set showInfoTable to true", () => {
@@ -301,5 +313,74 @@ describe("useModelDestructionData", () => {
       wrapper: generateContainer(state, "*", "/models"),
     });
     expect(result.current.unitCount).toBe(0);
+  });
+
+  it("should return destroyBlockedReason as IS_CONTROLLER when model is a controller model", () => {
+    const state = rootStateFactory.build({
+      juju: jujuStateFactory.build({
+        modelData: {
+          abc123: modelDataFactory.build({
+            uuid: "abc123",
+            info: modelInfoFactory.build({ "is-controller": true }),
+          }),
+        },
+      }),
+    });
+
+    const { result } = renderHook(() => useModelDestructionData("abc123"), {
+      wrapper: generateContainer(state, "*", "/models"),
+    });
+    expect(result.current.destroyBlockedReason).toBe(
+      DestroyBlockedReason.IS_CONTROLLER,
+    );
+  });
+
+  it("should return destroyBlockedReason as NO_ACCESS when user cannot configure the model", () => {
+    vi.spyOn(
+      useCanConfigureModelModule,
+      "useCanConfigureModelWithUUID",
+    ).mockReturnValue(false);
+    const state = rootStateFactory.build({
+      juju: jujuStateFactory.build({
+        modelData: {
+          abc123: modelDataFactory.build({
+            uuid: "abc123",
+            info: modelInfoFactory.build({ "is-controller": false }),
+          }),
+        },
+      }),
+    });
+
+    const { result } = renderHook(() => useModelDestructionData("abc123"), {
+      wrapper: generateContainer(state, "*", "/models"),
+    });
+    expect(result.current.destroyBlockedReason).toBe(
+      DestroyBlockedReason.NO_ACCESS,
+    );
+  });
+
+  it("should return destroyBlockedReason as CONNECTED_OFFERS when there are offers with active connections", () => {
+    const state = rootStateFactory.build({
+      juju: jujuStateFactory.build({
+        modelData: {
+          abc123: modelDataFactory.build({
+            uuid: "abc123",
+            info: modelInfoFactory.build({ "is-controller": false }),
+            offers: {
+              db: applicationOfferStatusFactory.build({
+                "total-connected-count": 1,
+              }),
+            },
+          }),
+        },
+      }),
+    });
+
+    const { result } = renderHook(() => useModelDestructionData("abc123"), {
+      wrapper: generateContainer(state, "*", "/models"),
+    });
+    expect(result.current.destroyBlockedReason).toBe(
+      DestroyBlockedReason.CONNECTED_OFFERS,
+    );
   });
 });
--- a/src/hooks/useModelDestructionData.test.tsx
+++ b/src/hooks/useModelDestructionData.test.tsx
@@ -67,7 +67,7 @@ describe("useModelDestructionData", () => {
     expect(result.current.storageIDs).toEqual([]);
     expect(result.current.showInfoTable).toBe(false);
     expect(result.current.unitCount).toBe(0);
-    expect(result.current.destroyBlockedReason).toBe(DestroyBlockedReason.NONE);
+    expect(result.current.destroyBlockedReason).toBeNull();
   });
 
   it("should correctly count applications and machines and set showInfoTable to true", () => {
--- a/src/hooks/useModelDestructionData.test.tsx
+++ b/src/hooks/useModelDestructionData.test.tsx
@@ -260,4 +260,46 @@ describe("useModelDestructionData", () => {
       },
     });
   });
+
+  it("should sum unit counts across all applications", () => {
+    const modelData = modelDataFactory.build({
+      uuid: "abc123",
+      applications: {
+        easyrsa: applicationStatusFactory.build({
+          units: {
+            "easyrsa/0": unitStatusFactory.build(),
+            "easyrsa/1": unitStatusFactory.build(),
+          },
+        }),
+        mysql: applicationStatusFactory.build({
+          units: {
+            "mysql/0": unitStatusFactory.build(),
+          },
+        }),
+      },
+    });
+    const state = rootStateFactory.build({
+      juju: jujuStateFactory.build({
+        modelData: { abc123: modelData },
+      }),
+    });
+
+    const { result } = renderHook(() => useModelDestructionData("abc123"), {
+      wrapper: generateContainer(state, "*", "/models"),
+    });
+    expect(result.current.unitCount).toBe(3);
+  });
+
+  it("should return unitCount of 0 when there are no applications", () => {
+    const state = rootStateFactory.build({
+      juju: jujuStateFactory.build({
+        modelData: { abc123: modelDataFactory.build({ uuid: "abc123" }) },
+      }),
+    });
+
+    const { result } = renderHook(() => useModelDestructionData("abc123"), {
+      wrapper: generateContainer(state, "*", "/models"),
+    });
+    expect(result.current.unitCount).toBe(0);
+  });
 });
--- a/src/hooks/useModelDestructionData.ts
+++ b/src/hooks/useModelDestructionData.ts
@@ -27,6 +27,7 @@ type ModelDestructionData = {
   connectedOffers: ConnectedOffer[];
   showInfoTable: boolean;
   storageIDs: string[];
+  unitCount: number;
 };
 
 // Helper function to extract and format cross-model relations
@@ -114,6 +115,10 @@ export default function useModelDestructionData(
     machines.length > 0 ||
     crossModelRelations.length > 0;
   const hasStorage = modelStatusData?.storage !== undefined;
+  const unitCount = applications.reduce((prev, key) => {
+    const units = modelStatusData?.applications?.[key]?.units ?? {};
+    return prev + Object.keys(units).length;
+  }, 0);
 
   return {
     hasStorage,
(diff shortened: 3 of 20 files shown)


src/components/Panel/_panel.scss
```diff
@@ -36,6 +36,7 @@
     padding: $spv--large $sph--small 0;
     position: sticky;
     text-align: right;
+    z-index: z("gamma");
```
Ninfa-Jeon, line 39: The panel content was visible over the footer without this.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
```diff
@@ -0,0 +1,188 @@
+import type { RemoteEndpoint } from "@canonical/jujulib/dist/api/facades/application/ApplicationV22";
+import { Button, Icon, MainTable } from "@canonical/react-components";
+import type { MainTableRow } from "@canonical/react-components/dist/components/MainTable/MainTable";
+import { useMemo, type JSX } from "react";
+
+import { useCanConfigureModelWithUUID } from "hooks/useCanConfigureModel";
+import useModelDestructionData from "hooks/useModelDestructionData";
+
+// Helper to render the Applications
+const applicationsRow = (applications: string[]): MainTableRow | null => {
```
Ninfa-Jeon, line 10: The entire logic for rendering this table was pulled from `DestroyModelDialog`. I will try to extract this into a standalone component in a follow-up PR.

src/panels/DestroyModelsPanel/AccordionTitle/AccordionTitle.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <span
+      className={classNames("accordion-title", {
+        "accordion-title--is-removed": isRemoved,
+      })}
+    >
+      <span>
+        {modelName}
+        {isRemoved ? (
+          <Tooltip
+            message="Controller model cannot be deleted"
```
Ninfa-Jeon, line 32: Although there are other ways a model could be removed from this list (no access, CMRs), I am still discussing these tooltips with Lukas. I will make this text dynamic once I have a confirmation.

src/components/Panel/_panel.scss
Ninfa-Jeon: Screenshot 2026-09-03 at 2 35 11 PM

ale8k, line 39: Maybe be careful here, having to use z indexes suggests alignment and/or rendering is wrong and this is a workaround?

src/hooks/useModelDestructionData.test.tsx
```diff
@@ -3,6 +3,7 @@ import type { JSX, PropsWithChildren } from "react";
 import { Provider } from "react-redux";
 import { BrowserRouter, Route, Routes } from "react-router";
 
+import * as useCanConfigureModelModule from "hooks/useCanConfigureModel";
```
ale8k, line 6: The rest of the imports are explicitly named, maybe we could switch this one to be explicit too?

src/hooks/useModelDestructionData.ts
```diff
@@ -114,6 +126,19 @@ export default function useModelDestructionData(
...
+    return prev + Object.keys(units).length;
+  }, 0);
+
+  let destroyBlockedReason = DestroyBlockedReason.NONE;
+  if (modelStatusData?.info?.["is-controller"]) {
+    destroyBlockedReason = DestroyBlockedReason.IS_CONTROLLER;
+  } else if (!canConfigureModel) {
+    destroyBlockedReason = DestroyBlockedReason.NO_ACCESS;
+  } else if (connectedOffers.length > 0) {
+    destroyBlockedReason = DestroyBlockedReason.CONNECTED_OFFERS;
```
ale8k, line 139: I think Juju is going to report this anyway when you attempt to destroy it (double check maybe I am wrong) - if that is true, we're just duplicating the logic here.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
```diff
@@ -0,0 +1,185 @@
...
+      },
+      {
+        content: <div>{machines.join(", ")}</div>,
+        className: "p-table__cell--icon-placeholder",
+      },
+    ],
+  };
+};
+
+// Helper to render the Persistent Storage
```
ale8k, line 100: I'd prob just drops the comments above and rename it to renderStorageRow

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
```diff
@@ -0,0 +1,185 @@
...
+      },
+      {
+        content: (
+          <>
+            {crossModelRelations.map(
+              ({ name, endpoints, isConnectedOffer }) => (
+                <div key={name}>
+                  {isConnectedOffer ? <Icon name="warning" /> : null}
+                  {name}{" "}
+                  {endpoints.map((endpoint: RemoteEndpoint, index: number) => (
```
ale8k, line 61: Would forEach be better here? (For both as we aren't actually mapping anything it seems)

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
```diff
@@ -0,0 +1,185 @@
...
+    destroyBlockedReason,
+  } = useModelDestructionData(modelUUID);
+
+  const infoTableRows = useMemo(() => {
+    return [
+      applicationsRow(applications),
+      crossModelRelationsRow(crossModelRelations),
+      machinesRow(machines),
+      storageRow(hasStorage, storageIDs),
+    ].filter((row): row is Exclude<typeof row, null> => row !== null);
```
ale8k: Never seen this before, I understand we're just filtering nulls but what is this is Exclude stuff?

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
```diff
@@ -0,0 +1,185 @@
...
+            appearance="positive"
+            type="button"
+            onClick={() => {}}
+            hasIcon
+          >
+            <Icon name="success-grey" />
+            <span>Mark as reviewed</span>
+          </Button>
+        </span>
+      ) : null}
```
ale8k, line 181: Not seen much react in a while so I'm a bit lost but I'll just ask my question: So if destroy isn't blocked, it automatically marks it as reviewed? Because if that's the case, that's wrong right?

src/panels/DestroyModelsPanel/AccordionTitle/AccordionTitle.tsx
ale8k: Honestly, as Juju could change the conditions for destruction of entities at any point, it genuinely worries me we're trying to duplicate them in the dashboard for reporting. I would just expect Juju to report back if it is possible without trying to duplicate the [potentially] moving goal post.

ale8k (commented):
Won't approve, but leaving some thoughts

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
Ninfa-Jeon: If destroy is blocked, it hides the action buttons. If it isn't blocked, it shows the buttons to the user.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
Ninfa-Jeon: Well the function is not rendering the row, only building it for render

src/hooks/useModelDestructionData.ts
Ninfa-Jeon: From memory, the destroy starts but doesn't seem to end. We hate the UX for this but rather than allowing the user to trigger a sure-to-fail operation, we block them.

src/hooks/useModelDestructionData.test.tsx
Ninfa-Jeon: This import pattern is used because of how ES module mocking works with Vitest. We are to mock this module in this file and this is the most lightweight pattern for this.

src/components/Panel/_panel.scss
Ninfa-Jeon: Yes, the panel content was visible over the footer component. Upon checking, the `button` inside the `<Accordion/>` component from `react-components` has a `z-index: 2` which is why it was visible over the footer. In retrospect, I could lower the accordion button's `z-index` to `0` instead. I'll do whichever you like better :) 
Bottomline, I will have to change the `z-index` for one of the two.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
Ninfa-Jeon: Just a very typescript way of filtering out `null` values from the array, with a TypeScript type guard so the compiler knows the result contains no nulls. Typescript doesn't know on runtime that the row is not null although we are definitely providing it non-nulls. Just narrowing the type.

src/components/Panel/_panel.scss
ale8k: No preference, just thinking about it. My QA was dead so couldn't QA it but for instance on mobile it might break for example.

src/hooks/useModelDestructionData.ts
ale8k: Really? That's interesting, perhaps the CLI is removing offers first then I'd need to check.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
ale8k: Ahhh I'm with ya.

src/hooks/useModelDestructionData.ts
```diff
@@ -19,6 +21,13 @@ type ConnectedOffer = {
   endpoint: { name: string; interface: string };
 };
 
+export enum DestroyBlockedReason {
+  NONE = "",
```
huwshimi: It seems strange to have an enum value for when the destroy is not blocked. Wouldn't it be better to set the variables to `null` (it's also semantically a little confusing, i.e. is it blocked without a known reason or is it not blocked)?

src/panels/DestroyModelsPanel/AccordionContent/_accordion-content.scss
```diff
@@ -0,0 +1,41 @@
+.accordion-content {
+  border-top: 1px solid $colors--theme--border-low-contrast;
+
+  &__info-table {
```
huwshimi, line 4: Is there a reason these icons are scoped to the table?

src/panels/DestroyModelsPanel/AccordionContent/_accordion-content.scss
huwshimi, line 1: other than the border, the spacing should use vanilla variables or rems rather than pixels.

src/panels/DestroyModelsPanel/AccordionContent/AccordionContent.tsx
huwshimi: You could use our filterBoolean util here I think, unless the you need it to not omit other falsy values.

src/panels/DestroyModelsPanel/AccordionTitle/AccordionTitle.tsx
```diff
@@ -0,0 +1,59 @@
...
+        </div>
+        <div className="accordion-title__model-summary-item">
+          <Icon name="units" className="u-no-margin--top" />
+          {unitCount}
+        </div>
+        <div className="accordion-title__model-summary-item">
+          <Icon name="machines" className="u-no-margin--top" />
+          {machines.length}
+        </div>
+      </div>
```
huwshimi, line 52: It would be nice if this could be a shared component with the model list.

src/panels/DestroyModelsPanel/AccordionTitle/_accordion-title.scss
```diff
@@ -0,0 +1,25 @@
...
+    display: flex;
+    float: right;
+    gap: $spv--small;
+  }
+
+  &__model-summary-item {
+    align-items: center;
+    border-right: 1px solid $colors--light-theme--border-default;
+    display: flex;
+    float: left;
```
huwshimi: The parent has flex so I'm not sure what this float is changing.

src/panels/DestroyModelsPanel/AccordionTitle/_accordion-title.scss
```diff
@@ -0,0 +1,25 @@
+.accordion-title {
+  &--is-removed {
+    text-decoration: line-through;
+  }
+
+  &__model-summary {
+    display: flex;
+    float: right;
+    gap: $spv--small;
```
huwshimi: All these variables where you've used `$spv--*` need to use the correct `$sph--` variant (v = vertical i.e. top/bottom, h = horizontal i.e. left/right).

huwshimi (commented):
When a model can't be deleted and it also doesn't have any apps then the expanded view is empty:

Image

src/panels/DestroyModelsPanel/AccordionContent/_accordion-content.scss
Ninfa-Jeon: We don't need them in `color-mid-dark` outside of this table.

Ninfa-Jeon:
> When a model can't be deleted and it also doesn't have any apps then the expanded view is empty

Thanks for pointing that out. For edge cases like this, I will make a follow-up PR because there are labels that need to be put there depending on why the destroy is blocked.

src/panels/DestroyModelsPanel/AccordionTitle/_accordion-title.scss
Ninfa-Jeon: oops

src/panels/DestroyModelsPanel/AccordionTitle/_accordion-title.scss
Ninfa-Jeon: Sorry, leftover rules.

src/panels/DestroyModelsPanel/AccordionTitle/AccordionTitle.tsx
Ninfa-Jeon: Besides the visual differences between the two, this one renders plain counts with icons while the `ModelSummary` on models list renders them with a link to navigate to another page in the dashboard.

This one:
Screenshot 2026-09-04 at 2 23 41 PM
Models list:
Screenshot 2026-09-04 at 4 29 30 PM

While it would still be technically possible, I'm unsure if loading the component with multiple conditions for a small thing like this is the best course.

Ninfa-Jeon: Let me know if you still want that implemented though.

src/hooks/useModelDestructionData.ts
Ninfa-Jeon: i have removed it and built logic over null for when a model can be destroy, i.e., destroy isn't blocked

huwshimi (approved):

kian99 (approved):
I didn't review the code but QA worked well for me. I had to do  some extra setup as follows:
1. Install [dotrun](https://github.com/canonical/dotrun) (I used the `curl ... | bash` method)
2. Update JIMM's docker compose to tweak the CORS setting and redirect back to the development dashboard. See the diff below.
3. Setup JIMM with `docker compose --profile dev up -d --wait` and added a LXD controller `make qa-lxd` also should be fine for this.
4. Went to http://localhost:8036/models and followed the PR instructions.

I created 3 models, one with Postgres deployed and another 2 that involved a CMR. The data in the "review destroy" side panel was accurate. Nice work!

```diff
-      JIMM_DASHBOARD_FINAL_REDIRECT_URL: "https://juju-dashboard.localhost"
-      CORS_ALLOWED_ORIGINS: "https://juju-dashboard.localhost"
+      JIMM_DASHBOARD_FINAL_REDIRECT_URL: "http://localhost:8036"
+      CORS_ALLOWED_ORIGINS: "https://juju-dashboard.localhost http://localhost:8036"
```

PR #2482, opened 2026-09-14
chore: add gating logic and tooltip labels for edge cases

Ninfa-Jeon:
Depends on https://github.com/canonical/juju-dashboard/pull/2470 and https://github.com/canonical/juju-dashboard/pull/2476

[Changes from this PR only](https://github.com/Ninfa-Jeon/juju-dashboard/compare/batch-destroy-review...Ninfa-Jeon:juju-dashboard:batch-destroy-edge-cases).

## Done

- Added gating logic and appropriate tooltip messages to "Review & Destroy" button
- Added specific tooltips to AccordionTitle for different cases of blocked destroy
- Replaced all occurrences of `removed` with `skipped`
- Added tests

## QA

- Pull code
- Run `dotrun clean && dotrun serve`
- Open http://0.0.0.0:8036/
- Try interacting with the bulk destroy flow as follows:
  - The review & destroy button stays disabled with appropriate tooltip when the only selected model to destroy is a:
     1) controller model
     2) model that the user does not have access to destroy
  - The review & destroy button allows the user to enter single model destruction flow for a model with consumed offers, however, the final cta remains disabled (existing logic, no change here)
  - Inside the sidebar, the accordion item is removed from selection with appropriate tooltip by default if it is a:
     1) controller model
     2) model that the user does not have access to destroy
     3) model with consumed offers
  - The cta on the sidebar remains enabled if at least one model of the selection is not removed from the selection (by default or manually)

## Details

https://warthogs.atlassian.net/browse/JUJU-10324

## Screenshots

https://github.com/user-attachments/assets/bd03a8a8-a9b5-432a-9515-311c516182d6

PR #2485, opened 2026-09-22
feat: add bulk model destruction dialog

Ninfa-Jeon:
## Done

- Implemented the `BulkDestroyModelDialog`

## QA

- Pull code
- Open http://0.0.0.0:8036/
- Select multiple models and click `Review & destroy x models` button at the top right of the models list page
- A sidebar should open with this list of models and their specific details
- The controller model, a model that you don't have access to delete and a model with connected offers should be struck from the selection on render
- All models should show accurate info inside the accordion content
- The count of skipped, reviewed and destroyable models should be correct in the footer
- When you have at least one destroyable model, the submit button should be enabled
- If you have only one such model, you should see a normal model destruction model (existing flow)
- If you have more than one destroyable models, you should see the new `BulkDestroyModelDialog` as per [figma](https://www.figma.com/design/Rs1v3IS0oY2RK4CU4yo4Ly/25.10-JAAS-model-management---Juju?node-id=2647-203571&p=f&m=dev)
- All info on the dialog should be accurate
- You should see a warning if you moved on to the dialog without manually marking the models as `Reviewed`. This is normal as reviewing is advised not compulsory.
- You should see an input prompt that blocks the final submit unless you type it out correctly.
- The destruction itself should work as expected

Note: The notifications are a WIP hence decided, per-model toasts will be shown as a part of this PR.

## Details

Part of work for [JUJU-10311](https://warthogs.atlassian.net/browse/JUJU-10311) and [JUJU-10371](https://warthogs.atlassian.net/browse/JUJU-10371)

## Screenshots

https://github.com/user-attachments/assets/0d832e27-5fbe-46b8-ba6c-141ff05848b5

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

[JUJU-10311]: https://warthogs.atlassian.net/browse/JUJU-10311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/DestroyModelDialog/DestroyModelDialog.tsx
```diff
@@ -196,6 +200,7 @@ export default function DestroyModelDialog({
         wsControllerURL,
       }),
     );
+    afterConfirmClicked?.();
```
Ninfa-Jeon, line 203: An optional callback so that when this dialog is triggered from the bulk destruction sidebar, we also close the sidebar on submit.

src/panels/DestroyModelsPanel/DestroyModelsPanel.tsx
```diff
@@ -29,57 +37,78 @@ const DestroyModelsPanel: FC = () => {
...
-      drawer={
-        <div className="destroy-models-panel__actions">
-          <div className="u-align--left">
-            {`${reviewed.length}/${destroyable.length} Reviewed`}
-            {skipped.length > 0 ? `, ${skipped.length} Skipped.` : null}
+    <>
+      {showDestroyDialog && (
+        <Portal>
+          {destroyable.length === 1 ? (
+            <DestroyModelDialog
```
Ninfa-Jeon: open the existing single model destruction drawer if the effective number of destroyable models is just 1.

src/components/DestroyModelDialog/DestroyModelDialog.test.tsx
```diff
@@ -132,6 +132,12 @@ describe("DestroyModelDialog", () => {
             offers: {
               http: applicationOfferStatusFactory.build({
                 "total-connected-count": 2,
+                endpoints: {
+                  mockEndpoint: {
+                    interface: "mockInterface",
+                    name: "mockName",
+                  },
+                },
```
Ninfa-Jeon: just covering a previously uncovered branch

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+                content: (
+                  <ResourceCount
+                    resourceType={ResourceType.STORAGE}
+                    resources={storageIDs}
+                  />
+                ),
+              },
+            ],
+          }
+        : null,
```
Kxiru: Can this be simplified into some kind of iteration?

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+    }
+  };
+
+  return (
+    <ConfirmationModal
+      title={<div>Destroy {destroyableCount} models</div>}
+      {...testId(TestId.DIALOG)}
+      className="bulk-destroy-model-dialog"
+      confirmButtonLabel={`Destroy ${destroyableCount} models`}
+      confirmButtonDisabled={!isConfirmValid}
```
Kxiru: Does the dashboard have any understanding of "permissions"? Or do only admins have access to this? If you do have the concept of permissions, should the confirmation button be permission-gated?

src/panels/DestroyModelsPanel/DestroyModelsPanel.tsx
Kxiru: Image

In your screen recording, you select 7 models, but 4 of them are already struck through when you open up the side panel. Is there some kind of form persistence happening, and is this expected / required behaviour? From a UX perspective, I would expect a clean slate of the 7 items I selected, to be put up for review.

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+        {reviewed.length < destroyableCount ? (
+          <div className="u-flex u-flex-items-center u-flex--gap-x-small is-caution">
+            <Icon name="warning" />
+            {reviewed.length}/{destroyableCount} models have not been reviewed.
+          </div>
+        ) : null}
+        <div className="u-sv1 u-sv1--top">
+          By destroying {destroyableCount} models, you will also be destroying
+          all applications, machines and storage within. This is irreversible.
+        </div>
```
Kxiru: As this is used twice, once in the panel and once in the modal, perhaps it should be a component? I wonder if you could use the information or warning notification from vanilla components [here](https://canonical.github.io/react-components/?path=/docs/components-notifications-notification--docs). (If you already use vanilla/React components, of course).

Kxiru (commented):
The logic and screen recording look great (thanks for including these) and the side panel and dialogue are intuitive. Tests seem to have good coverage over all the dialog behaviours.

I do wonder why a "skip" function is required when you can just not select the models which you do not want included in the bulk deletion. It would simplify the implementation to do so.

In future, I suggest building operations and their tests in smaller PRs to enable easier review. This feature-rich PR encompasses multiple concerns, making it harder to identify what logic relates where, and to spot errors.

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
Ninfa-Jeon: Yes! We do have permission gating but not to this confirm button. It exists at the very beginning of the flow making sure that a user without permission never makes it this far.

src/panels/DestroyModelsPanel/DestroyModelsPanel.tsx
Ninfa-Jeon: We do have a lightweight slice in the redux store to make sure the checkboxes, the sidebar and the final dialog box are working on the same data, however we clear this slice as soon as the sidebar is closed. The reason a bunch of selected models are struck for review is because the user either doesn't have the permission to delete or other related "cannot delete" issues. We provide more context on the "why" inside the help tooltip next to such struck items. You can find more on this in a previous PR just for this logic [here](https://github.com/canonical/juju-dashboard/pull/2482).

Ninfa-Jeon: https://github.com/user-attachments/assets/c29f7eee-08a9-4165-901a-20c0220f01a1

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
Ninfa-Jeon: Great catch! I completely missed that the text was near identical. While a notification cannot be used as this appears as plain text on the screen, I could extract this to a component but it would just be a named `div` wrapper around this text. The appearances are different so I might have to send in classnames just like I'm doing now with the `div`s, do you think extracting a simple div to a component is worthwhile?

src/components/BulkDestroyModelDialog/BulkDestroyModelDialog.tsx
Ninfa-Jeon: wonderful idea! on it.

Ninfa-Jeon: Done!

